PR #2, opened 2018-06-20
Roll loot for a single raid specifying the number of points

ms813:


commands/RuneScape - Fun/raid.js
```diff
@@ -0,0 +1,189 @@
...
+		const reply = [
+			`You finished a raid with ${points} points.`,
+			`You get ${chances.length} roll${plural} on the loot table.`
+		];
+
+		chances.forEach((chance, i) => {
+			const drop = loot[i] ? `${loot[i].name} ${RaidsEmojis[loot[i].shortName]}` : 'You no lucky';
+			const prefix = chances.length > 1 ? `Roll ${i + 1}:` : '';
+			const dropChance = Math.round(chance * 100);
+			reply.push(`${prefix} ${dropChance}% chance for a drop - ${drop}`);
```
gc: If there is only 1 roll, theres an extra space inserted at the beginning of this line
untitled

commands/RuneScape - Fun/raid.js
```diff
@@ -0,0 +1,189 @@
...
+const RaidsEmojis = require('../../resources/monsters/raids').drops;
+
+module.exports = class extends Command {
+
+	constructor(...args) {
+		super(...args, {
+			cooldown: 1,
+			description: 'Roll on the raids drop table based on the number of points specified',
+			usage: '<points:int{1}>',
+			usageDelim: ' '
```
gc: According to the wikipedia: 

> The current maximum amount of points attainable per player is 131,071 points.  

To limit the input to only being from 1 to 131,071, please change the `usage` config to be `usage: '<points:int{1,131071}>',`

gc: you can completely delete the usageDelim line, as there is only 1 input parameter, theres nothing to Deliminate

ms813: Function is meant to be for a raid of any team size not just solos. 
Perhaps the maximum number of points should be limited to 1,710,000 which is where all 3 rolls cap out at 80% drop chance?

gc: sounds good

gc:
Merged. Tag me in the support server for your contributor role :)

PR #14, opened 2018-10-28
+kill/finish argument list

devin8:
added field to +help
added argument list for monsters/bosses in +kill/finish


gc:
I think it would be better to show these to the user if they use `+kill` or `+finish` with no arguments, or with an invalid monster name. e.g.:

```
The available monsters are: Chaos Fanatic, Skotizo, Jad, .....```

devin8:
moved boss argument link to +kill/finish

PR #15, opened 2018-10-28
+diary update

devin8:
added boosts to diary command - indicated by bold text


gc:
What's this new format?
```js
Thieving: [0, 40, 50, 0],
Thieving: [0, 40, 58, 13, 0, 1, 5, 0],```

it seems you've changed the unboosted requirements, and I dont see exactly how you've put the boosted requirements in, i think it would be better to keep them separated so that its clear which are the requirements for which
```js
Thieving : {
    boosted: [],
    unboosted: []
}```

devin8:
The format was stat requirements for easy, medium, hard, elite and the boosts available for easy, medium, hard, elite. I've changed it to make it more clear - following the structure you provided. 

For changing the unboosted requirements, there were lots of required quests that weren't being accounted for previously - since there is so much data, there could be some wrong stats. However, I am confident it is more accurate than before.

PR #18, opened 2018-10-31
+kill dks fix

devin8:
Spaces in const loot caused error when nothing received. Spaces were added to format it correctly. Formatting is now moved to msg.send() function.


gc:
fixed https://github.com/gc/oldschoolbot/commit/bae09318936c320c79a043ba2d04495863d906ac

PR #19, opened 2018-11-02
+kill gwd expansions

devin8:
made gwd bosses able to handle larger quantity of kills - also added on a few extra cases for kurasks, brutal black dragons, and lava dragons


gc:
Whats with the changes to the imports at the top?

devin8:
I am not sure why it keeps saying there are changes elsewhere, but there is none. It is just cases for kurasks, brutal black dragons, lava dragons and kill count number for the other gwd bosses in kill.js. It might be comparing to my forked master file? I am not quite sure.

PR #21, opened 2018-11-09
Added a calculator command using math.js

devin8:
mathjs needs installing into ../node_modules from http://mathjs.org/download.html


gc:
Does it support stuff like `12m` and `500k`? if not, we should work on supporting that

PR #22, opened 2018-11-26
fix false positives for 73

FearrMe:
this should filter out all false positives (mentions, emojis) from the boaty integer thing
hope i didnt do anything wrong i havent used javascript before

### DIFF
--- a/monitors/73.js
+++ b/monitors/73.js
@@ -8,7 +8,7 @@ module.exports = class extends Monitor {
 
 	async run(msg) {
 		if (!msg.guild || !msg.guild.configs.joyReactions) return;
-		if (msg.content.includes(' 73') || msg.content === '73') {
+		if (msg.cleanContent().replace(/<a?:.*?:[0-9]+>/g, '').includes('73')) {
 			msg.react('😂').catch(() => null);
 		}
 	}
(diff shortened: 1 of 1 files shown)


gc:
Where are you getting `msg.cleanContent()` from? I tested out your regex and I don't think it has the effect you intended (its not removing the emotes from the string, which I think you intended it to)
discord_2018-12-17_21-54-26

FearrMe:
There are [no situations](https://discordapp.com/developers/docs/reference#message-formatting) where a string like `<example:1337>` gets sent. The only 2 cases you have to bother with(since `msg.cleanContent()` cleans mentions) are `<:NAME:ID>` and `<a:NAME:ID>`. Both [should get caught](https://i.imgur.com/HAKTLXG.png) with my regex.


I found `msg.cleanContent()` in the [discord.js API](https://discord.js.org/#/docs/main/stable/class/Message?scrollTo=cleanContent).

Stylomonnik:
e

gc:
Looks good to me, except `msg.cleanContent` is a property and not a method - also note that the bot is using the master branch of d.js (cleanContent is the same for both though)

FearrMe:
o

you can just make the changes yourself as i have no idea how to use git

PR #23, opened 2018-12-13
Added mod bruno to jmod tweets

robinwithes:
Added mod bruno to jmod tweets


gc:
Thanks! Landed here: https://github.com/gc/oldschoolbot/commit/2d3e4544966773155a3ed50b7344358b45022f55

PR #24, opened 2018-12-13
Removed Mod Ayiza from tweets.js

robinwithes:
Ayiza is officially leaving jagex on the 14th.


gc:
Thanks! Landed here: https://github.com/gc/oldschoolbot/commit/93123dc7df694ded166fef9c9fb479c1b671c88c

RIP mod ayiza and mod stone

PR #25, opened 2018-12-29
Update wintertodt.js

devin8:
previously drop rate was based on 1 roll / kill. At 500 points, there is 2 rolls / kill - therefore drop rates are cut in half.


gc:
looks right to me.looks right to me.

PR #62, opened 2019-07-11
tob: update/refactor tob item drop weights

Alexsuperfly:
based on Kierens recent tweet the rate of tob items is 1/9.1 for a deathless run
this also refactors the loot determination process to not have it duplicated between small and big kills

### DIFF
--- a/data/monsters/tob.js
+++ b/data/monsters/tob.js
@@ -1,24 +1,53 @@
 const tob = {
 	drops: {
-		pet: '<:Lil_zik:479460344423776266>',
-		scytheOfVitur: '<:Scythe_of_vitur_uncharged:455403545294929920>',
-		sanguinestiStaff: '<:Sanguinesti_staff_uncharged:455403545298993162>',
-		justiciarLegguards: '<:Justiciar_legguards:455403545265700874>',
-		justiciarFaceguard: '<:Justiciar_faceguard:455403544971968523>',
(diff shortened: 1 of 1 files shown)


data/monsters/tob.js
```diff
@@ -31,32 +60,7 @@ const tob = {
...
-						break;
-					case number <= 11:
-						loot.push(this.drops.justiciarChestguard);
-						break;
-					default:
-						loot.push(this.drops.avernicDefenderHilt);
-						break;
-				}
-			}
+			if (this.roll(9.1)) loot.push(this.determineItem().emoji);
```
gc: The `.1` of this is not effecting the chance of the roll. Also, the `this.determineItem().emoji` gives an error on pretty much every command usage as `this.determineItem()` doesnt return anything, and it tries to read `undefined.emoji`.

`.1` is used elsewhere in the changes too. You can not use `roll` at all and just make a custom RNG method for CoX based on however it works.

Alexsuperfly: well thats what i get for not trying to fix my local build to even run, that if in determineItem was leftover that i didn't delete when i initially built it, sorry about that.

as far as i can tell the `.1` will actually do something in the roll method since its multiplied before the floor

gc:
LGTM. Sorry for the big delay I've been a bit busy! 

Thanks!

PR #68, opened 2019-09-30
raid: correct point threshold values

### LINKED ISSUE #66: +raid (points) wrong calculation
Points at olm is according to the wiki points/8675 (https://oldschool.runescape.wiki/w/Chambers_of_Xeric).

Using their example: 413,250/8,675 ≈ 47.6%, or 0.476

Doing +raid 413250 gives 58% of purple.

Alexsuperfly:
fixes #66


gc:
Thanks :)

PR #71, opened 2019-10-16
finish: fix olm to save item with pet

### LINKED ISSUE #70: Minor issue + suggestion
https://i.imgur.com/pNIsZch.png

865 kc olmlet with ances legs, but also 865 kc bow and first pair of ances legs at 1990 kc.
Probably just something like pet rolling separately or something, but figured I'd post it anyways.

As for the suggestion, I was wondering if you'd consider increasing the amount of rolls some monsters can do with the +kill command? For example, shamans have a max of 1000, as do a few others. 

Thank you!

Alexsuperfly:
i thought i accounted for it but that was only the kill command, oops
Fixes #70 

image


gc:
Thanks!

PR #89, opened 2020-01-20
Updated Minion.ts

SomeRangerOSRS:
Added a new command for +bosses which lists the name of each boss and how many can be done per 30 minutes.

### DIFF
--- a/src/commands/Fun/minion.ts
+++ b/src/commands/Fun/minion.ts
@@ -148,6 +148,15 @@ export default class extends BotCommand {
 			return msg.channel.send('Cancelled minion purchase.');
 		}
 	}
+	
+	async bosses(msg: KlasaMessage) {
+		await msg.author.settings.sync(true);
+		if (!this.hasMinion(msg)) {
+			throw hasNoMinion(msg.cmdPrefix);
+		}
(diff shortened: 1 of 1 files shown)


src/commands/Fun/minion.ts
```diff
@@ -148,6 +148,15 @@ export default class extends BotCommand {
 			return msg.channel.send('Cancelled minion purchase.');
 		}
 	}
+	
+	async bosses(msg: KlasaMessage) {
+		await msg.author.settings.sync(true);
+		if (!this.hasMinion(msg)) {
+			throw hasNoMinion(msg.cmdPrefix);
+		}
+		return msg.send('Monster Name (Amount Killable in An Hour) n\ Barrows (7) \n Dagannoth Prime (14) \n Dagannoth Rex (14) \n Dagannoth Supreme (14) \n Cerberus (12) \n Giant Mole (18) \n Vorkath (10) \n Zulrah (10) \n General Graardor (6) \n Commander Zilyana (6) \n Kree arra (6) \n Kril Tsutsaroth (6) \n Man (409) \n Woman (409) \n Goblins (409) \n Guard (243)');
```
Alexsuperfly: these values will become incorrect when a process to upgrade the minion gets released

SomeRangerOSRS: When upgrading a minion is possible we can use a variable to alter how many can be killed based on if the user has an upgraded minion or not.

Alexsuperfly: do that now then

PR #91, opened 2020-01-21
Updated cmd.js

SomeRangerOSRS:
Updated the Cmd.Js to require permission level 6 rather than 7 as per https://klasa.js.org/#/docs/klasa/stable/Tutorials/UnderstandingPermissionLevels

This should stop the "You do not have permission to use this command." error for people with Manage Server rights. 

Klasa states the following 

6 | false | Members of guilds must have 'MANAGE_GUILD' permission
7 | false | Guild Owner


gc:
Closing for reasons discussed in discord.

PR #92, opened 2020-01-21
Remove pay command

Alexsuperfly:
all the command does is enable people to create alt accounts to collect dailies and pay them over to one main to cheat the leaderboard, removing this would allow you to stop having to police users attempting to break the rules.

I reduced the minion buy amount to account for not being able to have a friend give you the startup money to be able to own one, the buy in was already low enough it wasn't a useful gp sink any ways but having to buy one still gives the user some satisfaction.


gc:
Not something I want to do, as discussed.

PR #94, opened 2020-01-24
Minion: remove quant from clue task

Alexsuperfly:


gc:
Needs to be prettier'd/linted

PR #104, opened 2020-01-28
sell all (+ --max= flag)

arkoe:
Implemented `+sell all` with optional `--max=` flag (for example: `--max=50k` or `--max=1000`.


shermaza:
Should be able to resolve DeepScan issues by making undefined a string since typeof returns a string. `typeof itemName === "undefined"`

PR #121, opened 2020-02-04
give command

wyattos:


src/commands/Utility/give.ts
```diff
@@ -0,0 +1,78 @@
...
+            oneAtTime: true,
+            cooldown: 5,
+            altProtection: true
+        });
+    }
+
+    async run(msg: KlasaMessage, [user, quantity, itemName]: [KlasaUser, number, string]) {
+        await msg.author.settings.sync(true);
+        let re = /’/gi;
+        const osItem = Items.get(itemName.replace(re, "'"));
```
gc: should use the `cleanItemName` util (see sellto command)

src/commands/Utility/give.ts
```diff
@@ -0,0 +1,78 @@
...
+        if (user.id === msg.author.id) throw `You can't send items to yourself.`;
+        if (user.bot) throw `You can't send items to a bot.`;
+        if (!osItem) throw `That item doesnt exist.`;
+        if (
+            specialUntradeables.includes(osItem.id) ||
+            !('tradeable' in osItem) ||
+            !osItem.tradeable
+        ) {
+            throw `That item isn't tradeable.`;
+        }
```
gc: should use the `itemIsTradeable` util, see sellto command

src/commands/Utility/give.ts
```diff
@@ -0,0 +1,78 @@
...
+        );
+        try {
+            const collected = await msg.channel.awaitMessages(
+                _msg =>
+                    _msg.author.id === msg.author.id && _msg.content.toLowerCase() === 'give',
+                options
+            );
+            if (!collected || !collected.first()) {
+                throw "This shouldn't be possible...";
+            }
```
gc: no need for the `if` or `const collected =` anymore. (see sellto cmd)

PR #123, opened 2020-02-06
Clues: add mimic possibility to casket and open commands

Alexsuperfly:
depends on gc/oldschooljs#60

### DIFF
--- a/src/commands/Minion/open.ts
+++ b/src/commands/Minion/open.ts
@@ -1,10 +1,11 @@
 import { KlasaMessage, CommandStore } from 'klasa';
 import { MessageAttachment } from 'discord.js';
+import { Misc } from 'oldschooljs';
 
 import { Events } from '../../lib/constants';
 import { BotCommand } from '../../lib/BotCommand';
 import Openables from '../../lib/openables';
-import { stringMatches, itemNameFromID } from '../../lib/util';
+import { stringMatches, itemNameFromID, addBankToBank, roll } from '../../lib/util';
 import bankFromLootTableOutput from '../../lib/util/bankFromLootTableOutput';
 import createReadableItemListFromBank from '../../lib/util/createReadableItemListFromTuple';
 import { cluesRares } from '../../lib/collectionLog';
@@ -43,9 +44,15 @@ export default class extends BotCommand {
 
 		await msg.author.removeItemFromBank(clueTier.id);
 
-		const loot = clueTier.table.open();
+		let loot = clueTier.table.open();
+		let opened = `You opened one of your ${clueTier.name} Clue Caskets`;
 
-		const opened = `You opened one of your ${clueTier.name} Clue Caskets`;
+		let hadMimic = false;
+		if (clueTier.mimicChance && roll(clueTier.mimicChance)) {
+			loot = addBankToBank(Misc.Mimic.open(clueTier.name as 'master' | 'elite'), loot);
+			opened += ' and defeated the Mimic inside';
+			hadMimic = true;
+		}
 
 		const nthCasket = msg.author.settings.get(UserSettings.ClueScores)[clueTier.id] ?? 0 + 1;
 
@@ -75,7 +82,7 @@ export default class extends BotCommand {
 			);
 		}
 
-		if (keys.length === 0) {
+		if (Object.keys(loot).length === 0) {
 			return msg.send(`${opened} and got nothing :(`);
 		}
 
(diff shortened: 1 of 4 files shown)


src/commands/Fun/open.ts
```diff
@@ -20,6 +20,10 @@ export default class extends BotCommand {
 		});
 	}
 
+	roll(max: number) {
```
gc: use the util instead of copypasting this

src/commands/Fun/open.ts
```diff
@@ -33,9 +37,19 @@ export default class extends BotCommand {
...
-
-		const opened = `You opened one of your ${clueTier.name} Clue Caskets`;
+		let loot = clueTier.table.open();
+		let opened = `You opened one of your ${clueTier.name} Clue Caskets`;
+
+		const MIMIC_INDEX = 6;
+		let hadMimic = false;
+		if (clueTier.mimicChance > 0) {
+			if (this.roll(clueTier.mimicChance)) {
+				console.log('in mimic');
```
gc: no console.logs

src/lib/clueTiers.ts
```diff
@@ -10,42 +10,56 @@ const clueTiers: ClueTier[] = [
...
+		mimicChance: 15
+	},
+	{
+		name: 'Mimic',
+		table: Mimic,
+		id: 23184,
+		scrollID: 23184,
+		timeToFinish: 0,
+		mimicChance: 0
 	}
```
gc: as discussed in discord

src/commands/Minion/open.ts
```diff
@@ -2,10 +2,11 @@ import { KlasaMessage, CommandStore } from 'klasa';
 import { MessageAttachment } from 'discord.js';
 
 import ClueTiers from '../../lib/clueTiers';
+import { Misc } from 'oldschooljs';
```
gc, line 5: we always put external imports at the very top

src/commands/Minion/open.ts
```diff
@@ -32,9 +33,18 @@ export default class extends BotCommand {
...
-		const loot = clueTier.table.open();
-
-		const opened = `You opened one of your ${clueTier.name} Clue Caskets`;
+		let loot = clueTier.table.open();
+		let opened = `You opened one of your ${clueTier.name} Clue Caskets`;
+
+		let hadMimic = false;
+		if (clueTier.mimicChance > 0) {
+			if (roll(clueTier.mimicChance)) {
+				console.log('in mimic');
```
gc, line 42: remove the log

src/commands/OSRS_Fun/casket.ts
```diff
@@ -2,8 +2,10 @@ import { KlasaUser, KlasaMessage, Command, CommandStore } from 'klasa';
 import { MessageAttachment } from 'discord.js';
 
 import clueTiers from '../../lib/clueTiers';
+import { Misc } from 'oldschooljs';
```
gc, line 5: ^

src/commands/Minion/open.ts
```diff
@@ -55,10 +65,14 @@ export default class extends BotCommand {
...
 			loot,
-			`You opened a ${clueTier.name} clue and received...`
+			`You opened a ${clueTier.name} clue ${hadMimic ? 'with a mimic ' : ''}and received...`
 		);
 
 		msg.author.incrementClueScore(clueTier.id);
+		const MIMIC_ID = 23184;
+		if (hadMimic) {
+			msg.author.incrementClueScore(MIMIC_ID);
+		}
```
gc, line 75: this should be removed, as mimics arent clues

Alexsuperfly: if i were to check my clue kcs i would want mimic to be shown

Alexsuperfly: i would put this in monster scores but then it would need to be a killable monster which is even more wrong than it being a clue

src/commands/Minion/open.ts
```diff
@@ -43,9 +44,17 @@ export default class extends BotCommand {
 
 		await msg.author.removeItemFromBank(clueTier.id);
 
-		const loot = clueTier.table.open();
+		let loot = clueTier.table.open();
+		let opened = `You opened one of your ${clueTier.name} Clue Caskets`;
 
-		const opened = `You opened one of your ${clueTier.name} Clue Caskets`;
+		let hadMimic = false;
+		if (clueTier.mimicChance! > 0) {
```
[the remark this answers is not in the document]
Alexsuperfly: what about setting mimicChance to `0` on clues that cant get one, so you dont have to juggle two different types

PR #124, opened 2020-02-07
Increase response timer for continuing minion trip with "yes"

Ameobea:
* The two minute timer led to a large amount of my attempts to extend the minion trip being rejected.  Increasing it to 5 minutes gives a much more reasonable window for users realizing that their minion has returned and quickly sending them back.


gc:
I plan on increasing the response timer for patrons and maybe boosters, as a non-P2W perk, so wont be merging this.

PR #125, opened 2020-02-08
craft command

wyattos:


wyattos:
theres no current way to compare the skill requirements for items to skills that don't exist, thus the deepscan errors

PR #126, opened 2020-02-08
Improve collection log types

Ameobea:
Tighten up types involving collection log
   * Give a type to the `type` parameter used in `generateCollectionLogImage` derived from the array which it is pulled from
  * Make `removeDuplicatesFromArray` generic to preserve types
  * Make the desriminator key `name` of `collectionLogTypes` `const` to allow it to be matched on more clearly

Update `README.md` with a few other required steps for building
   * A couple new config items have been added since it was last updated, and a trivia database is required

### DIFF
--- a/src/lib/collectionLog.ts
+++ b/src/lib/collectionLog.ts
@@ -1,4 +1,5 @@
 import { removeDuplicatesFromArray } from './util';
+import { CollectionLogTypeName, CollectionLogType } from './types/index';
 
 export const bosses = {
 	Zulrah: [12921, 12936, 13201, 13200, 6571, 12927, 12922, 12932],
@@ -559,9 +560,9 @@ export const clues = {
 	'9421': [20128, 20131, 20134, 20137, 20140]
 };
 
-export const collectionLogTypes = [
+export const collectionLogTypes: CollectionLogType[] = [
 	{
-		name: 'Overall' as const,
+		name: CollectionLogTypeName.Overall,
 		aliases: ['all', 'overall'],
 		items: removeDuplicatesFromArray(
 			[...Object.values(bosses), ...Object.values(clues), ...Object.values(pets)].flat(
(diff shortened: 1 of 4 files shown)


gc:
Cant you do the `as const` on an interface instead of writing it manually for every item (P.S that file needs a rewrite, as its a mess, its just a really low priority for me atm)

The current setup instructions are wrong, and I'd rather contributors didnt have to make files to get it running, so I'll look at improving them myself now and making the trivia questions and stuff optional, so if you could remove the instructions changes from the PR ill fix those up.

Ameobea:
Putting `as const` on the array makes it into a tuple type, which imo isn't exactly what's trying to be expressed for that.  Also breaks that `ArrayElementOf` inference as a result.  There may be a better way to express that, but this is how I've always done it :p

For setup instructions, sounds good - I'll drop that commit and squash everything.

gc:
im not a fan of having to write `as const` on every object, are you sure you cant just make a new type/interface for the array items and declare it there?

Ameobea:
We could make it an enum if you'd prefer that:

```ts
export enum CollectionLogTypeName {
	Overall = 'Overall',
	Boss = 'Boss',
	Clue = 'Clue',
	Pets = 'Pets'
}

export type CollectionLogType = { aliases: string[] } & (
	| {
			name: CollectionLogTypeName.Overall;
			items: number[];
	  }
	| { name: CollectionLogTypeName.Overall; items: typeof bosses }
	| { name: CollectionLogTypeName.Clue; items: typeof clues }
	| { name: CollectionLogTypeName.Pets; items: typeof pets }
);

export const collectionLogTypes: CollectionLogType[] = [
	{
		name: CollectionLogTypeName.Overall,
		aliases: ['all', 'overall'],
		items: removeDuplicatesFromArray(
			[...Object.values(bosses), ...Object.values(clues), ...Object.values(pets)].flat(
				Infinity
			) as number[]
		)
	},
	{
		name: CollectionLogTypeName.Overall,
		aliases: ['bosses', 'boss'],
		items: bosses
	},
	{
		name: CollectionLogTypeName.Clue,
		aliases: ['clues', 'clue'],
		items: clues
	},
	{
		name: CollectionLogTypeName.Pets,
		aliases: ['pet', 'pets'],
		items: pets
	}
];
```

Of, if you don't mind not having constant keys for each type, you can do this:

```ts
export type CollectionLogType = { aliases: string[] } & (
	| {
			name: CollectionLogTypeName.Overall;
			items: number[];
	  }
	| {
			name: Exclude<CollectionLogTypeName, CollectionLogTypeName.Overall>;
			items: { [key: string]: number[] | number[][] };
	  }
);
```

Lmk what you prefer.

gc:
Looks good. Just move it in the central types file so its easier/simpler to access elsewhere

gc:
This PR needs to be rebased/updated please (conflicts)

gc:
Closing as its been 2 weeks with no updates. Feel free to re-open. This will probably have a mountain of conflicts, partially my fault for taking so long, I apologize - being busy and all that

Ameobea:
@gc this has been rebased and updated and should be ready to go.  If you can re-open the PR, that works otherwise lmk and I'll open a new PR.

PR #149, opened 2020-02-29
Remove Mod Gambit from tweets

Umdlye:


gc:
Thanks - Note that this wont come into effect until the next bot reboot.

PR #152, opened 2020-03-02
add missing hard ornament kits and +cl holiday

cadyb:
also moved dboots ornament kit from elite to hard which is where it should be


cadyb:
added `+cl holiday` the output looks like the following:
image

PR #161, opened 2020-03-11
Show 3 decimals in bank value when bank value >= 1B

Ameobea:
* This allows users to feel a better sense of progression when farming and more precisely judge the impact of trades on their net worth.

### DIFF
--- a/src/lib/util/formatBankValue.ts
+++ b/src/lib/util/formatBankValue.ts
@@ -1,8 +1,8 @@
-import { Util } from 'oldschooljs';
+import { toKMB } from 'oldschooljs/dist/util/util';
 
 const formatBankValue = (totalValue: number): string => {
 	if (totalValue < 1e9) {
-		return Util.toKMB(totalValue);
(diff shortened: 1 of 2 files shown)


gc:
This should be a standalone function in the util folder

Ameobea:
ok done.

src/lib/util/formatBankValue.ts
```diff
@@ -0,0 +1,22 @@
+import { toKMB } from 'oldschooljs/dist/util/util';
+
+const formatBankValue = (totalValue: number): string => {
+	if (totalValue < 1e9) {
```
gc: Use `const billion = 1_000_000_000;` instead (declared outside the function), for readability purposes

gc:
Closing as its been 2 weeks with no updates. Feel free to re-open.

Ameobea:
Srry, missed the feedback notification.

I've updated my branch with the requested change.  If you want, I can open another PR, or you could re-open this one.

PR #174, opened 2020-03-18
change +bank --text --full to show amount then item for copy paste

duracell33:


gc:
this doesnt make sense to me, do people want this? having the name first makes more sense to me

duracell33:
Discord_fjroPVgAyb
i was just looking for easy suggestions

gc:
As discussed in discord i think this should be opt-in with a msg flag

gc:
Closed due to inactivity

PR #178, opened 2020-03-21
update +m kc so it doesnt take up the whole page

wyattos:
this is what it'll look like https://imgur.com/a/B4D7bfJ
depending on how many monsters are killable 'if (i % 10 === 0)' will need to be raised


src/commands/Minion/minion.ts
```diff
@@ -99,15 +100,25 @@ ${Emoji.QuestIcon} QP: ${msg.author.settings.get(UserSettings.QP)}
...
 			const mon = killableMonsters.find(m => m.id === parseInt(monID));
 			res += `${mon!.emoji} **${mon!.name}**: ${monKC}\n`;
+			if (i % 10 === 0) {
+				embed.addField('\u200b', `${res}`, true);
+				res = ``;
+			}
+			i++;
 		}
-		return msg.send(res);
+		embed.addField('\u200b', `${res}`, true);
```
Alexsuperfly: only add if res is not empty

src/commands/Minion/minion.ts
```diff
@@ -99,15 +100,25 @@ ${Emoji.QuestIcon} QP: ${msg.author.settings.get(UserSettings.QP)}
...
 
 		const monsterScores = msg.author.settings.get(UserSettings.MonsterScores);
-
-		let res = `**${getMinionName(msg.author)}'s KCs:**\n\n`;
+		let res = ``;
+		let i = 1;
 		for (const [monID, monKC] of Object.entries(monsterScores)) {
 			const mon = killableMonsters.find(m => m.id === parseInt(monID));
 			res += `${mon!.emoji} **${mon!.name}**: ${monKC}\n`;
+			if (i % 10 === 0) {
```
Alexsuperfly: dynamically determine the breakpoint so its not just a single line of 10

src/commands/Minion/minion.ts
```diff
@@ -99,15 +100,25 @@ ${Emoji.QuestIcon} QP: ${msg.author.settings.get(UserSettings.QP)}
 		if (!msg.author.hasMinion) {
 			throw hasNoMinion(msg.cmdPrefix);
 		}
+		const embed = new MessageEmbed()
+			.setColor(Color.Orange)
+			.setTitle(`**${getMinionName(msg.author)}'s KCs:**\n\n`)
+			.setDescription(`Type ${msg.cmdPrefix}kill to see a list of killable monsters`);
```
Alexsuperfly: why does it need this description

gc:
has conflicts that need fixing

PR #181, opened 2020-03-22
Added master clues to craftables

duracell33:
sorry for pr spam and mess, still bad at using github


gc:
the issue with this is, if they create a master clue scroll and have one already, the command will go through but they wont get another one added to their bank. so we need to solve that in some way, for example we might need to add `cantHaveItems: ['Clue scroll (master)']` to the command/craftables.

src/commands/Minion/create.ts
```diff
@@ -51,6 +56,11 @@ export default class extends BotCommand {
 			throw `You don't have the required items to create this item. You need: ${inputItemsString}.`;
 		}
 
+		// Check for any items they cant have 2 of.
+		if (bankHasAllItemsFromBank(userBank, craftableItem.cantHaveItems)) {
```
gc: this is a bit flawed as it will only return true if the user has ALL the items from `cantHaveItems`

src/commands/Minion/create.ts
```diff
@@ -51,6 +56,11 @@ export default class extends BotCommand {
...
+			throw `You already have a ${cantHaveItemsString} in your bank.`;
```
gc: `a` isnt grammatically correct here as its a list

PR #192, opened 2020-03-27
Add birds nest

coolbop32:


src/tasks/minions/woodcuttingActivity.ts
```diff
@@ -45,6 +46,14 @@ export default class extends Task {
 			loot[WoodcuttingPet!] = 1;
 			str += `\nYou have a funny feeling you're being followed...`;
 		}
+		// roll for a birds nest with an item in it that is not a clue
+		let nests = 0;
+		for (let i = 0; i < quantity; i++) {
+			if (roll(256)) {
+				nests++;
+				loot[5073] = nests;
```
Alexsuperfly: move this out of the for loop

src/lib/simulation/birdsNest.ts
```diff
@@ -0,0 +1,48 @@
...
+const RingTable = new LootTable()
+	.every('crushed nest')
+	.add('Gold ring', 1, 32)
+	.add('Sapphire ring', 1, 32)
+	.add('Emerald ring', 1, 16)
+	.add('Ruby ring', 1, 8)
+	.add('Diamond ring', 1, 2);
+
+// chance at a piece of the evil chicken outfit
+const EggTable = new LootTable()
```
Alexsuperfly: this always drops a seednest with a chance at an outfit piece

gc:
Not a fan that its coded differently to how it works ingame, it should give the literal different kinds of nests instead of 1 nest that drops all the different kinds, when people open them, it will just open any of them as they all have the same name which is no issue

gc:
Closed due to inactivity.

PR #197, opened 2020-03-31
filter by category

wyattos:
+bank --filter=
current categories are 'gear', 'skilling', 'barrows'


wyattos:
> I think we could make this more user-friendly and easier to type maybe, if we did it like +bank --gear instead of +bank --filter=gear? Shouldn't be too hard

fixed

src/lib/util/filterByCategory.ts
```diff
@@ -0,0 +1,26 @@
...
+import { ItemTuple } from '../types';
+import { stringMatches } from '../util';
+import filterables from '../filterables';
+
+export default function filterByCategory(filterQuery: string, items: ItemTuple[]) {
+	const filteredBank: ItemTuple[] = [];
+	const filtered = filterables.find(
+		filtered =>
+			stringMatches(filtered.name, filterQuery) ||
+			stringMatches(filtered.name.split(' ')[0], filterQuery)
```
gc, line 12: the `|| ...` second check isnt really needed, none of the filterables are more than 1 word

src/lib/filterables.ts
```diff
@@ -0,0 +1,207 @@
+interface Filterable {
+	name: string;
+	items: string[];
+}
+
+const filterables: Filterable[] = [
+	{
+		name: 'Barrows',
+		items: [
```
gc, line 9: these arrays should  be arrays of item IDs, using `transformArrayOfResolvableItems()` util (and some code below changed after)

PR #199, opened 2020-04-01
add infernal axe to craftables

wyattos:


gc:
is it checking in the command if they have the firemaking level?

PR #204, opened 2020-04-05
Implements the alch command to highalch items

LethalLeonard:
Implements the alch command.

Ensures that the user has all required runes and items in order to cast high alchemy on their items. This takes 5 ticks/3 seconds per item.

`+alch rune platelegs` -> confirms with user, then alchs the max amount of rune platelegs possible
`+alch 5 rune platelegs` ->  confirms with user, then alchs 5x rune platelegs, 
`+alch rune platelegs --[cf/confirm]` -> alchs the max amount of rune platelegs without second confirmation.


LethalLeonard:
Hm interesting. After I wake up I’ll revert and redo the changes

gc:
We have alching implemented now

PR #208, opened 2020-04-06
Killtimes: change calculation to scale on time spent

Alexsuperfly:
simplify the equation and scale it with regards to time spend killing the mob instead of sharp drop offs

### DIFF
--- a/src/lib/minions/functions/reducedTimeFromKC.ts
+++ b/src/lib/minions/functions/reducedTimeFromKC.ts
@@ -1,24 +1,13 @@
 import { KillableMonster } from '../../killableMonsters';
 import { Time } from '../../constants';
 
-function determineCape(duration: number) {
-	if (duration < Time.Second * 30) return 50_000;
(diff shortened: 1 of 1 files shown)


src/lib/minions/functions/reducedTimeFromKC.ts
```diff
@@ -1,24 +1,13 @@
 import { KillableMonster } from '../../killableMonsters';
 import { Time } from '../../constants';
 
-function determineCape(duration: number) {
-	if (duration < Time.Second * 30) return 50_000;
-	if (duration < Time.Minute * 2) return 10_000;
-	return 1000;
-}
+const FIVE_HOURS = Time.Minute * 300;
```
gc, line 4: `Time.Hour * 5`

Alexsuperfly:
yes, i didnt do every single monster possible but i calced a few at where breakpoints should be and each came out correct when sending the minion off at right before and right after that breakpoint

PR #212, opened 2020-04-08
Buy: show gp required when throwing for not enough

Alexsuperfly:


src/commands/Minion/buy.ts
```diff
@@ -33,7 +33,8 @@ export default class extends BotCommand {
 
 		await msg.author.settings.sync(true);
 		const GP = msg.author.settings.get(UserSettings.GP);
-		if (GP < buyable.gpCost) throw `You don't have enough GP to buy this item.`;
+		if (GP < buyable.gpCost)
+			throw `You need ${toKMB(buyable.gpCost)} GP to purchase this item.`;
```
gc, line 37: if it doesnt fit on the same line, we always wrap it in `{}`, like:
```js
if (GP < buyable.gpCost) {
			throw `You need ${toKMB(buyable.gpCost)} GP to purchase this item.`;
}
```

PR #213, opened 2020-04-09
slayer skill

wyattos:
Few things about the pr...
src/commands/minion/minion i just put it as a 10% time boost thats applied the same way weekend boosts are
src/commands/minion/slayertask i left in the filter by combat level but commented it out in case magna wants to add combat level requirements, super easy to get rid of if needed
src/lib/skills i added all the combat skills in case the combat level requirement stays (that explains the 25 files changed)
the only slayer master right now is nieve and when we add turael a few things will need to be changed to make both work but thats easy as well
the shop isnt implemented yet, ie u cant skip/block tasks yet
the filter by combat level thing isnt necessary & is just extra work if we can easily get the data, although we will need to check the users combat level to make sure they can use the slayer master which is currently in


src/commands/Minion/slayertask.ts
```diff
@@ -0,0 +1,101 @@
...
+			throw `You don't have a high enough Slayer level to get a task from that Master.`;
+		}
+		let number = rand(1, totalweight);
+		for (let i = 0; i < filteredTasks.length; i++) {
+			number -= filteredTasks[i].weight;
+			if (number <= 0) {
+				const slayerMonster = filteredTasks[i];
+				const minQuantity = slayerMonster.amount[0];
+				const maxQuantity = slayerMonster.amount[1];
+				if (typeof minQuantity === 'number' && typeof maxQuantity === 'number') {
```
Alexsuperfly, line 85: probaby need to handle when this isnt true, unless you actually intend for it to just skip the task if it doesnt have quantities for some reason

src/lib/slayer/nieveTasks.ts
```diff
@@ -0,0 +1,530 @@
...
+			combatLevel: 65,
+			questPoints: 32
+		},
+		alternatives: ['Blue dragon', 'Baby blue dragon', 'Brutal blue dragon', 'Vorkath'],
+		weight: 4,
+		ID: Monsters.BlueDragon.id
+	},
+	{
+		name: 'Boss',
+		amount: [3, 20],
```
Alexsuperfly, line 128:
```suggestion
		amount: [3, 35],
```
special cases for barrows of 36 and zulrah of 15

wyattos: fixed in src/lib/slayer/bossTasks -- special case accounted for in the slayertask command

src/commands/Minion/slayertask.ts
```diff
@@ -0,0 +1,247 @@
...
+						_msg.content.toLowerCase() === 'confirm',
+					options
+				);
+			} catch (err) {
+				throw `Cancelled request to cancel ${
+					Monsters.get(msg.author.slayerTaskID)?.name
+				} slayer task.`;
+			}
+			const newSlayerPoints = msg.author.slayerPoints - 30;
+			await msg.author.settings.update(UserSettings.Slayer.SlayerTaskQuantity, 0);
```
gc, line 76: You should be able to update these all at once by passing it as a single tuple

src/commands/Minion/blocklist.ts
```diff
@@ -0,0 +1,149 @@
...
+					_msg =>
+						_msg.author.id === msg.author.id &&
+						_msg.content.toLowerCase() === 'confirm',
+					options
+				);
+			} catch (err) {
+				throw `Cancelling block list removal of ${taskname}.`;
+			}
+			await msg.author.settings.update(UserSettings.Slayer.BlockList, task[0].ID, {
+				arrayAction: 'remove'
```
gc, line 99: Use the enum for the array actions from Klasa, in all occurrences (Array actions)

src/commands/Minion/blocklist.ts
```diff
@@ -0,0 +1,149 @@
+import { CommandStore, KlasaMessage } from 'klasa';
+
+import { BotCommand } from '../../lib/BotCommand';
+import { stringMatches } from '../../lib/util';
+import { Monsters } from 'oldschooljs';
```
gc, line 5: External imports always at the top

src/lib/constants.ts
```diff
@@ -67,7 +68,21 @@ export const enum Emoji {
 	Warning = '⚠️',
 	Ironman = '<:ironman:626647335900020746>',
 	Firemaking = '<:firemaking:630911040175210518>',
-	EasterEgg = '<:easterEgg:695473553314938920>'
+	EasterEgg = '<:easterEgg:695473553314938920>',
+	runecraft = '<:runecraft:630911040435257364>',
```
gc, line 72: These should be pascalcase

src/extendables/UserExtendables.ts
```diff
@@ -117,6 +117,28 @@ export default class extends Extendable {
 		return this.settings.update(UserSettings.QP, newQP);
 	}
 
+	public async addSlayerPoints(this: User, amount: number) {
```
gc, line 120: Move all of these that you added to a new file, like I made Gear and Bank ones, make a Slayer one

src/lib/minions/monsters/gwdBosses.ts
```diff
@@ -0,0 +1,105 @@
+import { Monsters } from 'oldschooljs';
+
+import { Bank, ArrayItemsResolved } from '../../types';
+import { Time } from 'oldschooljs/dist/constants';
```
gc, line 4: Imported from wrong place

src/lib/schemas/UserSchema.ts
```diff
@@ -25,6 +25,23 @@ Client.defaultUserSchema
...
+	.add('slayer', folder =>
+		folder
+
+			.add('hasSlayerTask', 'boolean', { default: false })
+			.add('slayerTaskID', 'number', { default: 0 })
+			.add('slayerTaskQuantity', 'number', { default: 0 })
+			.add('slayerPoints', 'integer', { default: 0 })
+			.add('blockList', 'integer', { array: true, default: [] })
+			.add('unlockedAviansie', 'boolean', { default: false })
+			.add('unlockedBasilisk', 'boolean', { default: false })
```
gc, line 37: These should use a singular array setting as discussed

PR #222, opened 2020-04-13
Fix cooking

Alexsuperfly:


src/tasks/minions/cookingActivity.ts
```diff
@@ -10,6 +10,16 @@ import { channelIsSendable } from '../../lib/util/channelIsSendable';
...
+function getAmountBurned(qtyCooking: number, stopBurningLvl: number, cookingLvl: number) {
+	let burnedAmount = 0;
+	for (let i = 0; i < qtyCooking; i++) {
+		if (rand(0, 100) < stopBurningLvl - cookingLvl) {
+			burnedAmount++;
+		}
+	}
+	return burnedAmount;
+}
+
```
gc, line 22: should be in `lib/skilling/functions`

PR #225, opened 2020-04-15
Smithing items

coolbop32:
added smithing on standard items


src/lib/constants.ts
```diff
@@ -88,6 +88,7 @@ export const enum Tasks {
 	FishingActivity = 'fishingActivity',
 	MiningActivity = 'miningActivity',
 	SmithingActivity = 'smithingActivity',
+	SmithedActivity = 'smithedActivity',
```
gc, line 91: the names Smithed  and SmithedActivity just make no sense, you should rename  the existing one to smelting, and the new one to Smithing (And test that change). I'll lock smithing for a couple hours to ensure it doesnt cause any issue when i release it.

gc: you'll also need to update anything else relevant that needs to be, so everything smelting related is named appropriately and so on

src/tasks/minions/smithedActivity.ts
```diff
@@ -0,0 +1,67 @@
+import { Task, KlasaMessage } from 'klasa';
+
+import { saidYes, noOp } from '../../lib/util';
+import { Time } from '../../lib/constants';
+import { SkillsEnum } from '../../lib/skilling/types';
+import { SmithedActivityTaskOptions } from '../../lib/types/minions';
+import getUsersPerkTier from '../../lib/util/getUsersPerkTier';
+
```
gc, line 8: no empty line here

PR #252, opened 2020-05-02
create: fix cantHaveItems overriding quant to 1

Alexsuperfly:
### Description:

if an item had cantHaveItems set quantity would become 1

### Changes:

remove empty cantHaveItems fields for createables and only set quant to one if cantHaveItems is set

~~i cant test cuz something in #242 broke local hosting and i cant see why with my feeble eyes~~


Alexsuperfly:
seems both things in here are already in master

PR #253, opened 2020-05-02
groupkill emoji ID change

eX-C0n:
changed old IDs to new IDs
on disc, delete old emojis and rename new ones to match old names.


eX-C0n:
Let's do it.
Delete old emojis. Rename 'NewStart' to 'start', etc. 
Merge and deploy. 
#Esketit.

PR #257, opened 2020-05-09
organizes createables

coolbop32:
### Description:

-   _Organize createables_

### Changes:

-   _add ability charge and uncharge toxic staff and serp helm_
-   _make it so you can color serp helm with mutagens_
-   _remove level req to attach hilt to godsword also make it so you can remove hilt_

-  fix prayer not counting for total xp leader board


src/commands/Minion/create.ts
```diff
@@ -10,7 +10,7 @@ import {
 	multiplyBankQuantity
 } from '../../lib/util';
 import createReadableItemListFromBank from '../../lib/util/createReadableItemListFromTuple';
-import Createables from '../../lib/createables';
+import Createables from '../../lib/createables/createables';
```
maxdeviant, line 13: We should probably rename `creatables.ts` to `index.ts` so that this refactoring is transparent outside of the module. Then we can keep this import the same.

Alexsuperfly:
i have updated the targeted branch to the new dev branch. Can i also ask that you split your commits into single logical additions, such as having a new commit for the addition of new items to the createables file separate from the reordering commit.

PR #261, opened 2020-05-11
smithing/smelting set max quant

### LINKED ISSUE #250: Can't smelt bronze bars
The bot will always reply with "You don't have enough Copper ore" despite the amount of copper ore you actually have.

Alexsuperfly:
### Description:

sets quants based on owned items to prevent the ownage of less than max time number of items from erroring

fixes #250 

### Changes:

Smelt.ts - check owned items when setting quant
Smith.ts - check owned items when setting quant


Alexsuperfly:
closing in favor of #307

PR #262, opened 2020-05-11
add cancel command to end minion trip immediately with no rewards

Alexsuperfly:
### Description:

Adds the `cancel` command to immediately end the users minion trip with no rewards.  The user is prompted with the current trip the minion is on and then asked to confirm the cancellation before it goes through.  
Upon confirmation the minion immediately returns from their trip with no rewards and without reimbursing any items used when the trip was originally sent out.

### Changes:

the cooldown doesnt seem to work but i put it at 180 seconds since thats reasonable to avoid abuse when spam canceling trying to look for a low rolled rng time increase. Perhaps since i am the "owner" of my test bot i dont get rate limited

### DIFF
--- a/src/commands/Minion/cancel.ts
+++ b/src/commands/Minion/cancel.ts
@@ -1,0 +1,66 @@
+import { KlasaMessage, CommandStore } from 'klasa';
+
+import { BotCommand } from '../../lib/BotCommand';
+import getActivityOfUser from '../../lib/util/getActivityOfUser';
+import removeSubTasksFromActivityTask from '../../lib/util/removeSubTasksFromActivityTask';
+import { Tasks, Activity, Time } from '../../lib/constants';
+import { requiresMinion } from '../../lib/minions/decorators';
+
+const options = {
+	max: 1,
+	time: 10000,
+	errors: ['time']
+};
+
+export default class extends BotCommand {
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			oneAtTime: true,
+			cooldown: 180
+		});
+	}
+
+	@requiresMinion
+	async run(msg: KlasaMessage) {
+		const currentTask = getActivityOfUser(this.client, msg.author);
+
+		if (!currentTask) {
+			throw `${msg.author.minionName} isn't doing anything at the moment, so there's nothing to cancel.`;
+		}
+
+		if (currentTask.finishDate - Date.now() < Time.Minute * 1.5) {
+			throw `${msg.author.minionName} is already on their way back from the trip, returning in around a minute, no point in cancelling now!`;
+		}
+
+		if (currentTask.type === Activity.GroupMonsterKilling) {
+			throw `${msg.author.minionName} is in a group PVM trip, their team wouldn't like it if you left them!`;
+		}
+
+		const taskTicker =
+			currentTask.type === Activity.ClueCompletion
+				? Tasks.ClueTicker
+				: currentTask.type === Activity.MonsterKilling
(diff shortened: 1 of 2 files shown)


src/commands/Minion/cancel.ts
```diff
@@ -0,0 +1,68 @@
...
+			oneAtTime: true,
+			cooldown: 180
+		});
+	}
+
+	@requiresMinion
+	async run(msg: KlasaMessage) {
+		if (!msg.author.minionIsBusy) {
+			return msg.send(`${msg.author.minionStatus}`);
+		}
```
gc, line 27: this should be removed shouldnt it? if they are busy, we'll go forward and cancel it, if they aren't, we'll throw an error a few lines below this

src/commands/Minion/cancel.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		const currentTask = getActivityOfUser(this.client, msg.author);
+
+		if (!currentTask) {
+			throw `Your minion has no current task`;
+		}
+
```
gc, line 34: `${msg.author.minionName} isn't doing anything at the moment, so there's nothing to cancel.`

src/commands/Minion/cancel.ts
```diff
@@ -0,0 +1,68 @@
...
+		if (currentTask.finishDate - Date.now() < Time.Minute * 2) {
+			throw `You can not cancel a task with less than two minutes remaining`;
```
gc, line 36: `${msg.author.minionName} is already on their way back from the trip, and will be back in around a minute, no point in cancelling now!`

Also, I think we can change this to `Time.Minute * 1.5` just to give 30 seconds more leeway.

src/commands/Minion/cancel.ts
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		if (currentTask.type === Activity.GroupMonsterKilling) {
+			throw `You can not leave a group activity`;
+		}
```
gc, line 41: `${msg.author.minionName} is in a group PVM trip, their team wouldn't like it if you left them!`

src/commands/Minion/cancel.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		const taskTicker =
+			currentTask.type === Activity.ClueCompletion
+				? Tasks.ClueTicker
+				: currentTask.type === Activity.MonsterKilling
+				? Tasks.MonsterKillingTicker
+				: Tasks.SkillingTicker;
+
+		const cancelMsg = await msg.channel.send(
+			`${msg.author} ${msg.author.minionStatus}\n Say \`confirm\` to cancel this activity. you will recieve **NO** reimbursment of items consumed when your minion was sent out on this activity.`
```
gc, line 51: `${msg.author} ${msg.author.minionStatus}\n Say \`confirm\` if you want to call your minion back from their trip. They'll drop all their current loot to get back as fast as they can, so you won't receive any loot from this trip if you cancel it.`

src/commands/Minion/cancel.ts
```diff
@@ -0,0 +1,68 @@
...
+					_msg.author.id === msg.author.id && _msg.content.toLowerCase() === 'confirm',
+				options
+			);
+		} catch (err) {
+			return cancelMsg.edit(`Halting cancelation of minion task.`);
+		}
+
+		await removeSubTasksFromActivityTask(this.client, taskTicker, [currentTask.id]);
+
+		return msg.send(`Minion task canceled.`);
```
gc, line 66: `${msg.author.minionName}'s trip was cancelled, and they're now available.`

PR #263, opened 2020-05-13
Open: fix milestones being given out at count+1

Alexsuperfly:
### Description:

clue milestone are supposed to be given out at the milestone count, currently the code captures the users clue score before its incremented, not taking into account the clue that is currently being opened
### Changes:

moves the incrementation of the user clue score to the beginning and awaits the response before capturing it


gc:
Its already adding +1 to it to account for that?

gc:
I unintentionally sniped this, sorry

PR #266, opened 2020-05-15
create item argument to resolve items passed to commands using the usage string

### LINKED ISSUE #156: Can't drop beach boxing gloves
Can't drop beach boxing gloves because .get only returns 1 pair. Use `Items.find` instead to match the name + return the one that the user has some in their bank.

Alexsuperfly:
### Description:

creates the `item` klasa argument to resolve items by declaring a variable will be an item in the commands usage string

modifies the integer argument to only accept valid KMB numbers

closes #156 

### Changes:

- creation of src/arguments/item.ts and src/arguments/restItem.ts
- number checking in the integer argument to only accept valid KMB numbers, throw if otherwise
- refactor drop, equip, equippet, sacrifice, sell, sellto, unequip to declare their item variable as a `...item` in the usage string and use that accordingly in the command

### DIFF
--- a/src/arguments/integer.ts
+++ b/src/arguments/integer.ts
@@ -8,6 +8,19 @@ export default class extends Argument {
 
 	run(arg: string, possible: Possible, message: KlasaMessage) {
 		const { min, max } = possible;
+
+		/* 
+			remove all valid characters from the passed arg
+			if there are still some remaining, the passed arg is an invalid kmb number
+			throw and return
+			
+			this fixes "3rd age" items from being resolved as integers
+		*/
+		const checkValidKMB = arg.replace(/[0-9,_.kmb]/gi, '');
+		if (checkValidKMB.length > 0) {
+			throw message.language.get('RESOLVER_INVALID_INT', possible.name);
+		}
+
 		const number = Util.fromKMB(arg.replace(/[,_]/gi, ''));
 		if (!Number.isInteger(number)) {
 			throw message.language.get('RESOLVER_INVALID_INT', possible.name);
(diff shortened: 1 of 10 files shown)


src/lib/util/getOSItemsArray.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default function getOSItemsArray(itemName: string | number): Item[] | PartialItem[] {
+	const searchedItem = typeof itemName === 'number' ? itemName : cleanItemName(itemName);
+	const osItems = Items.filter(i => {
+		if (typeof searchedItem === 'string') {
+			return i.name.toLowerCase() === searchedItem.toLowerCase();
+		}
+		return i.id === searchedItem;
+	}).array() as Item[];
+	if (!osItems) throw `${itemName} doesnt exist.`;
```
maxdeviant, line 14: Shouldn't this be `!osItems.length`?

Empty arrays are truthy values, so `![]` is `false`. This condition will never be hit as `osItems` (if we're to believe the type signatures) will never be `null` or `undefined`.

Alexsuperfly: is that what filter does when it doesnt find anything? returns empty array instead of undefined?

maxdeviant: Yes, `filter` will always return an empty array if nothing matches the predicate.

I glanced at the [implementation](https://github.com/gc/oldschooljs/blob/master/src/structures/Collection.ts#L254-L262) and it will always return a valid collection type, never `null` or `undefined`.

Alexsuperfly: thanks for the heads up, pushed the change now

PR #274, opened 2020-05-29
Update sellto.ts so that qty is optional

BrandonEnglish848:
If the quantity being sold is not specified in the original command, the bot will find the item's id and search the user's bank and then set the quantity being sold to that number. 
I also normalized a bit of your layout discrepancies between the if statements.
✅ I have tested my changes


src/commands/Minion/sellto.ts
```diff
@@ -30,29 +30,44 @@ export default class extends BotCommand {
...
+		}
+		if (buyerMember.user.isIronman) {
+			throw `Iron players can't be sold items.`;
+		}
+		if (buyerMember.user.id === msg.author.id) {
+			throw `You can't trade yourself.`;
+		}
+		if (buyerMember.user.bot) {
+			throw `You can't trade a bot.`;
+		}
```
gc, line 46: Why did you change all these?

src/commands/Minion/sellto.ts
```diff
@@ -30,29 +30,44 @@ export default class extends BotCommand {
...
 		if (buyerMember.user.isBusy) {
 			throw `That user is busy right now.`;
 		}
-
```
gc, line 42: Why remove a random empty line?

BrandonEnglish848: Because ESLint told me to.

src/commands/Minion/sellto.ts
BrandonEnglish848: Because ESLint told me to.

BrandonEnglish848:
I have not tested this, and will be messaging in the dev channel

BrandonEnglish848:
### Description:

The Sellto command can have more versatility if the users have already agreed on a price, and are going to sell entire stacks of items. It's much easier for one user to sell a whole stack rather than selling small chucks or have to type out specific quantities.

### Changes:

* Generally Utilized the Sell and Sellto commands
* Changed the run parameters to where quantity is optional.
* Did NOT change the order in which the parameters are passed into the run command so that users who are already familiar with the command will not be confused with the change.

- [✔️] I have tested all my changes thoroughly.

Lajnux:
We going over to slashcommands and a system to make qty optional is now implemented.

PR #293, opened 2020-06-02
Fletching fix

coolbop32:
### Description:

-   _fix flecthing arrow shafts (wyatt)_

### Changes:

-   _add runecrafting cape to cl capes_


src/extendables/User/Minion.ts
```diff
@@ -258,7 +258,7 @@ export default class extends Extendable {
 			case Activity.Fletching: {
 				const data = currentTask as FletchingActivityTaskOptions;
 				const fletchable = Fletching.Fletchables.find(
-					item => item.id === data.fletchableID
+					item => item.name === data.fletchableName
```
Alexsuperfly, line 261: since the activity already has the name, you dont need to find it in fletchable and can instead just straight use the name in the return message

coolbop32: fixed

Alexsuperfly:
needs rebase

PR #301, opened 2020-06-03
Bank Sale For Running Sell Command On Array

BrandonEnglish848:
This is a rough draft for a command that will take an array of arrays like so:
```+bankSale [["Rune battleaxe",5], ["Feather",""],["Bass",100000]]```
It will then process each underlying array by passing the Item name and quantity into the functions I mirrored from the Sell command. 
I intended for it to only be able to process a set number of items by setting the Limit variable. This could be dynamically changed based on the user's patron status in order to allow for more lines to be processed in this command.
If the user does not specify a quantity for the item being sold, it will pull the maximum amount that user has in their bank, and set that as the quantity. If the user does not have enough of an item in their bank, it throws the item into the invalidSalesItems array so that it won't be processed.

### DIFF
--- a/src/commands/Minion/bankSale.ts
+++ b/src/commands/Minion/bankSale.ts
@@ -1,0 +1,134 @@
+import { KlasaMessage, CommandStore } from 'klasa';
+import { Util } from 'oldschooljs';
+
+import { BotCommand } from '../../lib/BotCommand';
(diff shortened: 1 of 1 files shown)


gc:
I won't be adding this, at the current time, sorry.

PR #306, opened 2020-06-04
Fix Barbarian Fishing Requirements and Rates

ard35:
### Description:

- Previously, leaping sturgeons had a 48 agility requirement. This has been fixed to 45. 
- Previously, the minion always obtained a leaping trout even though there is supposed to be a chance for a failed roll. This roll for a failed catch has been added. Exp rates are now normalized thanks to this addition.

### Changes:

-  Add a chance to NOT catch a fish during barbarian fishing.
- Changed the message received when starting a barbarian fishing run to reflect that a player now attempts a certain amount of fish.
- Changed the message received when minion finishes a barb fishing run to reflect that a player failed a fish an amount of failed/attempts.

### DIFF
--- a/src/commands/Minion/fish.ts
+++ b/src/commands/Minion/fish.ts
@@ -74,14 +74,30 @@ export default class extends BotCommand {
 			quantity = Math.floor(msg.author.maxTripLength / scaledTimePerFish);
 		}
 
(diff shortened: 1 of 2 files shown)


ard35:
Test

PR #334, opened 2020-06-19
create eatables

Alexsuperfly:
### Description:

eatables file to hold all food items that can heal

### Changes:

creation of eatables.ts
removal of healAmount from the cookables array
let wt command read from the eatables array


BrandonEnglish848:
Could we add another parameter to the eatables so that we could include status effects?

PR #348, opened 2020-06-25
Fix to Blowpipe in Filterables

osrsolympus:
### Description:

Blowpipe filter name was incorrect (my fault, sorry). Fix from (uncharged) to (empty)


gc:
Thanks! I fixed this locally already in a branch so I'll just include this in my next PR.

PR #356, opened 2020-07-07
groupKill: fix kc always 1

ard35:
### Description:

-   Fixes the bug where each user only gains 1 kc after a mass despite getting multiple kills/loot drops.

### Changes:

-   Initializes all users to begin with kcAmounts = 0 at the start of the mass.
-   Increments kcAmounts by 1 for a user each time that user is randomly selected for a drop.

**-   [X] I have tested all my changes thoroughly.**


src/tasks/minions/groupMonsterActivity.ts
```diff
@@ -16,14 +16,16 @@ export default class extends Task {
...
+			kcAmounts[users[j]] = 0;
+		}
+
 		for (let i = 0; i < quantity; i++) {
 			const loot = monster.table.kill(1);
 			const userWhoGetsLoot = randomItemFromArray(users);
 			const currentLoot = teamsLoot[userWhoGetsLoot];
 			teamsLoot[userWhoGetsLoot] = addBankToBank(currentLoot ?? {}, loot);
-			kcAmounts[userWhoGetsLoot] = Boolean(kcAmounts[userWhoGetsLoot])
-				? kcAmounts[userWhoGetsLoot]++
```
Alexsuperfly, line 25: could you try `? ++kcAmounts[userWhoGetsLoot]` or alternatively `? kcAmounts[userWhoGetsLoot] + 1`
instead of the current commit.  The problem seems to be the post-increment operator returning the original value instead of the modified one

ard35: Both worked but I changed it to `? ++kcAmounts[userWhoGetsLoot]` as it was cleaner.

PR #363, opened 2020-07-10
DO NOT PUSH -> Helpful Dev Commands

BrandonEnglish848:
#363 Description:

THESE SHOULD **NEVER** BE PUSHED TO THE STANDARD USER COMMANDS
These are to help with local instance testing and are ONLY for *development purposes*.

#363  Changes:

enhance.ts:
Will process a number and a string as the amount of exp to increase / decrease and the skill that will be changed.
Helpful when the skill / rewards being tested need specific level requirements for your testing bot.

spawn.ts:
Will process a number and a string as the number of items (defaults to 1 if none passed in) and the name of the item that will be generated.
Helpful when you need to test specific functions in your bot that require you to have an item in your bank.

empty.ts:
Issuing this command will clear the user's bank with no confirmation prompt.
Helpful when you need to "reset" the bot's inventory when testing that your requirements for your activity are met / unmet.

✅ These were tested in the ../src/commands/Minion folder, but for dev purposes I created the PR in the OWNER folder.


BrandonEnglish848:
Please confirm that these should be pushed into the Owner Folder. I have thoroughly tested these commands in the Minions folder, but I was unsure if I should open the PR in that folder because I assumed it would have a greater chance to get accidentally pushed if the files were in there instead of the Owner folder.

gc:
I'm handling the dev-commands in a different way,  and they've been added by now.

PR #375, opened 2020-07-12
Openable changes

Lajnux:
### Description:

Openable changes, to include the keys mentioned below.

### Changes:
	'Crystal key',
	'Enhanced crystal key',
	'Sinister key',
	'Seed pack',
	'Muddy key',
	'Bronze key',
	'Iron key',
	'Steel key',
	'Silver key',
	'Grubby key',
	`Larran's key`,
	'Brimstone key',
	'Ogre coffin key'
along with filterable for openables and changed open to handle all new keys. 

image
image
The open several keys at once loop have been removed.


BrandonEnglish848:
Note on the changes being submitted:
The open command will list all possible openables as it does normally (including the newly submitted ones in this pr)
The open command will roll as many times as there are for the quantity of the item in your bank. (See how all he had to do was “,open sinister” and it returned all 100,000 rolls. Limits could be placed based on patron tier if desired)
The Larran’s Key is tagged with the ‘’ for markup usage, but could be changed to “”. It works as-is and would solely be a personal preference to change it.

PR #385, opened 2020-07-13
Multiple open buy

Alexsuperfly:
## Description:

allows users to specify a quantity of an item to open or buy

### open.ts 
- add quantity parameter defaulted to 1 to open quantity amount of the item to cut down on spam
- show bank images on all item opens
- creation of clueOpen(), osjsOpenablesOpen(), and botOpenablesOpen() to handle opening items of each type.
### openables.ts
- removal of Mystery box since its in osjs Openables

### buy.ts
 - add quantity parameter defaulted to 1 to buy quantity amount of the item to cut down on spam

### buyables.ts 
- set every item purchased to a quantity of 1 to allow more granular purchasing with new quantity parameter in buy.ts


eX-C0n:
:this: doing God's work

src/commands/Minion/open.ts
```diff
@@ -53,36 +54,52 @@ export default class extends BotCommand {
...
 		}
+		return this.clueOpen(msg, quantity, clue);
+	}
 
-		const hasCasket = await msg.author.hasItem(clueTier.id);
-		if (!hasCasket) {
-			throw `You don't have any ${
+	async clueOpen(msg: KlasaMessage, quantity: number, clueTier: ClueTier) {
+		const userBank = msg.author.settings.get(UserSettings.Bank);
+		if ((userBank[clueTier.id] ?? 0) < quantity) {
```
gc, line 77: use `msg.author.numItemsInBankSync(clueTier.id) < quantity`, and the same for any similar checks

src/commands/Minion/open.ts
```diff
@@ -142,36 +158,76 @@ export default class extends BotCommand {
...
+		const task = this.client.tasks.get('bankImage')!;
+		const image = await task.generateBankImage(
+			loot.values(),
+			`You opened ${quantity} ${botOpenable.name}`,
+			true,
+			{ showNewCL: 1 },
+			msg.author
 		);
+
+		return msg.send(new MessageAttachment(image, 'osbot.png'));
```
gc, line 231: use `return msg.channel.sendBankImage` instead of these 3 things, you can find examples of using it in the bank cmd: https://github.com/oldschoolgg/oldschoolbot/blob/master/src/commands/Minion/bank.ts

(and to anywhere else in this file that does this)

src/commands/Minion/open.ts
Alexsuperfly: whats wrong with this, other than the manual check if its not null

PR #392, opened 2020-07-16
Decant: Only show left over doses if there are any

jamosaur:
### Description:

-  Only shows the `with Saradomin Brew(1) left over` if there is a left over dose. Hides the message otherwise

### Changes:

### DIFF
--- a/src/commands/Minion/decant.ts
+++ b/src/commands/Minion/decant.ts
@@ -50,7 +50,7 @@ export default class extends BotCommand {
 		newBank = addItemToBank(newBank, potionTiers[3], totalDecantedPotions);
 		if (dosesToGiveBack > 0) {
 			newBank = addItemToBank(newBank, potionTiers[dosesToGiveBack - 1]);
-			leftOverMessage = `, with a ${potionToDecant.name}(${dosesToGiveBack}) left over.`
+			leftOverMessage = `, with a ${potionToDecant.name}(${dosesToGiveBack}) left over.`;
 		}
 
 		await msg.author.settings.update(UserSettings.Bank, newBank);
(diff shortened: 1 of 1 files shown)


src/commands/Minion/decant.ts
```diff
@@ -1,9 +1,9 @@
-import { KlasaMessage, CommandStore } from 'klasa';
+import { CommandStore, KlasaMessage } from 'klasa';
```
Alexsuperfly, line 1: revert these

src/commands/Minion/decant.ts
```diff
@@ -54,8 +54,13 @@ export default class extends BotCommand {
 
 		await msg.author.settings.update(UserSettings.Bank, newBank);
 
+		const leftOverMessage =
```
Alexsuperfly, line 57: i think this would look better if placed inside the if block that already determines if there are doses to give back.

src/commands/Minion/decant.ts
```diff
@@ -1,9 +1,8 @@
 import { KlasaMessage, CommandStore } from 'klasa';
-
```
Alexsuperfly, line 2: revert this

src/commands/Minion/decant.ts
```diff
@@ -1,9 +1,8 @@
...
 import { BotCommand } from '../../lib/BotCommand';
 import { UserSettings } from '../../lib/settings/types/UserSettings';
 import Potions from '../../lib/minions/data/potions';
-import { stringMatches, addItemToBank } from '../../lib/util';
```
Alexsuperfly, line 6: revert this

PR #393, opened 2020-07-16
drop: add --cf and --confirm flags

### LINKED ISSUE #378: Add --cf to drop command.
Currently, we have --cf to sacrifice, sell and sellto. Adding it to drop would make it more consistent.

jamosaur:
### Description:

-   Add --cf and --confirm to drop command

### Changes:

### DIFF
--- a/src/commands/Minion/drop.ts
+++ b/src/commands/Minion/drop.ts
@@ -34,17 +34,19 @@ export default class extends BotCommand {
 			throw `You don't have ${quantity}x ${osItem.name}.`;
 		}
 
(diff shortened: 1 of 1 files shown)


src/commands/Minion/drop.ts
```diff
@@ -1,4 +1,4 @@
-import { KlasaMessage, CommandStore } from 'klasa';
+import { CommandStore, KlasaMessage } from 'klasa';
```
Alexsuperfly, line 1: revert this is unnecessary diff

PR #400, opened 2020-07-18
Dry calc

Lajnux:
### Description:

Made a dryCalc command for fun commands

### Changes:

New file in fun commands so that drycalc can be done for items/drops. example: +drycalc 5k 10k
Would give you the chance of a dragon warhammer in 10k kills.


src/commands/OSRS_Fun/dryCalc.ts
```diff
@@ -0,0 +1,45 @@
...
+
+	async run(msg: KlasaMessage, [dropRate = 1, rolls = 1]: [number, number]) {
+		if (dropRate > 100000000 || rolls > 100000000) {
+			throw "I can't process a number higher than 100 million! enneUni has already claimed the only uncut onyx!";
+		}
+		if (dropRate < 1 || rolls < 1) {
+			throw "I can't calculate with zeros nor negative numbers. Do more rolls or update the drop values.";
+		}
+		const noDropChance = Math.pow(1 - 1 / dropRate, rolls);
+		const dropChance = 100 * (1 - noDropChance);
```
Alexsuperfly, line 26: i dont see the reason to include this, its just the inverse of how dry you are, and this command is called `drystreak`

Lajnux: it was just the way the oldschool bot drycalc was showing so I figured that's what most ppl would prefer. For some the inverse is obvious but not for everyone so wanted to make sure that it includes a greater auidience.

src/commands/OSRS_Fun/dryCalc.ts
```diff
@@ -0,0 +1,45 @@
...
+		});
+	}
+
+	async run(msg: KlasaMessage, [dropRate = 1, rolls = 1]: [number, number]) {
+		if (dropRate > 100000000 || rolls > 100000000) {
+			throw "I can't process a number higher than 100 million! enneUni has already claimed the only uncut onyx!";
+		}
+		if (dropRate < 1 || rolls < 1) {
+			throw "I can't calculate with zeros nor negative numbers. Do more rolls or update the drop values.";
+		}
```
gc, line 24: Remove both of these checks, and instead edit the `usage` option to this: `usage: '<dropRate:int{1,1000000}> <rolls:int{1,1000000}>',` which will handle those checks for you, after doing this also remove the `=1` default assignments

src/commands/OSRS_Fun/dryCalc.ts
```diff
@@ -0,0 +1,45 @@
+import { Command, CommandStore, KlasaMessage } from 'klasa';
+
+/* 
+  Witty Switch Comments Pulled From:
+  https://oldschool.runescape.wiki/w/Calculator:Dry_calc
+*/
```
gc, line 6: remove this comment

PR #413, opened 2020-07-21
BSO: Return the correct ID's from mystery boxes

jamosaur:
If a clue scroll or casket is rolled, it will get the correct ID for it based on the `ClueTier` that it should belong to.

If a duplicate item is rolled, it will try to find the non-duplicate version and return that, if not it will re-roll


gc:
This is no longer an issue because the bot only gives tradeable items now - and therefore no clues.

Issue #436, opened 2020-07-27
Show custom items in bank


gc:
## 🚀 Feature Proposal

e.g. `+bank Fire Rune, Earth Rune, Logs` shows all of those 3 items you have in your bank. Could also do support for runelite bank tags.

PR #441, opened 2020-07-29
Command(Bank): Add support for passing multiple item names, runelite bank tags

gc:
### Description:

-   Add support for passing multiple item names, for example: `+bank Shark, Tuna, Bass`. You can also pass item IDs, runelite bank tags work too. `getOSItem` was modified to handle strings more intelligently. It will try to turn the string into a number, if it works it works, if it doesnt then we consider it an item name.

### Changes:

-  Memoize getOSItem
- Tests


gc:
Closes #436

PR #446, opened 2020-07-30
Debug commands

Alexsuperfly:
### Description:

creation of more testing commands

### Changes:

update `spawn` to use the new `item` argument
add `setxp` to set the exp of any skill to a particular value
add `setgp` to set the cash stack to any value

### DIFF
--- a/src/arguments/item.ts
+++ b/src/arguments/item.ts
@@ -8,7 +8,10 @@ import { stringMatches } from '../lib/util';
 export default class extends Argument {
 	async run(itemName: string) {
 		// guarantee all characters are numbers
-		if (parseInt(itemName).toString().length === itemName.length) {
+		if (
+			!isNaN(parseInt(itemName)) &&
+			parseInt(itemName).toString().length === itemName.length
+		) {
 			return [getOSItem(parseInt(itemName))];
 		}
 
--- a/src/commands/Fun/mypets.js
+++ b/src/commands/Fun/mypets.js
@@ -1,6 +1,5 @@
 const { Command } = require('klasa');
 const pets = require('../../../data/pets');
-const { queuedMessageSend } = require('../../lib/util');
 
 module.exports = class extends Command {
 	constructor(...args) {
@@ -23,6 +22,6 @@ module.exports = class extends Command {
 			return formatted.push(`${pet.emoji} ${pet.name}: ${userPets[id]}`);
 		});
 
-		return queuedMessageSend(this.client, msg.channel.id, formatted.join('\n'));
+		return msg.send(formatted.join('\n'), { split: true });
 	}
 };
--- a/src/commands/Fun/mypets.js
+++ b/src/commands/Fun/mypets.js
@@ -1,5 +1,6 @@
 const { Command } = require('klasa');
 const pets = require('../../../data/pets');
+const { queuedMessageSend } = require('../../lib/util');
 
 module.exports = class extends Command {
 	constructor(...args) {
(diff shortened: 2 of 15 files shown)


src/arguments/item.ts
```diff
@@ -8,7 +8,10 @@ import { stringMatches } from '../lib/util';
 export default class extends Argument {
 	async run(itemName: string) {
 		// guarantee all characters are numbers
-		if (parseInt(itemName).toString().length === itemName.length) {
+		if (
+			!isNaN(parseInt(itemName)) &&
+			parseInt(itemName).toString().length === itemName.length
+		) {
```
gc, line 14: I made changes to `getOSItem` already which allow it to accept item IDs as strings, maybe makes this unnecessary?

src/commands/Fun/mypets.js
```diff
@@ -22,6 +23,6 @@ module.exports = class extends Command {
 			return formatted.push(`${pet.emoji} ${pet.name}: ${userPets[id]}`);
 		});
 
-		return msg.send(formatted.join('\n'));
+		return queuedMessageSend(this.client, msg.channel.id, formatted.join('\n'));
```
gc, line 26: whats the point of doing this?

src/commands/Testing/spawn.ts
```diff
@@ -1,20 +1,23 @@
...
-			usage: '[number:int{1,100}] <itemname:...string>',
+			usage: '[qty:integer{1,1000000}] (item:...item)',
 			usageDelim: ' ',
 			oneAtTime: true
 		});
 		this.enabled = !this.client.production;
 	}
 
-	async run(msg: KlasaMessage, [qtyOrName, itemNamePossible]: [number, number | undefined]) {
+	async run(msg: KlasaMessage, [qty = 1, item]: [number, Item[]]) {
```
gc, line 20: if `item` is always an array of items, this should be called `items` or `itemArray`

src/commands/Testing/spawn.ts
```diff
@@ -24,20 +27,28 @@ export default class extends BotCommand {
 			return;
 		}
 
-		const qty = typeof qtyOrName === 'number' ? qtyOrName : 1;
-		const itemName =
-			typeof qtyOrName === 'string' ? qtyOrName : itemNamePossible ?? 'Bunny feet';
+		if (msg.flagArgs.all) {
+			const items: Bank = {};
+			item.forEach(i => (items[i.id] = qty));
```
gc, line 32: theres pretty much no case where we want to use `forEach`, use a `for of` loop.

src/commands/Fun/mypets.js
Alexsuperfly: mylife has too many pets so it errors when trying to send the message without splitting

gc: i think it would be better to do just `, { split: true }` here, `queuedMessageSend` is used only for the minion tasks finishing as an extra precaution to not overstep ratelimits, hugely so after a reboot when 500-1500 trips could be finishing at once.

src/arguments/item.ts
Alexsuperfly: this makes sure the passed string is actually a number by checking that when parsed its length is the same, the isNaN catches the case that was found yesterday where you could pass `cod` return `NaN` and pass the length check.  without the length check `3rd` age would be interpreted as a number

src/commands/Testing/spawn.ts
Alexsuperfly: i just wanted a one liner is all

PR #451, opened 2020-08-01
General updates

gc:
### Description:

-   _Write description here_

### Changes:

-   _List of changes here_


src/tasks/minions/agilityActivity.ts
```diff
@@ -63,6 +73,17 @@ export default class extends Task {
 			[markOfGrace]: totalMarks
 		};
 
+		if (course.id === 6) {
+			const currentLapCount = user.settings.get(UserSettings.LapsScores)[course.id] ?? 0;
```
Alexsuperfly, line 77: dont think you need the `?? 0` since if it is 6 then youve already set the count in the update above this

PR #482, opened 2020-08-06
revs-bso

Lajnux:
### Description:

adding revs to BSO

### Changes:

adding revs to BSO


Lajnux:
The revenants in osrs have been reworked drop rates and this PR doesn't fit BSO anymore since custom monsters and bosses can also be made. Revenants will come in future together with Slayer most likely.

PR #488, opened 2020-08-08
Prevent Mass Crashes

ard35:
### Description:

-   Prevents mass crashing by removing players who do not meet the requirements for a mass.

### Changes:

- Allowed removing users to prevent crashes.
- Reworked the food check into the checkReqs function.

### DIFF
--- a/src/commands/Minion/groupkill.ts
+++ b/src/commands/Minion/groupkill.ts
@@ -39,28 +39,85 @@ export default class extends BotCommand {
 		return [quantity, duration, perKillTime];
 	}
 
-	checkReqs(users: KlasaUser[], monster: KillableMonster, quantity: number) {
+	async checkReqs(users: KlasaUser[], monster: KillableMonster, quantity: number) {
 		// Check if every user has the requirements for this monster.
+		const removedUsers: string[] = [];
+		const removedReasonStr: string[] = [];
+		const numberOfUsers = users.length;
+		let foodId = 0;
+		let amountOfFoodNeeded = 0;
+
 		for (const user of users) {
+			let checkReqsState = true;
 			if (!user.hasMinion) {
-				throw `${user.username} doesn't have a minion, so they can't join!`;
+				checkReqsState = false;
+				removedReasonStr.push(
+					`${user.username} doesn't have a minion, so they can't join!`
+				);
 			}
 
 			if (user.minionIsBusy) {
-				throw `${user.username} is busy right now and can't join!`;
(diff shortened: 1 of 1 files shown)


src/commands/Minion/groupkill.ts
```diff
@@ -39,28 +39,84 @@ export default class extends BotCommand {
...
+								`${user.username} doesn't have enough food to kill ${
+									monster.name
+								} in this sized mass! You need enough food to heal atleast ${healAmountNeeded} HP (${healAmountNeeded /
+									quantity} per kill) You can use these food items: ${Eatables.map(
+									i => i.name
+								).join(', ')}.`
+							);
+						}
+
+						foodId = food.id;
```
Alexsuperfly, line 96: shouldnt these be one level higher, you dont care about the food that they dont have

src/commands/Minion/groupkill.ts
```diff
@@ -39,28 +39,84 @@ export default class extends BotCommand {
...
 			}
 
-			if (1 > 2 && !hasEnoughFoodForMonster(monster, user, quantity)) {
-				throw `${user.username} doesn't have enough food.`;
+			if (1 > 2) {
+				let [healAmountNeeded] = calculateMonsterFood(monster, user);
+				healAmountNeeded = Math.ceil(healAmountNeeded / numberOfUsers);
+
+				for (const food of Eatables) {
+					const amountNeeded = ceil(healAmountNeeded / food.healAmount!) * quantity;
```
Alexsuperfly, line 82: dont need the `!` on healAmount

src/commands/Minion/groupkill.ts
```diff
@@ -39,28 +39,84 @@ export default class extends BotCommand {
...
+				}
 			}
 
-			if (1 > 2 && !hasEnoughFoodForMonster(monster, user, quantity)) {
-				throw `${user.username} doesn't have enough food.`;
+			if (!checkReqsState) {
+				users.splice(users.indexOf(user), 1);
+				user.send(removedReasonStr.join(', '));
+			} else {
+				await user.removeItemFromBank(foodId, amountOfFoodNeeded);
```
Alexsuperfly, line 109: the food removal cant be here because this function is used as a preliminary check for the party leader, meaning they would have food taken out for that check and the final one.  That prelim check could maybe be removed and if the leader fails the final check they can get removed from their own party without crashing it, or just dont move the food removal checks.

Lajnux:
Not gonna be further maintained.

PR #489, opened 2020-08-08
Dev

### LINKED ISSUE #458: +gearstats mage unexpected error
image

Getting a reply from Old School Bot that an unexpected error occured when trying to check my mage gear stats.

### LINKED ISSUE #528: 3rd age pick / axe not giving bonuses
On BSO,

When going to mine, cut logs, etc. 3rd age pick and axe are no longer giving correct bonuses to activity. Watched Veras unequip and equip axe while re-attempting sulliusceps, and then tg03 attempt coal several times.

That's all. Cheers - thanks for yalls work on the bot.

Alexsuperfly:
### Description:

- buyables: add Ball of wool
- mining: add Gem rocks
- mining: set iron and gold to correct xp/ore
- craft: crush semiprecious gems when cutting
- item: use var for parsed parameter - refactor to save 2 `parseInt` executions
- getOSItem: fix parsing non numbers as numbers - fix that prevents item names that start with numbers from being treated as numbers
- cleanString: default to empty str to avoid problems with undefined
- alch throw useful error messages - show when an item is untradeable or not found correctly
- alch: allow runes to be alched
- alchingActivity: allow bryophytas staff in any gear set
- calculateMonsterFood: create specialReducers for food reduction - adds a 17.5% reduction when wearing an ely and creates a framework to add other reductions like this to
- gearstats: magic -> mage

fixes #458 
fixes #528

### DIFF
--- a/src/arguments/item.ts
+++ b/src/arguments/item.ts
@@ -8,11 +8,9 @@ import { stringMatches } from '../lib/util';
 export default class extends Argument {
 	async run(itemName: string) {
 		// guarantee all characters are numbers
-		if (
-			!isNaN(parseInt(itemName)) &&
-			parseInt(itemName).toString().length === itemName.length
-		) {
-			return [getOSItem(parseInt(itemName))];
+		const parsed = parseInt(itemName);
+		if (!isNaN(parsed) && parsed.toString().length === itemName.length) {
+			return [getOSItem(parsed)];
 		}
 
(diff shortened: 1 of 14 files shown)


Alexsuperfly:
as ive told you at least 3 times now its not a bug, its a feature.  If you could just tell me why its wrong i would consider changing it

Alexsuperfly:
The dev branch became depricated a while ago, i have split all of the PRs left in it back out individually and they now point at master

PR #504, opened 2020-08-13
bankstand(BSO)

Lajnux:
### Description:

Added a new command to bankstand at either a random bank or set bank

### Changes:

Added a new command to bankstand at either a random bank or set bank, with activity which gives mystery boxes or scams you from getting any mystery boxes.


gc:
As I told you in discord, I'm not adding this

PR #513, opened 2020-08-16
Add smithing and crafting bank filters, changed fletching filter

Sethmare:
### Description:

Added bank filters for smithing and crafting, made fletching and crafting filters pull from fletching and crafting skill item lists respectively

### Changes:

-   Edited "filterables" to add a smithing filter with ores, bars, and misc smithing items, created crafting filter to draw things from the crafting skill, and did the same for the fletching filter. Removed some redundancies from the skilling filter.


Alexsuperfly:
i would rather see these extracted in some way from the required items field of the skill, that would allow it to be up to date with any future additions to the it without the need to remember to add a new item to the filter

Sethmare:
Makes total sense. Definitely above my knowledge level, but I'm willing to help with this if somebody could teach me

Sethmare:
I believe I've made the changes you wanted, as well as doing something similar for fletching @Alexsuperfly

src/lib/filterables.ts
```diff
@@ -2,6 +2,97 @@ import resolveItems from './util/resolveItems';
...
+	`Diamond`,
+	`Dragonstone`,
+	`Onyx`,
+	`Zenyte`,
+	`Uncut zenyte`,
+	`Zenyte shard`
+]);
+
+const craftingItems = Crafting.Craftables.flatMap(item =>
+	Object.keys(item.inputItems).map(key => parseInt(key))
```
Alexsuperfly, line 90: concat the output item from the craft here so that its also tracked in the final array
```suggestion
	Object.keys(item.inputItems).map(key => parseInt(key)).concat(item.id)
```

unless you specifically think that finished items shouldnt be in this filter which i suppose can be valid idk

src/lib/filterables.ts
```diff
@@ -168,134 +259,15 @@ const bones = resolveItems([
...
-	'Uncut onyx',
-	'Sapphire',
-	'Emerald',
-	'Ruby',
-	'Diamond',
-	'Dragonstone',
-	'Onyx'
-]);
+const fletchingItems = Fletching.Fletchables.flatMap(item =>
+	Object.keys(item.inputItems).map(key => parseInt(key))
```
Alexsuperfly, line 263:
```suggestion
	Object.keys(item.inputItems).map(key => parseInt(key)).concat(item.id)
```
same as crafting

src/lib/filterables.ts
Alexsuperfly: after second thought having output items in this filter doesnt have much use

src/lib/filterables.ts
```diff
@@ -2,6 +2,97 @@ import resolveItems from './util/resolveItems';
...
+	`Blessed Spirit Shield`,
+	`Smouldering stone`,
+	`Dragon metal lump`,
+	`Draconic visage`,
+	`Skeletal visage`,
+	`Wyvern visage`
+]);
+
+const gems = resolveItems([
+	`Uncut opal`,
```
Alexsuperfly, line 65: the craftingItemsSet array already has the uncut versions in it, so you only need to have the cut versions here

Alexsuperfly:
remove your extraneous commits and rebase

Sethmare:
@Alexsuperfly Should be good now

PR #516, opened 2020-08-16
sync user settings before numItemsInBankSync call

### LINKED ISSUE #497: Master clue dissapeared
Looks like my master clue dissapeared after i completed a clue. Not able to open a casket. 

osrs disc bot master clue
## 🐛 Bug Report

Alexsuperfly:
### Description:

sync user settings before `numItemsInBankSync` is called for some commands because there are cases where a minion can return from a trip but doesnt have the bank synced fast enough to realize the new items are there for the next command

fixes #497
### Changes:


gc:
the `sync` is for "synchronous", intended for use where you dont want to sync (which requres an asynchronous function). we could perhaps modify `numItemsInBankSync` to conditionally return a promise based on a  overload. I.e if you pass `true`, it returns a promise, otherwise it doesnt -  sounds like it would work to me.

Alexsuperfly:
ok that makes sense, the terminalogy of `settings.sync` and `numItemsInBankSync` got me confused i guess.  I thought that since it had Sync in the name it would sync the data then access it

PR #521, opened 2020-08-18
shieldRightHalf-exploit-fix

Lajnux:
### Description:

The shield right half was put at a price making sacrificing stuff devalued because you can buy the halfs from the store for under sacrifice value. The other problem with it being so close to +sell value is that if GE one day gets a increased shield half price, the bot GP would be infinit. 

### Changes:

Doubled the GP cost for shield right half from 750k (the in game buy price) to 1.5M, this also makes it more balanced considering all other buyable items in the +buy ''store''

### DIFF
--- a/src/lib/buyables.ts
+++ b/src/lib/buyables.ts
@@ -193,7 +193,7 @@ const Buyables: Buyable[] = [
 			'Shield right half': 1
 		}),
 		qpRequired: 111,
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
this is mirroring the live game as intended, it cannot be price manipulated since it is only acquired through a shop.

PR #523, opened 2020-08-19
Fix barrows ID & inquisitor name

Ravestrike:
### Description:

-   Barrows role id is different than the one on discord, this changes it to its correct one.  Changed inquisitor's role to its correct name. 

### Changes:


src/monitors/rareRoles.ts
```diff
@@ -12,14 +12,14 @@ const rareRoles: [string, number, string][] = [
 	['705987772372221984', 7500, 'Void'],
 	['688563333464457304', 10_000, 'Mithril'],
 	['688563389185917072', 15_000, 'Adamant'],
-	['705988547202515016', 20_000, "Inquisitors's"],
+	['705988547202515016', 20_000, "Inquisitor's"],
```
Alexsuperfly, line 15: i would rather this be singular not possessive, nothing else is possessive and only one is plural but thats because `barrows` describes all the brothers

PR #552, opened 2020-08-25
[BSO] Changed clue opening extra rolls

kevinsmendes:
# Description:

-   People on BSO were complaining about how their large openings were being 'messed up' because you were better opening 1 at time to avoid opening 500 and rolling on x1 multiplier.

### Changes:

- Changed how extra rolls work for clue openings. Currently, you either get 1x, 2x or 3x the total amount you opened. That creates the possibility that someone may lose all their extra rolls (or gain a lot). Most people on BSO currently are doing small openings to prevent losing too much. This change make it fair, as a number between the opened one and 3 times that is chosen. Realistically, someone can still get 0 extra rolls, but that is much more rare.

https://images-ext-2.discordapp.net/external/2pyBGC4YMPbdtLoMtjBH6hb49VmBLUuVMeo4TRGZ6Ms/https/i.imgur.com/reT0SmE.png?width=597&height=910
image

image

### DIFF
--- a/src/commands/Minion/open.ts
+++ b/src/commands/Minion/open.ts
@@ -141,8 +141,6 @@ export default class extends BotCommand {
 			msg.author.incrementMonsterScore(MIMIC_MONSTER_ID, mimicNumber);
 		}
 
-		console.log(extraClueRolls);
-
 		return msg.channel.sendBankImage({
 			bank: loot,
 			content: `You have completed ${nthCasket} ${clueTier.name.toLowerCase()} Treasure Trails.${
--- a/src/commands/Minion/open.ts
+++ b/src/commands/Minion/open.ts
@@ -142,7 +142,9 @@ export default class extends BotCommand {
 			bank: loot,
 			content: `You have completed ${nthCasket} ${clueTier.name.toLowerCase()} Treasure Trails.${
 				newQuantity > quantity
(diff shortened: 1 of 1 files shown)


src/commands/Minion/open.ts
```diff
@@ -74,7 +74,8 @@ export default class extends BotCommand {
 
 		await msg.author.removeItemFromBank(clueTier.id, quantity);
 
-		let loot = clueTier.table.open(quantity * rand(1, 3));
+		const newQuantity = Math.floor(Math.random() * (quantity * 3 - quantity + 1) + quantity);
```
Lajnux, line 77: to make this accurate to what it used to be consider using a for loop for each casket case. 

something along the lines of 
for (let i = 0; i < quantity; i++ {
  //Roll 1 to 3
  //Add and track extra rolls
 //Combine loot for each cycle.
}

src/commands/Minion/open.ts
```diff
@@ -135,9 +141,17 @@ export default class extends BotCommand {
 			msg.author.incrementMonsterScore(MIMIC_MONSTER_ID, mimicNumber);
 		}
 
+		console.log(extraClueRolls);
```
gc, line 144: remove this log

PR #658, opened 2020-09-20
[BSO] Moves Hween/Phat set from TMB to their respective tables

kevinsmendes:
### Description:

-   Removes Halloween mask sets and Partyhat sets from tradeable mystery boxes and adds it to holiday mystery boxes and crackers, respectively. They have the same odds

### Changes:

-   Adds `Halloween mask set` to `HolidayItems` and `Partyhat set` to `PartyhatTable`.


gc (changes requested):
I dont want entire sets to be dropped, regardless of chance.

PR #668, opened 2020-09-22
[BSO] Mass watcher

kevinsmendes:
### Description:

-   BSO Mass Channel is cancer

### Changes:

-   Adds a new monitor to remove everything that isn't =mass / =pingmass / =groupkill


gc:
Not in favour of this. I would rather just lock the channel so only hosters can talk in it.

PR #669, opened 2020-09-22
[BSO] Removes bank title border and hamstare if the bankBg is not the default

kevinsmendes:
### Description:

-   Checks if the bankBg is the default before it tries to draw the title border and the hamstare
-   Removes the awkward spacing on the right of the bank image and makes the distance the same as the left side

### Changes:

-   Removes bank title border and hamstare if the bankBg is not the default
-   Changed the formula of the items spacing a little bit

### DIFF
--- a/src/tasks/bankImage.ts
+++ b/src/tasks/bankImage.ts
@@ -145,7 +145,7 @@ export default class BankImageTask extends Task {
 		this.itemIconImagesCache.set(itemID, image);
 	}
 
-	drawBorder(canvas: Canvas) {
+	drawBorder(canvas: Canvas, titleLine = true) {
 		const ctx = canvas.getContext('2d');
 		// Draw top border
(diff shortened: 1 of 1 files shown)


src/tasks/bankImage.ts
```diff
@@ -366,7 +370,7 @@ export default class BankImageTask extends Task {
 		for (let i = 0; i < items.length; i++) {
 			if (i % itemsPerRow === 0) yLoc += Math.floor((itemSize + spacer / 2) * 1.08);
 			xLoc = Math.floor(
-				spacer + (i % itemsPerRow) * ((canvas.width - 40) / itemsPerRow) + distanceFromSide
+				spacer + (i % itemsPerRow) * ((canvas.width - 24) / itemsPerRow) + distanceFromSide
```
gc, line 373: whats this change? also we need to remove the title border on osb

PR #687, opened 2020-09-25
Sarachnis

Lajnux:
### Description:

Added Sarachnis to killable monsters

### Changes:

Added Sarachnis to killable monsters


gc:
Doing this myself now because you're asleep

PR #693, opened 2020-09-25
[OSB] Add improved food handling for +mass and +k

kevinsmendes:
### Description:

-   Improve food removal when the food updates are re-added;
-   Adds a little handler for mass reactions, to avoid users having to react two times.
-- That still doesn't fix the reaction not accounting if the user reacts too fast while the mass reactions are still being generated.

### Changes:

- Add a couple new functions that can be used to handle food usage. 
-- `hasEnoughFoodForMonster` will return true or false depending if the user has enough food to fulfil the function params;
-- `getUserFoodFromBank` will return the best food combination to remove from the user bank, according to the total healing amount needed or false, if the user doesn't have enough food to filly heal that amount;
-- `removeFoodFromUser` will try to remove food form the user bank and will throw an error if it can't.
- Add reaction removal if leader tries to join or non-leader tries to start/stop mass [triggers reaction rate-limit but it is executed at next possible time]

### DIFF
--- a/src/commands/Minion/groupkill.ts
+++ b/src/commands/Minion/groupkill.ts
@@ -2,18 +2,15 @@ import { CommandStore, KlasaMessage, KlasaUser } from 'klasa';
 
 import { BotCommand } from '../../lib/BotCommand';
 import { Activity, Emoji, Tasks } from '../../lib/constants';
-import { Eatables } from '../../lib/eatables';
 import { ironsCantUse, minionNotBusy, requiresMinion } from '../../lib/minions/decorators';
 import { findMonster, reducedTimeForGroup } from '../../lib/minions/functions';
 import calculateMonsterFood from '../../lib/minions/functions/calculateMonsterFood';
 import hasEnoughFoodForMonster from '../../lib/minions/functions/hasEnoughFoodForMonster';
+import removeFoodFromUser from '../../lib/minions/functions/removeFoodFromUser';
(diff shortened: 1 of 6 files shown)


src/lib/minions/functions/removeFoodFromUser.ts
```diff
@@ -0,0 +1,56 @@
+import { KlasaClient, KlasaUser } from 'klasa';
+import { addBanks, removeBankFromBank } from 'oldschooljs/dist/util';
+import { O } from 'ts-toolbelt';
+
+import { Eatables } from '../../eatables';
+import { ClientSettings } from '../../settings/types/ClientSettings';
+import { UserSettings } from '../../settings/types/UserSettings';
+import { Bank, ItemBank } from '../../types';
+
+export function getUserFoodFromBank(
```
gc, line 10: youre exporting 2 functions from this file, when the file is named after only 1, just put the other in a new file

src/commands/Minion/groupkill.ts
```diff
@@ -92,15 +94,21 @@ export default class extends BotCommand {
 					return [true, `you don't have the requirements for this monster; ${reason}`];
 				}
 
-				if (1 > 2) {
+				if (monster.healAmountNeeded) {
```
gc, line 97: put the `1 > 2` back to keep it disabled in master for now.

gc (changes requested):

PR #694, opened 2020-09-25
[OSB] Add Zamorakian hasta as a possible createable

kevinsmendes:
### Description:

-   With the recent createables.ts skill change, it is now possible to implement this;
-   Add a new property called GPCost, that will charge the user that amount when trying to create the item.

### Changes:

-   Add GPCost to createables.ts and create.ts
-   Add Zamorakian hasta as a possible createable

### DIFF
--- a/src/commands/Minion/create.ts
+++ b/src/commands/Minion/create.ts
@@ -124,8 +124,7 @@ export default class extends BotCommand {
 		);
 
 		if (createableItem.GPCost) {
-			const userGP = msg.author.settings.get(UserSettings.GP);
-			await msg.author.settings.update(UserSettings.GP, userGP - createableItem.GPCost);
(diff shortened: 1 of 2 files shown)


src/commands/Minion/create.ts
```diff
@@ -107,7 +123,12 @@ export default class extends BotCommand {
 			addBanks([outItems, removeBankFromBank(userBank, inItems)])
 		);
 
-		if (!createableItem.noCl) msg.author.addItemsToCollectionLog(outItems);
+		if (createableItem.GPCost) {
+			const userGP = msg.author.settings.get(UserSettings.GP);
+			await msg.author.settings.update(UserSettings.GP, userGP - createableItem.GPCost);
```
gc, line 128: use `await msg.author.removeGP(createableItem.GPCost);` instead

PR #695, opened 2020-09-25
[OSB] Hide most drops if 10 or less players have not been dropped it yet

kevinsmendes:
### Description:

- Hide most drops if 10 or less players have not been dropped it yet

### Changes:

- Changed mostdrops to see if the result is less than 10 instead of 0.

### DIFF
--- a/src/commands/Patrons/mostdrops.ts
+++ b/src/commands/Patrons/mostdrops.ts
@@ -27,7 +27,7 @@ export default class extends BotCommand {
 			}[]
 		>(query);
 
-		if (result.length === 0) throw `No results found.`;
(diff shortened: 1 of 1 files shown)


src/commands/Patrons/mostdrops.ts
```diff
@@ -27,7 +27,7 @@ export default class extends BotCommand {
 			}[]
 		>(query);
 
-		if (result.length === 0) throw `No results found.`;
+		if (result.length < 10) throw `Not enough of this item have been dropped yet.`;
```
gc, line 30: Just anonymize it. If theres less than 10, show `Anonymous` as the username instead.

PR #782, opened 2020-10-02
Add Zammy Hasta Reversion to Creatables

osrsolympus:
Should allow for people to revert their hasta to a spear for free like in game. Dunno if it actually works since I’m on mobile and don’t have a test bot.

-   [] I have not tested all my changes thoroughly.


osrsolympus:
Ok regardless of that working or not due to mobile formatting issues the base should be good please add it people don’t realize they can’t Corp with the hasta.

PR #791, opened 2020-10-02
Allow Nightmare Solo Trips to be Continued and Modify handleTripFinish Function

ard35:
### Description:

-   Allows responding with the specified letter in order to continue a The Nightmare solo.
-   Allows handleTripFinish() function to pass in loot (for clue scroll on 'c' support) and images. This would let us use the function for monster killing activities, or for activities which drop clue scrolls (like skills in the upcoming future).  
-   This change will reduce code bloat for multiple activities that previously did not support handleTripFinish() such as monster killing, or solos such as Nightmare, Raids (BSO), etc.

### Changes:

-   Added optional args in handleTripFinish() for loot and images.
-   Added complete clue from loot with 'c' into handleTripFinish() using the new loot arg.
-   Allows handleTripFinish() to send a string attached to an image (keeps it 1 message sent for both) using the new image arg.
-   Added the handleTripFinish() to nightmareActivity.ts and monsterActivity.ts.


gc (changes requested):
You're sending 2 messages now here, 1 for the image and 1 for the text.. You need to pass the image as an arg to `handleTripFinish` - see `monsterActivity.ts` for an example.

PR #812, opened 2020-10-04
Fix some small glitches in Tokkulshop

### LINKED ISSUE #796: Minion TokkulShop insufficient items message
## 🐛 Bug Report

There is an error in the error message that is given if you attempt to sell more items than you have in the discord minion bot tokkul shop.
I believe "inItems" and "outItems" on lines 81 and 82 should be switched around, or they should be switched at line 92 in src/commands/Minion/tokkulshop.ts.
An example of the current error message:
tksError
This is a pretty confusing message, when you are trying to sell death runes for tokkul.

### LINKED ISSUE #811: Tokkulshop unbuyable items
## 🐛 Bug Report

When you try to buy certain ores from the new tokkulshop, you get the following error message
afbeelding

This is a confusing message as you are trying to buy an item that it says it can sell to you.
The issue is caused by the fact that coal+ ores do not have a tokkul cost price src/lib/buyables/tokkulBuyables.ts, which is probably to simulate a default stock of 0.
But when the list of buyable items is created by *mapping All items in tokkulBuyables*, the situation arises where you get an error if you try to buy an item that it says it can sell.

### LINKED ISSUE #816: Selling uncut onyx to tokkul shop
## 🐛 Bug Report

+tokkulshop sell uncut onyx gives error

kevinsmendes:
### Description:

-   The message saying you can only buy certain wares was reading from the incorrect variable
-   The error message when selling more than you have was inverted
-   Fix users trying to sell what they dont have (was giving `An unexpected error occurred :RSSad:`)
-   Resolves #811
-   Resolves #796 
-   Resolves #816 

### Changes:

-   Changes it so only the items that actually be sold are shown on that message
-   Changes the order of the variables in the sell message error
-   Adds a check if user is selling and don't own the any of what they are selling

### DIFF
--- a/src/commands/Minion/tokkulshop.ts
+++ b/src/commands/Minion/tokkulshop.ts
@@ -62,6 +62,12 @@ export default class extends BotCommand {
 			);
 		}
 
+		if (type === 'sell' && msg.author.numItemsInBankSync(shopInventory.inputItem) === 0) {
+			return msg.send(
+				`I am sorry JalYt. You don't have any **${shopInventory.name}** to sell me.`
+			);
+		}
+
 		if (quantity === undefined) {
 			quantity = type === 'sell' ? userBank[shopInventory.inputItem] : 1;
 		}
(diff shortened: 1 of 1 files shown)


imTurbo:
Awesome. Thanks for showing me this. New to github especially mobile haha.

PR #818, opened 2020-10-04
Add gem rocks mining

Alexsuperfly:
### Description:

rebased from Dev since that branch is depricated

### Changes:

- addition of gem rocks to mine
- adjustment of gold and iron mining exp to be the same as ingame
- crushing of semiprecious gems when trying to cut them


src/tasks/minions/miningActivity.ts
```diff
@@ -1,5 +1,6 @@
 import { roll } from 'e';
 import { Task } from 'klasa';
+import Loot from 'oldschooljs/dist/structures/Loot';
```
gc, line 3: this should be `Bank`, not `Loot`.

src/commands/Minion/mine.ts
```diff
@@ -120,6 +139,16 @@ export default class extends BotCommand {
 				}
 			}
 		}
+		// Check if the user got a amulet of glory
+		if (ore.name.toLowerCase() === 'gem rock') {
```
gc, line 143: use ID.

gc (changes requested):

PR #823, opened 2020-10-04
Improve and add more testing commands

kevinsmendes:
### Description/Changes:

-   Some testing commands were lacking and people would expend a lot of time figuring out what were missing and spawning things
-   `+hax` now adds every item in this page [https://oldschool.runescape.wiki/w/Armour/Highest_bonuses], set MAX_QP, Max cash stack, fightcave potions, all foods and auto-equip best mage, range, slash melee gear, graceful, and so on
-   `+clean` Add new command. Resets the user minion. `+clean --bank` only clears the user bank
-   `+setqp` Add new command. Allows the user to set the QP to it's desired amount
-   `+setgp` Removed 1B limit
-   `+spawn` Remove 1M limit, make spawned items automatically added to collectionLog, Allow items to be equipped on misc gear set.

### DIFF
--- a/src/commands/Testing/clean.ts
+++ b/src/commands/Testing/clean.ts
@@ -1,0 +1,43 @@
+import { CommandStore, KlasaMessage } from 'klasa';
+
+import { BotCommand } from '../../lib/BotCommand';
+import { UserSettings } from '../../lib/settings/types/UserSettings';
+
+export default class extends BotCommand {
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			cooldown: 1,
+			oneAtTime: true,
+			aliases: ['clear']
+		});
+		// this.enabled = !this.client.production;
+	}
+
+	async run(msg: KlasaMessage) {
+		// Make 100% sure this command can never be used in prod
+		if (
+			this.client.production ||
+			!this.client.user ||
+			this.client.user.id === '156230339621027840'
+		) {
+			return;
+		}
+
+		if (!msg.flagArgs.bank) {
+			await msg.author.settings.reset();
+			await msg.author.settings.update(UserSettings.Minion.HasBought, true);
+		} else {
+			await msg.author.settings.update(UserSettings.Bank, {});
+			await msg.author.settings.update(UserSettings.CollectionLogBank, {});
+			await msg.author.settings.update(UserSettings.GP, 0);
+		}
+
+		return msg.send(
+			`${
+				!msg.flagArgs.bank
+					? 'Your minion is reset! You can use --bank to clear only your bank!'
+					: 'You cleared your bank!'
+			}`
+		);
+	}
+}
--- a/src/commands/Testing/clean.ts
+++ b/src/commands/Testing/clean.ts
@@ -10,7 +10,7 @@ export default class extends BotCommand {
 			oneAtTime: true,
 			aliases: ['clear']
 		});
-		// this.enabled = !this.client.production;
+		this.enabled = !this.client.production;
 	}
 
 	async run(msg: KlasaMessage) {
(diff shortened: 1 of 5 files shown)


src/commands/Testing/hax.ts
```diff
@@ -23,19 +26,192 @@ export default class extends BotCommand {
...
+				neck: { item: itemID('Necklace of anguish'), quantity: 1 },
+				body: { item: itemID('Armadyl chestplate'), quantity: 1 },
+				legs: { item: itemID('Armadyl chainskirt'), quantity: 1 },
+				feet: { item: itemID('Pegasian boots'), quantity: 1 },
+				cape: { item: itemID("Ava's assembler"), quantity: 1 },
+				shield: null,
+				hands: { item: itemID('Barrows gloves'), quantity: 1 },
+				ring: { item: itemID('Archers ring (i)'), quantity: 1 },
+				weapon: null,
+				ammo: { item: itemID('Onyx bolts (e)'), quantity: 99 }
```
gc, line 47: These could be made much cleaner with the new `constructGearSetup` thing I added. Feel free to move this function into `lib/minions/gear` or somewhere if we're using it outside tests now.

src/commands/Testing/clean.ts
```diff
@@ -0,0 +1,43 @@
...
+import { UserSettings } from '../../lib/settings/types/UserSettings';
+
+export default class extends BotCommand {
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			cooldown: 1,
+			oneAtTime: true,
+			aliases: ['clear']
+		});
+		// this.enabled = !this.client.production;
```
gc, line 13: commented out?

src/commands/Testing/hax.ts
```diff
@@ -23,19 +26,192 @@ export default class extends BotCommand {
...
+				ring: { item: itemID('Warrior ring (i)'), quantity: 1 },
+				weapon: null,
+				ammo: { item: itemID('Unholy blessing'), quantity: 1 },
+				'2h': { item: itemID('Armadyl godsword'), quantity: 1 }
+			};
+			await msg.author.settings.update(UserSettings.Gear.Range, rangeGear);
+			await msg.author.settings.update(UserSettings.Gear.Mage, mageGear);
+			await msg.author.settings.update(UserSettings.Gear.Melee, meleeGear);
+			await msg.author.settings.update(UserSettings.Gear.Skilling, skillingGear);
+			await msg.author.settings.update(UserSettings.Gear.Misc, miscGear);
```
gc, line 109: update this all at once with the tuple syntax, for example:
```ts

			await msg.author.settings.update([
				[UserSettings.Minion.Ironman, true],
				[UserSettings.Minion.HasBought, true]
			]);
```

src/commands/Testing/hax.ts
```diff
@@ -23,19 +26,192 @@ export default class extends BotCommand {
...
+		} else {
+			await this.client.queuePromise(async () =>
+				msg.channel.send(
+					'As you are busy, BiS gear was not added to your gear presets. Run this again when you are not busy to have BiS gear equipped!',
+					{ split: false }
+				)
+			);
+		}
+		const loot = {
```
gc, line 118: use `resolveNameBank` instead of `itemID` on every key.

src/commands/Testing/setgp.ts
```diff
@@ -7,7 +7,7 @@ import { UserSettings } from '../../lib/settings/types/UserSettings';
 export default class extends BotCommand {
 	public constructor(store: CommandStore, file: string[], directory: string) {
 		super(store, file, directory, {
-			usage: '<amount:int{1,1000000000}>'
+			usage: '<amount:int{1}>'
```
gc, line 10: why did you remove the limit?

src/commands/Testing/setqp.ts
```diff
@@ -0,0 +1,31 @@
+import { CommandStore, KlasaMessage } from 'klasa';
+
+import { BotCommand } from '../../lib/BotCommand';
+import { MAX_QP } from '../../lib/constants';
+import { UserSettings } from '../../lib/settings/types/UserSettings';
+
+export default class extends BotCommand {
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			usage: '<qp:int{0,2147483647}>',
```
gc, line 10: you put the gp limit on qp?

gc (changes requested):

src/commands/Testing/setqp.ts
kevinsmendes: Resolved them all but the gear one. We don't use that anywhere currently but +hax, no need to change it's implementation until we really need it.

PR #825, opened 2020-10-04
Prevents WTF items from breaking the bank

kevinsmendes:
### Description:

-   WTF items were breaking the bot, this allows WTF items to still show on the bank images in case that happens in the future

### Changes:

-   Add some checks on  `bank.ts`, `bankImage.ts`, `Client.ts`, `createTupleOfItemsFromBank.ts` and `createReadableItemListFromTuple.ts`.
-   WTF items that doesn't have a icon will show the icon of the item `7542 - Cake of guidance`
-   Dwarven remains, shown on the picture below, requires the following OSJS PR to be accepted https://github.com/oldschoolgg/oldschooljs/pull/156, but it won't affect this update. All it'll do is show WTF-0 instead of Dwarven remains

image


LookImHigh:
seems fixed to me

PR #836, opened 2020-10-05
Add Sandworms to Buyables

osrsolympus:
Adds sandworms to buyables (at 200gp ea as a filler given I don't know how you want your scaling on prices v. in game osrs prices). Now that food has a purpose it seems like fishing anglers might be an activity players would like to engage in.

-   [x] I have not tested all my changes thoroughly.

### DIFF
--- a/src/lib/buyables/buyables.ts
+++ b/src/lib/buyables/buyables.ts
@@ -232,6 +232,15 @@ const Buyables: Buyable[] = [
 		gpCost: 1500
 	},
 	{
+		name: 'Sandworms',
+		aliases: ['sandworms'],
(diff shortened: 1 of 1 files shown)


gc:
This seems useless if you're not also adding anglers.

osrsolympus:
Anglers should already be fishable

gc:
You need kourend favour to fish anglers, which we plan to add, therefore we'll hold off on this until kourend favor is added - as we want them to be locked behind the requirement like ingame.

osrsolympus:
Ok I thought kourned things were just being quest point locked (like how Anglers are currently coded) my b.

Issue #855, opened 2020-10-08
Add 5-man for Nightmare


enriquevilla:
Since 5-man is optimal for drops at Nightmare, how about adding a command like this:

`+nightmare 5man`

with the objective of having a maximum of 5 people join. Another alternative would be to specify the mass size when creating (obviously capping at 11), so something like this for example:

`+nightmare mass 5`

Either of these two suggestions would save a lot of headaches for everyone trying to host specific group sized teams for Nightmare.

PR #860, opened 2020-10-09
Allow Specific Mass Sizes for Nightmare / Groupkill (GWD, Corp)

ard35:
### Description:

-   Allow users to specify the max mass size for a nightmare or groupkill (GWD, Corp) mass.
-   Resolves feature request #855.  

### Changes:

- Add functionality for a specific mass size from 2 to 10 (including leader) for nightmare.
- Add functionality for a specific mass size from 2 to 50 (including leader) for groupkill (GWD, Corp).
- Add a timeout feature to the function startTrip() so it allows time for the remove user debounce and processes users accordingly.

### DIFF
--- a/src/commands/Minion/groupkill.ts
+++ b/src/commands/Minion/groupkill.ts
@@ -16,7 +16,7 @@ import calcDurQty from '../../lib/util/calcMassDurationQuantity';
 export default class extends BotCommand {
 	public constructor(store: CommandStore, file: string[], directory: string) {
 		super(store, file, directory, {
-			usage: '[quantity:int] <monster:string>',
+			usage: '<monster:string> [quantity:int{2,50}]',
 			usageDelim: ' ',
 			cooldown: 5,
 			oneAtTime: true,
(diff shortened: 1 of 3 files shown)


src/commands/Minion/groupkill.ts
```diff
@@ -60,18 +60,21 @@ export default class extends BotCommand {
...
-	async run(msg: KlasaMessage, [inputQuantity, monsterName]: [number | undefined, string]) {
+	async run(msg: KlasaMessage, [monsterName, maximumSizeForParty]: [string, number]) {
 		const monster = findMonster(monsterName);
 		if (!monster) throw `That monster doesn't exist!`;
 		if (!monster.groupKillable) throw `This monster can't be killed in groups!`;
 
 		this.checkReqs([msg.author], monster, 2);
 
+		let maximumSize = 50;
+		if (maximumSizeForParty) maximumSize = maximumSizeForParty;
```
gc, line 71: this would be a lot simpler by just doing: `(maxSize ?? maxPartySize) - 1`  below I think.

gc (changes requested):

PR #883, opened 2020-10-12
Fix Patron Tier Needed Message (OSB & BSO)

ard35:
### Description:

-   Some patron only commands also allow for server boosters to use. This resulted in the message below which was unclear on what was needed due to the nature of how the tiers are coded:
image

- This will be fixed to say: 
image
 
### Changes:

-   If the needed `perkTier - 1 = 0`, use `perkTier or server booster` instead in the return message.  


gc:
This needs to target OSB/master.

gc:
Fixed in https://github.com/oldschoolgg/oldschoolbot/commit/fe21aa38d8873f59d9d286802ff68c45dc312310

Issue #829, opened 2020-10-05
[URGENT] [FIX THIS SOON] [FUCKY WUCKY ALERT]


ghost:
## 🚀 Feature Proposal

You make us use crush weapons for nightmare, but then reduce our food based off slash attack? lol
https://cdn.discordapp.com/attachments/625027094493134868/762493970012962826/unknown.png
https://cdn.discordapp.com/attachments/625027094493134868/762494201107185664/unknown.png

I'd also like a food refund for finding this
+ev this.client.users.get('314102784116916226').addItemsToBank({[391]: 10000}, false);

PR #888, opened 2020-10-13
Fix Food Usage Calculation for Bossing

ard35:
### Description:

-   Resolves issue #829, with some additional fixes.
-   The current bug uses the same attack style for both the monster and the player. This means, even though the monster is weak to crush and attacks with slash, the function would calculate your slash attack bonus for reducing food, even though it had forced you to use a crush weapon. This means that the current BIS is godsword over inquisitor's mace + elysian spirit shield at The Nightmare, for example.
-   Caused an over usage for food in most bossing which requires food.

### Changes:

-  Add separate attack styles used for Monster and Player. This lets the `calculateMonsterFood()` function use the monster's attack styles in your defensive gear calculator and the player's attack style in your offensive gear calculation.

### DIFF
--- a/src/commands/Minion/minion.ts
+++ b/src/commands/Minion/minion.ts
@@ -500,7 +500,12 @@ ${Emoji.QuestIcon} QP: ${msg.author.settings.get(UserSettings.QP)}
 		}
 
 		// Check food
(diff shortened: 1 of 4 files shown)


Lajnux:
Not gonna be further maintained.

PR #913, opened 2020-10-15
Remove intents from BSO

kevinsmendes:
### Description:

-   BSO is not being allowed into new servers
-   Not sure how I am able to test this out.

### Changes:

-   Disable the server intents from the bot


gc (changes requested):
Dont remove all intents, all you need to remove is the `'GUILD_MEMBERS',` one.

src/lib/config/Intents.ts
```diff
@@ -31,12 +30,6 @@ export class Intents extends BitField<IntentsString> {
 		 * - CHANNEL_PINS_UPDATE
 		 */
 		GUILDS: 1 << 0,
-		/**
-		 * - GUILD_MEMBER_ADD
-		 * - GUILD_MEMBER_UPDATE
-		 * - GUILD_MEMBER_REMOVE
-		 */
-		GUILD_MEMBERS: 1 << 1,
```
gc, line 39: you dont change this file. you only have to remove the 1 line from the other file

PR #933, opened 2020-10-21
zalcano

gc:
### Description:

-   _Write description here_

### Changes:

-   _List of changes here_


src/tasks/minions/minigames/zalcanoActivity.ts
```diff
@@ -0,0 +1,70 @@
...
+				true,
+				{ showNewCL: 1 },
+				user
+			);
+
+		handleTripFinish(
+			this.client,
+			user,
+			channelID,
+			`${user}, ${user.minionName} finished killing ${quantity}x Zalcano.`,
```
ard35, line 62: Have this show total KC when the trip is returned

ard35 (changes requested):
Make sure to also add a "minion is currently killing zalcano" status to `src/extendables/user/minion.ts` under `case Activity.Zalcano`. This would prevent +m breaking when on a Zalcano trip.

src/commands/Minion/zalcano.ts
```diff
@@ -0,0 +1,104 @@
...
+		const kc = msg.author.getMinigameScore(MinigameIDsEnum.Zalcano);
+		const kcLearned = Math.min(100, calcWhatPercent(kc, 100));
+
+		const boosts = [];
+		let baseTime = Time.Minute * 6;
+		baseTime = reduceNumByPercent(baseTime, kcLearned / 6);
+		boosts.push(`${(kcLearned / 6).toFixed(2)}% boost for experience`);
+
+		const skillPercentage =
+			msg.author.skillLevel(SkillsEnum.Mining) + msg.author.skillLevel(SkillsEnum.Smithing);
```
Lajnux, line 58: After looking some YouTube videos where they did comparison between a 70 mining account and 96 mining account. There is some slight difference, but it only scales with mining and not smithing. So removing the smithing part makes sense. Other than that the rest of the code looks good, maybe have some kind of small scaling with the pickaxes but only like 1,2,3,4,5%

Also either having graceful gears as a requirement or boost makes sense since everyone on the youtube videos uses is.

Issue #949, opened 2020-10-25
Say q to repeat this trip.


DrunkKaiser:
Please, for the love of all that is holy.
Remove "Say q to repeat this trip.".
Currently when you get this and enter "q" it will instead start questing for you.

PR #958, opened 2020-10-27
Remove 'q' From Continuation Characters

ard35:
### Description:

-   Solves issue #949.

### Changes:

-   Remove q from the list of possible continuation characters for non-patrons.


gc:
Rather than doing this, we should probably just set `noPrefixDM: true,` to false, so that you have to use prefixes in DMs, otherwise this will just happen with other letters and commands.

gc:
Fixed in https://github.com/oldschoolgg/oldschoolbot/pull/982

PR #969, opened 2020-10-30
remove q from continuationChars

### LINKED ISSUE #949: Say q to repeat this trip.

Lajnux:
### Description:

removed q from continuationChars so that quests isn't sent out instead of trip.

closes #949 

### Changes:

removed q from continuationChars so that quests isn't sent out instead of trip.


Lajnux:
Too trigger fingered, already PRed.

PR #993, opened 2020-11-07
Allow Ori to work for mass activities

kevinsmendes:
### Description:

- Ori is the only PvM pet so far and it is not working (giving extra loot) on massings

### Changes:

- Add the Ori effect on mass activities by adding an extra loot if the user has Ori equipped during the for teamsLoot loop.

### Other checks:


kevinsmendes:
Let me know if I should add Ori's effect to nightmare and raids too.

PR #1059, opened 2020-12-01
MagnaThieving

Lajnux:
### Description:

Changed so that pickpocket should be accurate to in game. We can change the xp rate by changing the constant = pickpocketSpeed. Also added damagaTaken as a output. That we can further work on

### Changes:

Added stunTime, stunDamage, slope, intercept, to make line equations for the pickpockable npcs.

### Other checks:


gc:
Commited myself due to PR issues, thank you.

PR #1154, opened 2021-01-01
Fix Payment Bug for Planting in Empty Patches

### LINKED ISSUE #1153: Patches dieing while protected
## 🐛 Bug Report

I paid to protect 2 dragonfruit trees and they both died. 
Screenshot_20210101-093900_Discord
Screenshot_20210101-093851_Discord

ard35:
### Description:

- Fixes #1153 
- While planting in empty patches, the payment property was not being properly updated. 

### Changes:

- Have planting in empty patches check the payment argument instead of the last planted crop (which would never exist in this case, causing a null return).

### Other checks:


src/tasks/minions/farmingActivity.ts
```diff
@@ -161,7 +161,7 @@ export default class extends Task {
 				plantTime: currentDate + duration,
 				lastQuantity: quantity,
 				lastUpgradeType: upgradeType,
-				lastPayment: patchType.lastPayment
+				lastPayment: payment ? payment : false
```
gc, line 164: make this: `payment ?? false`

PR #1409, opened 2021-02-21
Exclusive item boosts

notmeta:
### Description:

Adds a new type of item boosts: exclusive
- Picks the best item boost from a subset, i.e. pick pegasian boots over ranger boots, but not both.

### Changes:

- Pegasian boots now have +7% boost in the same places ranger boots have a boost
- Moved all monster item boosts to new system
  - No other boosts have been introduced, except for Pegasians
- Reformatted `+monster` slightly to show the boost exclusivity.
- Added `Minion#resolveAvailableItemBoosts` which returns all boosts the user currently has in respect to the given monster.
  - Respects item boost exclusivity

### Other checks:

  - Also tested with multiple 'boost sets', i.e. armadyl crossbow / dragon crossbow, where ACB is picked over DCB (due to having higher boost)

sdf
sdf

### DIFF
--- a/src/extendables/User/Minion.ts
+++ b/src/extendables/User/Minion.ts
@@ -1,5 +1,6 @@
 import { User } from 'discord.js';
 import { Extendable, ExtendableStore, KlasaClient, KlasaUser } from 'klasa';
+import { Bank } from 'oldschooljs';
 import Monster from 'oldschooljs/dist/structures/Monster';
 import SimpleTable from 'oldschooljs/dist/structures/SimpleTable';
 
@@ -7,7 +8,7 @@ import { Activity, Emoji, Events, MAX_QP, PerkTier, Time, ZALCANO_ID } from '../
 import ClueTiers from '../../lib/minions/data/clueTiers';
 import killableMonsters, { NightmareMonster } from '../../lib/minions/data/killableMonsters';
 import { Planks } from '../../lib/minions/data/planks';
-import { GroupMonsterActivityTaskOptions } from '../../lib/minions/types';
+import { GroupMonsterActivityTaskOptions, KillableMonster } from '../../lib/minions/types';
 import { UserSettings } from '../../lib/settings/types/UserSettings';
 import Skills from '../../lib/skilling/skills';
 import Agility from '../../lib/skilling/skills/agility';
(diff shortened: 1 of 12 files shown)


Alexsuperfly:
Can you absorb the current boosting system into this new implementation and just replace the current with it, no need to maintain two fields that essentially do the same thing.

notmeta:
> Can you absorb the current boosting system into this new implementation and just replace the current with it, no need to maintain two fields that essentially do the same thing.

Done.

src/lib/minions/functions/reducedTimeForGroup.ts
```diff
@@ -27,19 +27,20 @@ export default async function reducedTimeForGroup(
...
 		let userItemBoost = 0;
-		if (monster.itemInBankBoosts) {
-			for (const [itemID, boostAmount] of Object.entries(monster.itemInBankBoosts)) {
-				if (!users[i].hasItemEquippedOrInBank(parseInt(itemID))) continue;
-				userItemBoost += boostAmount;
-			}
+		for (const [itemID, boostAmount] of Object.entries(
+			user.resolveAvailableItemBoosts(monster)
+		)) {
+			if (!user.hasItemEquippedOrInBank(parseInt(itemID))) continue;
```
Alexsuperfly, line 37: this should have already been taken care of by `resolveAvailableItemBoosts`

src/extendables/User/Minion.ts
```diff
@@ -761,4 +762,29 @@ export default class extends Extendable {
...
+		if (monster.itemInBankBoosts) {
+			for (const boostSet of monster.itemInBankBoosts) {
+				let highestBoostAmount = null;
+				let highestBoostItem = null;
+
+				// find the highest boost that the player has
+				for (const [itemID, boostAmount] of Object.entries(boostSet)) {
+					const parsedId = parseInt(itemID);
+					if (!this.hasItemEquippedOrInBank(parsedId)) continue;
+					if (!highestBoostAmount || boostAmount > highestBoostAmount) {
```
Alexsuperfly, line 777: `highestBoostAmoun` and `highestBoostItem` can just be initialized with `0` which would allow you not to have to check if it exists here before comparing

Alexsuperfly (changes requested):

src/extendables/User/Minion.ts
notmeta: I initially did this, but figured negative boosts might a thing but that sounds silly thinking about it again.

notmeta:
This has been merged in #1784

PR #1455, opened 2021-03-02
 Combat

gc:
### Other checks:


src/commands/Minion/buy.ts
```diff
@@ -51,6 +52,13 @@ export default class extends BotCommand {
 			}
 		}
 
+		if (
+			buyable.skillsNeeded &&
+			!skillsMeetRequirements(msg.author.rawSkills, buyable.skillsNeeded)
```
Lajnux, line 57: make it return what skills needed

src/commands/Minion/k.ts
```diff
@@ -1,6 +1,57 @@
...
+	styles: AttackStyles[]
+): [number, string] {
+	const skillTotal = addArrayOfNumbers(styles.map(s => user.skillLevel(s)));
+
+	let newDuration = duration;
+	let str = '';
+	let percent = round(calcWhatPercent(skillTotal, styles.length * 99), 2);
+
+	if (percent < 50) {
+		percent = 50 - percent;
```
Lajnux, line 44: so -50% penalty if less than 50 but max +15% if more than 50?

Seems wrong since you said you wanted just -15% to 15%+

PR #1496, opened 2021-03-07
Quest buyables

gc:
### Other checks:

### DIFF
--- a/src/lib/data/buyables/buyables.ts
+++ b/src/lib/data/buyables/buyables.ts
@@ -612,6 +612,13 @@ const Buyables: Buyable[] = [
 		gpCost: 500_000,
 		qpRequired: 105
 	},
+	{
+		name: 'Fishbowl pet',
+		outputItems: {
+			6672: 1
+		},
+		gpCost: 500_000
+	},
 	...sepulchreBuyables,
 	...constructionBuyables,
 	...hunterBuyables,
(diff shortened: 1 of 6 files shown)


notmeta (approved):

src/lib/data/buyables/buyables.ts
```diff
@@ -333,6 +311,230 @@ const Buyables: Buyable[] = [
...
+		gpCost: 100_000,
+		qpRequired: 52
+	},
+	{
+		name: 'Amulet of accuracy',
+		outputItems: {
+			[itemID('Amulet of accuracy')]: 1
+		},
+		gpCost: 50_000,
+		qpRequired: 52
```
Alexsuperfly, line 344: this is like one of the first quests you ever do, why is qp so high

src/lib/data/buyables/canifisClothes.ts
```diff
@@ -0,0 +1,75 @@
+import { resolveNameBank } from '../../util';
+import { Buyable } from './buyables';
+
+export const canifisClothes: Buyable[] = [
```
Alexsuperfly, line 4: these should have a qp requirement since getting to canifis requires a quest, something like 20 or so since its not a hard one.

src/lib/data/buyables/frem.ts
```diff
@@ -0,0 +1,82 @@
+import { resolveNameBank } from '../../util';
+import { Buyable } from './buyables';
+
+export const fremennikClothes: Buyable[] = [
```
Alexsuperfly, line 4: these should have a qp requirement since you have to complete the fremennik trials before she will speak to you, Its a med level quest so something like 75 qp sounds appropriate to me.

PR #1534, opened 2021-03-14
Added some lunar spells + 2 standard teleports

TizHub:
Added Lumbridge + Falador teleport

Added qp requirement + gp cost for lunars/plank make

Added Spin flax spell, tan leather spell, string jewellery spell (Doesn't apply to silver amulet (u) jewellery atm), Plank make spell & recharge dragonstone spell.

Could be done much better but I just wanted to attempt to at least get them added even in a basic way - Doesn't check for Fremmy hard on Tan leather and recharge dragonstone

Also done yarn lint - Quite a bit of help from Fishy to fix things and general help with coding

### Other checks:


src/commands/Minion/cast.ts
```diff
@@ -89,6 +99,14 @@ export default class extends BotCommand {
 			])
 		);
 
+		if (spell.gpCost) {
+			const gpCost = spell.gpCost * quantity;
+			if (gpCost < userGP) {
+				return msg.send(`You need ${gpCost} GP to create ${quantity} planks.`);
```
gc, line 105: Probably better to write just "to cast this spell quantity times" because it assumes every spell that requires gp is one that makes planks, unless that is actually true

gc (changes requested):

src/commands/Minion/cast.ts
TizHub: gp cost only applies to plank make, no cost on tan leather

src/commands/Minion/cast.ts
```diff
@@ -105,9 +124,9 @@ export default class extends BotCommand {
 		return msg.send(
 			`${msg.author.minionName} is now casting ${quantity}x ${
 				spell.name
-			}, it'll take around ${formatDuration(
-				duration
-			)} to finish. Removed ${cost} from your bank. **${xpHr}**`
+			}, it'll take around ${formatDuration(duration)} to finish. Removed ${cost}${
+				spell.gpCost ? ` and ${gpCost} Coins` : ``
+			} from your bank. ** ${xpHr}**`
```
Lajnux, line 129: remove space after **
```suggestion
			} from your bank. **${xpHr}**`
```

src/lib/skilling/skills/magic/castables.ts
```diff
@@ -42,5 +62,261 @@ export const Castables: Castable[] = [
 		xp: 55.5,
 		level: 45,
 		ticks: 5
+	},
+	{
+		id: itemID('Bow string'),
```
Lajnux, line 67: this spell gives crafting xp in game which is missing.

src/lib/skilling/skills/magic/castables.ts
```diff
@@ -42,5 +62,261 @@ export const Castables: Castable[] = [
...
+		xp: 81,
+		level: 78,
+		ticks: 3.5,
+		qpRequired: 50
+	},
+	{
+		id: itemID('Gold amulet'),
+		name: 'String Gold',
+		input: new Bank()
+			.add('Astral rune', 2)
```
Lajnux, line 140: also gives crafting xp in game, same goes for all other String Jewellery below.
which currently is missing.

src/lib/skilling/skills/magic/castables.ts
```diff
@@ -42,5 +62,261 @@ export const Castables: Castable[] = [
...
+		ticks: 3,
+		qpRequired: 50,
+		gpCost: 1050
+	},
+	{
+		id: itemID('Amulet of glory (4)'),
+		name: 'Recharge Glory',
+		input: new Bank()
+			.add('Astral rune', 1)
+			.add('Water rune', 5)
```
Lajnux, line 285: this consumes 4 water runes in game not 5, 
and the xp is 97.5 per item and require 89 magic.

and you cast it for a whole inventory and not one item.

same goes for all other Recharge Dragonstone below.

Lajnux (changes requested):

Lajnux:
Closed, Tiz still learning. Made new PR

PR #1535, opened 2021-03-15
small changes

coolbop32:
### Description:

-added clues to gem rocks
-added the ability to craft drift net
-made it so if you have the hard Fally diary or 99 crafting you are able to use the bank chest in the crafting guild for a 3.25x speed boost like in game
-added lil' creator to the pets.ts file

### Other checks:

### DIFF
--- a/src/commands/Minion/craft.ts
+++ b/src/commands/Minion/craft.ts
@@ -100,7 +100,7 @@ export default class extends BotCommand {
 		let timeToCraftSingleItem = Craft.tickRate * Time.Second * 0.6 + Time.Second / 4;
 
 		const hasDiary = hasFallyHardDiary(msg.author);
-		if ((Craft.bankChest && hasDiary) || msg.author.skillLevel(SkillsEnum.Crafting) >= 99) {
+		if (Craft.bankChest && (hasDiary || msg.author.skillLevel(SkillsEnum.Crafting) >= 99)) {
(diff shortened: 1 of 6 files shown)


src/commands/Minion/craft.ts
```diff
@@ -72,7 +97,13 @@ export default class extends BotCommand {
 		const requiredItems: [string, number][] = Object.entries(Craft.inputItems);
 
 		// Get the base time to craft the item then add on quarter of a second per item to account for banking/etc.
-		const timeToCraftSingleItem = Craft.tickRate * Time.Second * 0.6 + Time.Second / 4;
+
+		let timeToCraftSingleItem = Craft.tickRate * Time.Second * 0.6 + Time.Second / 4;
+
+		const hasDiary = hasFallyHardDiary(msg.author);
+		if ((Craft.bankChest && hasDiary) || msg.author.skillLevel(SkillsEnum.Crafting) >= 99) {
```
Alexsuperfly, line 104: this will speed everything up if >=99

coolbop32: no because only the tannable items have a value for bankChest

Alexsuperfly: you wrote `||`

coolbop32: I see what happened when I added diary didn't move bracket over

Issue #1664, opened 2021-04-09
[BSO] Jar of spirits doesn't seem to drop from corp in BSO


mylife212:
## 🐛 Bug Report

I can't find the jar in https://github.com/oldschoolgg/oldschoolbot/blob/bso/src/lib/simulation/Corp.ts for BSO, I could be wrong. But me and jess have done well over 4000+ rolls of jar with nothing, so im more inclined to believe its not added. PLEASE add it

gc:
Fixed in next update.

PR #1665, opened 2021-04-09
Adds Jar of spirits to BSO Corp

### LINKED ISSUE #1663: BSO Corp missing jar
Jar was never added to the BSO Corp drop table

osrsolympus:
Should resolve #1663 and #1664

### Description:
 Took language from OSB code and added it to the bso drop table.


osrsolympus:
Magna fixed it

PR #1722, opened 2021-04-22
Added Ornament Kits

osrsolympus:
Added some missing/bso-only ornament kit actions to the osb side of things. Includes nearly all ornament kits which resolve into distinct items (eg. this includes the abyssal whip into the volcanic abyssal whip but not things like the odium ward for whom the name is exactly the same when it has a kit applied as the base). 

Few of these can be currently done using osb items but it feels prudent to have them given the fact that slayer and cox are to be added soon.


Lajnux:
This is added

PR #1749, opened 2021-04-26
Fix Onyx Amulet Enchant Level

### LINKED ISSUE #1748: Magic level requirement for enchanting Onyx amulet is incorrect.
https://github.com/oldschoolgg/oldschoolbot/blob/master/src/lib/skilling/skills/magic/enchantables.ts

For some reason enchanting an Onyx amulet requires only 82 magic while rest of Onyx jewellery is 87.

osrsolympus:
Change the required magic level to enchant an onyx amulet from 82 to 87

Resolves #1748


gc:
Thank you

PR #1840, opened 2021-05-10
Scale Golden Nuggets to amount of ores mined like in game.

themrrobert:
**### Description:**

Changed 'nuggets' boolean to a number, where that number represents the number of ore you need to mine of that type times 10.  I aimed for 10 nuggets/hr at level 99 without boosts. The current rate is 7 per hour regardless of level or boosts.

**### Changes:**

Changed 'nuggets' from boolean => number
The number represents number of ore to mine for a 50% chance at nugged.
Gave each ore that can drop nuggets a base value that represents 10 nuggets/hr @ level 99 without boosts.
Changed nugget calculation to use the the Math.floor(quantity / (ore.nuggets / 10)) value

This scales linearly with the number of ore mined, so lower levels will notice fewer nuggets per hour, while higher levels should see more nuggets per hour.

**### Tuning:**
I calculated each ore's 30 minute trip qty (without boosts) to be: 
Gold: 204
Mithril: 107
Adamant: 66
Runite: 26
We will call this variable 'ore.basevalue' in the following calculation

It's important for the basevalue to be the Floor()'d quantity of a 30 minute trip or else you will get 1 fewer nuggets/hr than intended.

To target x Nuggets/hr we simply need to set ore.nuggets to:
Math.floor(10 * ore.basevalue/ x);

(ore.nuggets and ore.basevalue are equal when the target nuggets per hour is 10)

**### Notes:**
1. I am not 100% sure of the basevalues, as I got the data from someone with a 36 minute trip length while boosted, so I had to normalize the value. It should only be off by 1 or 2 if at all.
2. I targeted BSO, but this can just as easily be applied to master/OSB.

### DIFF
--- a/src/lib/skilling/skills/mining.ts
+++ b/src/lib/skilling/skills/mining.ts
@@ -116,7 +116,7 @@ const ores: Ore[] = [
 		name: 'Mithril ore',
 		respawnTime: 10,
 		petChance: 150_000,
-		nuggets: 107, //If you want N/hr at 99, for example, do round(floor(30_min_qty_without_boosts / N),1) * 10
+		nuggets: 107, //If you want N/hr at 99, for example, do Math.floor(10 * ore.nuggets / N);
 		clueScrollChance: 148_320
 	},
 	{


Lajnux:
Golden nuggets should come from Motherload Mine instead, which is a minigame, activity we not yet added. 

Would be better balancing it through that activity and leave the current state as is until that's done.

PR #1854, opened 2021-05-13
[BSO] Fix Fish sack - now extends fishing trip by 9 minutes as the BSO.md guide says

themrrobert:
### Description:

Added a switch case for Fishing skill in Minion.ts that checks for equipped Fish Sack and extends the trip. 

### Other checks:


themrrobert:
Updated to 9 minutes as the guide says, oops.

PR #1889, opened 2021-05-19
[BSO] Fix Sand Gecko's so they don't give 4x loot.

themrrobert:
### Description:

The current revision of bso branch has Sand Gecko's 4x-ing loot instead of 2x with Sandy + Hunter master cape.

This isn't in the live game, just want to be sure it doesn't get [re]introduced.

### Changes:

Removed extraneous loot.multiply(2).

### Other checks:


themrrobert:
No longer an issue after secret repo merge

PR #1903, opened 2021-05-21
Fix 'defaultfarming tier' error when no compost specified.

themrrobert:
### Description:
Fixed '+defaultfarming tier' command when no compost type is specified.

### Changes:
Added an if block to catch when the 'newCompostType' is undefined.

### Other checks:


src/commands/Minion/defaultfarming.ts
```diff
@@ -51,6 +51,13 @@ export default class extends BotCommand {
 	async tier(msg: KlasaMessage, [newCompostTier]: [CompostTier]) {
 		await msg.author.settings.sync(true);
 
+		if (newCompostTier === undefined) {
+			return msg.send(
+				'You must specify a valid compost type. The available tiers to select are `compost`, `supercompost`, and `ultracompost`.' +
+					`For example, \`${msg.cmdPrefix}defaultfarming tier supercompost\`.`
```
gc, line 57: This is gonna have a weird lack of a space after the fullstop, e.g. `ultracompost.For example`

themrrobert:
Ugh, now a different test failed looks like an internal server error from what i can tell, but i'm not a github integrations expert.

PR #1909, opened 2021-05-23
[BSO] Fixed =kill to work with BSO monsters

themrrobert:
### Description:
Enabled BSO monsters to be simulated with =kill

### Changes:
Updated kill.worker.ts to include the BSO libraries required, as slightly modified the default function there to use the BSO KillableMonster type.

### Other checks:

-   [yes] I have tested all my changes as thoroughly as possible without the BSO assets.


themrrobert:
Unit tests still broken from missing config it seems =/

PR #1919, opened 2021-05-25
[BSO] Fix Goldemar Mass issues

themrrobert:
### Description:
2 small fixes:
1) Un-joining doesn't remove user from Mass
2) When Mass has max users, it doesn't auto start like it should.

### Changes:
1) Added assignment of the filtered array to this.users, and 2) now we await for the user array to be updated before checking if the party is full.

### Other checks:


src/lib/structures/Mass.ts
```diff
@@ -169,7 +171,7 @@ This party will automatically depart in 2 minutes, or if the leader clicks the s
 	async removeUser(user: KlasaUser) {
 		if (user === this.leader) return;
 		if (!this.users.includes(user)) return false;
-		removeFromArr(this.users, user);
+		this.users = removeFromArr(this.users, user);
```
gc, line 174: `removeFromArr` already does mutate, so this is doing `this.users = this.users` ?

themrrobert: The comments of the fn says it mutates, but it doesn't actually. I also looked up the function online as well and it says it creates a new array that is returned, and does not mutate. 

If it mutated, there wouldn't have been a bug for me to catch and fix :)

PR #1929, opened 2021-05-27
[BSO] Brought back the ability to feed an item to Hammy with the =hammy command!

themrrobert:
### Description:
At mylife212's request, brought back the =hammy command to feed an item to Hammy. 

Now with eight more cute and devious ways to break your stuff!

### Changes:
Added hammy.ts to commands/bso/. Will destroy 1 of the item as it used to.
Also responds to =feed and =feedhammy.

### Other checks:


themrrobert:
Closing because Hammy 2.0 was released.

PR #1930, opened 2021-05-27
[BSO] Bringing you... Hammy 2.0! - Slightly buffed Hammy

themrrobert:
### Description:
A slightly buffed Hammy.
This one has a confirm option + --cf/--yes flag.
It also has a 1 in 10 chance to save the item.
With a rare 1 in 100 chance to double the item.

And of course, more fun Hammy hijinx! Hammy may or may not have a slight gambling addiction...

### Changes:
Added a confirm waiter. Can be bypassed with --confirm/--cf/--yes
1 in 10 chance to not destroy the item.
1 in 100 chance to double the item.

### Other checks:


src/commands/bso/hammy.ts
```diff
@@ -0,0 +1,107 @@
...
+			return msg.send(`You don't have a ${firstItem.name}.`);
+		}
+		if (!bank.has(itemID('Hammy'))) {
+			return msg.send(`You don't have a Hammy, so how could you feed it?`);
+		}
+
+		if (roll(chanceToDouble)) {
+			let loot = new Bank();
+			loot.add(firstItem.name);
+			await msg.author.addItemsToBank(loot, true);
```
gc, line 98: I dont think this should add to cl.

src/commands/bso/hammy.ts
```diff
@@ -0,0 +1,107 @@
...
+		if (!bank.has(firstItem.id)) {
+			return msg.send(`You don't have a ${firstItem.name}.`);
+		}
+		if (!bank.has(itemID('Hammy'))) {
+			return msg.send(`You don't have a Hammy, so how could you feed it?`);
+		}
+
+		if (roll(chanceToDouble)) {
+			let loot = new Bank();
+			loot.add(firstItem.name);
```
gc, line 97: use `.id`

gc (changes requested):

PR #1934, opened 2021-05-27
(BSO) Fixed a =ping typo

6enny6:
ping cmd was saying to use ‘+role name’ to add/remove rather than ‘=roles name’

Changed on BSO as OSB has no +roles yet, not sure if this will need changing again in the future if OSB code needs updating to follow suit


6enny6:
thanks codecov 👍🏻

PR #1941, opened 2021-05-28
[BSO] Fixed Displayed XP Multiplier in Sepulchre return msg

themrrobert:
### Description:
Fixed the sepulchre trip return message so it displays the correct XP for BSO.

This can probably be used as-is in master branch also.

### Changes:
Updated sepulchreActivity to use the string returned from user.addXP instead of it's own custom code.

### Other checks:

### DIFF
--- a/src/tasks/minions/minigames/sepulchreActivity.ts
+++ b/src/tasks/minions/minigames/sepulchreActivity.ts
@@ -43,19 +43,13 @@ export default class extends Task {
 		}
 
 		await user.addItemsToBank(loot.bank, true);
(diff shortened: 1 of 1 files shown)


gc:
>This can probably be used as-is in master branch also.

If this problem is also present in OSB, it should target OSB and then it'll be merged down to BSO in my usual updates.

themrrobert:
> > This can probably be used as-is in master branch also.
> 
> If this problem is also present in OSB, it should target OSB and then it'll be merged down to BSO in my usual updates.

I thought about it, but it's not actually a problem in OSB, because there's no multiplier there.

That said, I will just make a new PR for master, less differences is good.

PR #1947, opened 2021-05-30
Add --items to +create to attach a download of creatables.

themrrobert:
### Description:
Added ability to pass --items flag to +create to get a list of all creatables and their requirements.

Also works as-is on BSO.

### Changes:
In addition to the code described above, also changed the usage itemName to be optional, and then added a check After the --items flag check for an undefined itemname.

### Other checks:


gc:
I'd prefer if this used a `table`, I believe the fletch cmd does this.

themrrobert:
> I'd prefer if this used a `table`, I believe the fletch cmd does this.

done

PR #1951, opened 2021-05-31
Fixed Displayed XP Multiplier in Sepulchre return msg

themrrobert:
### Description:
Fixed the sepulchre trip return message so it displays the correct XP for BSO.

### Changes:
Updated sepulchreActivity to use the string returned from user.addXP instead of it's own custom code.

### Other checks:


gc:
Thanks!

PR #1962, opened 2021-06-02
[BSO] Fixed minor errors in descriptions / messages. Like =ods description, nex+kk kc boosts.

themrrobert:
### Description:
Fixed some description / message inaccuracies.

### Changes:
Changed boost messages in nex+kc so they say 50, 100, 200 instead of 200/200/200
Changed description of ODS from Barb assault to Ourania Delivery service.


gc:
Thanks!

PR #1980, opened 2021-06-03
[BSO] Fixed mysterious trees' planting boxes

themrrobert:
### Description:
Without this fix, if you plant/harvest with Mysterious trees already planted, you don't get the Mystery boxes you should for planting.

This fixes that issue.

### Changes:
Moved the 'planting' mystery box code above the if/else block so that it doesn't get skipped when plant+harvest happen together.

### Other checks:

### DIFF
--- a/src/tasks/minions/farmingActivity.ts
+++ b/src/tasks/minions/farmingActivity.ts
@@ -125,6 +125,15 @@ export default class extends Task {
 		}
 		if (farmersPiecesCheck === 4) bonusXpMultiplier += 0.005;
 
+		if (planting && plant !== undefined) {
+			if (plant.name === `Mysterious tree`) {
+				let upper = randInt(quantity, quantity * 4);
+				for (let i = 0; i < upper; i++) {
+					loot = addItemToBank(loot, getRandomMysteryBox());
+				}
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


gc:
There's 2 places in the file where it rolls for mysterious tree loot, do we need both? You copied the one with a max of 4 but I buffed the other to 6, too, so somethings gonna have to change there.

themrrobert:
Yes, we need both because it's always given you boxes for planting, and also for harvesting.

Currently, you have to harvest + plant separately to get the full benefit, otherwise you lose the "planting boxes" when you =farm harvest+plant together.

themrrobert:
I guess we could just leave the old code, and then just add in the additional boxes from planting into the roll.

This PR seemed like the best solution at the time, however it's probably cleaner to do as described here.

themrrobert:
Actually I just realized trees need another fix.
Example: Lets say you plant 5 trees, max would be 5x6 = 30.
That means you have the same chance to get box from 5 seeds as getting 30. This isn't right.

It should roll 1-6  5x. This means you're much more likely to get a mid-range number of boxes.
When you roll 5x 1-6 the probability graph looks like a bellshaped curved, with the greatest probability around 13-17 boxes, with very low chance of 1 or 30 boxes.

PR #1984, opened 2021-06-03
[BSO] Added --sv and --names to =open

themrrobert:
### Description:
With this patch you can use --sv and --names with =open on caskets, mystery boxes, and osjs openables :)

### Changes:

- Passthru msg.flagArgs to sendBankImage() function

### Other checks:


gc:
This should've targeted master, because it's a change not specific to BSO that would be useful to OSB people too, but I'll merge as is anyway.

PR #1990, opened 2021-06-04
[BSO] Spawnbox Return Message

6enny6:
Changed the return msg when using =spawnbox outside of #bot-channel-bso, as ‘The BSO Channel’ is now a little confusing with #bso-general

also the linter told me to add that return idk why

### DIFF
--- a/src/commands/bso/spawnbox.ts
+++ b/src/commands/bso/spawnbox.ts
@@ -30,7 +30,7 @@ export default class extends BotCommand {
 		}
 
 		if (msg.author.id !== '157797566833098752' && msg.channel.id !== '732207379818479756') {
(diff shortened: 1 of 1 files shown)


src/commands/bso/spawnbox.ts
```diff
@@ -30,7 +30,8 @@ export default class extends BotCommand {
 		}
 
 		if (msg.author.id !== '157797566833098752' && msg.channel.id !== '732207379818479756') {
-			return msg.send(`You can only use this in the BSO channel.`);
+			return msg.send(`
+You can only spawn boxes in #bot-channel-bso in the oldschool.gg server.`);
```
gc, line 34: This is going to cause an extra weird new-line at the start of the message.

gc (changes requested):

PR #2019, opened 2021-06-08
boosts for lower tier axes & pickaxes

6enny6:
Added smaller boosts for each of the lower tier axes for woodcutting and pickaxes for mining

will update on wiki if merged


Alexsuperfly:
original rates were made with rune gear in mind so i dont think this should be done

Lajnux:
This PR gonna have merge conflicts with #1871 and #1868 which def should have a priority to make the bot more alike osrs, other than that I don't see a problem with it since it won't have any major impact in the bot.

PR #2020, opened 2021-06-08
[BSO] Master Farmer Outfit 

6enny6:
Changed requirement to craft master farmer outfit from 110 to 105, to match other custom content


gc:
I don't really think this "has" to match the other content, although its a reasonable change but I thinks its random/unneeded.

PR #2036, opened 2021-06-10
Twisted/Occult Item Aliases

6enny6:
Added Twisted Ancestral and Occult necklace (or) aliases, so you can use monster boosts with those instead


themrrobert:
These already work, but only when they're in your bank. I fixed the bug that prevents the match on equipped items in my slayer PR, so the update is on the way already.

6enny6:
fixed in cyr’s slayer pr

PR #2053, opened 2021-06-13
Fix kodai/sotd cast bug. add cl entries, fix staff of water, ehp autoslay, cox rebalance

themrrobert:
### Description:
Fixed Kodai / Staff of the Dead bug causing non-rune costs to be altered.
Fixed bug that I introduced during code review breaking Staff of Water similaritems.
Added EHP autoslay
Rebalanced cox to only give 10% more boost instead 16% (10% shorter times, 26=>24 vs 26=>22 minutes)
Added some cl entries

### Other checks:

### DIFF
--- a/src/commands/Minion/k.ts
+++ b/src/commands/Minion/k.ts
@@ -44,6 +44,7 @@ import findMonster, {
 } from '../../lib/util';
 import addSubTaskToActivityTask from '../../lib/util/addSubTaskToActivityTask';
 import itemID from '../../lib/util/itemID';
+import {getSimilarItems} from "../../lib/data/similarItems";
 
 const validMonsters = killableMonsters.map(mon => mon.name).join(`\n`);
 const invalidMonsterMsg = (prefix: string) =>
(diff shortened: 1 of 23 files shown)


src/commands/Minion/k.ts
```diff
@@ -111,7 +112,8 @@ export default class extends BotCommand {
 		}
 		if (!name) return msg.channel.send(invalidMonsterMsg(msg.cmdPrefix));
 		const monster = findMonster(name);
-		if (!monster) return msg.channel.send(invalidMonsterMsg(msg.cmdPrefix));
+		if (!monster || !monster.canBeKilled)
+			return msg.channel.send(invalidMonsterMsg(msg.cmdPrefix));
```
gc, line 116: if an `if` statement is more than 1 line, wrap the body in parans `{}`

gc (changes requested):

themrrobert:
have to close since i need to force push, and it will break some of the tests.

PR #2065, opened 2021-06-14
Fix bugs, cl, and minor balancing

themrrobert:
### Description:
Added autoslay EHP/efficiency mode
Fixed sotd / kodai staff rune usage bug
Fix another bug related to kodai/sotd and other consumables.
Rebalanced Cox after talking to the dev who wrote it so dhl won't be too op in OSB.
Adjusted difficulty level on some mobs for autoslay.
Fixed some requirements on monsters that were incorrect.
Added wipeDBArrayByKey function that will clear all set values on an array DB key.
  ex: if you have a bitfield key with [1, 2, 4, 6], run the wipeDBArrayByKey command on the key to unset all values.

### Other checks:

### DIFF
--- a/src/commands/Minion/k.ts
+++ b/src/commands/Minion/k.ts
@@ -44,6 +44,7 @@ import findMonster, {
 } from '../../lib/util';
 import addSubTaskToActivityTask from '../../lib/util/addSubTaskToActivityTask';
 import itemID from '../../lib/util/itemID';
+import {getSimilarItems} from "../../lib/data/similarItems";
 
 const validMonsters = killableMonsters.map(mon => mon.name).join(`\n`);
 const invalidMonsterMsg = (prefix: string) =>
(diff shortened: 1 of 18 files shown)


src/commands/Minion/k.ts
```diff
@@ -111,7 +112,8 @@ export default class extends BotCommand {
 		}
 		if (!name) return msg.channel.send(invalidMonsterMsg(msg.cmdPrefix));
 		const monster = findMonster(name);
-		if (!monster) return msg.channel.send(invalidMonsterMsg(msg.cmdPrefix));
+		if (!monster || !monster.canBeKilled)
+			return msg.channel.send(invalidMonsterMsg(msg.cmdPrefix));
```
gc, line 116: ^ use parens. ill try make eslint force this later.

gc (changes requested):

PR #2082, opened 2021-06-15
Slayer BSO fixes and qol

themrrobert:
### Description:
Small fixes. All of the big stuff is in master.

### Changes:
Added malygos + sea kraken to their respective slayer tasks
Added the Ancient pool benefit so players don't get 0 boost for having the big pool.
Restored dragon hunter lance's 20% boost (for bso only. osb will only get 15%)

### Other checks:


themrrobert:
LOL magna we both did the same thing XD. alright i still have a few changes in this PR, but most are in master.

PR #2092, opened 2021-06-16
Fix OSB Daily reminder message.

themrrobert:
### Description:
Fix OSB Daily Reminder message

### Changes:
Reworked the code so it doesn't rely on user's being in the cache, and fixed the task so it runs on the new system.

### Other checks:

### DIFF
--- a/src/tasks/dailyReminder.ts
+++ b/src/tasks/dailyReminder.ts
@@ -1,15 +1,31 @@
 import { Task } from 'klasa';
 
+import { client } from '../index';
 import { PerkTier, Time } from '../lib/constants';
 import { UserSettings } from '../lib/settings/types/UserSettings';
 import { noOp } from '../lib/util';
 import getUsersPerkTier from '../lib/util/getUsersPerkTier';
 
 export default class extends Task {
+	async init() {
+		if (this.client.dailyReminderTicker) {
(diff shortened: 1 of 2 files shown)


src/tasks/dailyReminder.ts
```diff
@@ -1,24 +1,39 @@
...
-			if (getUsersPerkTier(user) < PerkTier.Two) continue;
-			const lastVoteDate = user.settings.get(UserSettings.LastDailyTimestamp);
-			if (lastVoteDate === -1) continue;
+		const dailyReady = currentDate - DailyInterval;
+		const result = await client.query<{ id: string }[]>(
+			`SELECT id FROM users WHERE bitfield && '{2,3,4,5,6}'::int[] AND "lastDailyTimestamp" != -1 AND "lastDailyTimestamp" < ${dailyReady};`
+		);
+
+		if (result.length === 0) return null;
+		for (const userID of result.values()) {
```
gc, line 32: `userID.id` doesnt make sense.

gc: you dont need to check if the length is 0, if theres no results, it will just have nothing to iterate over and theres nothing wrong with that.

gc (changes requested):

PR #2096, opened 2021-06-17
Fix Mysterious tree bug

themrrobert:
### Description:
Mysterious trees using the wrong variable to count # of mystery boxes. Should be using 'alivePlants'


themrrobert:
wait thsi is not right

PR #2127, opened 2021-06-19
[BSO] Remove twisted kit+metamorphic dust from boss log

mylife212:
### Description:.
This is for BSO only. Removed Metamorphic dust+twisted kit from log as they were only obtainable from early mboxes, and cannot be obtained anymore. Giving an unfair advantage to those who were lucky enough to pull it from boxes on the boss log leaderboard. Also doesn't make sense for a log to have unobtainable items 

### Changes:

- Deleted Metamorphic dust + Twisted ancestral kit from the "cox2" line in Boss Collection log

### Other checks:


mylife212:
can get dust+kits now

PR #2128, opened 2021-06-19
Add RC combination runes

Lajnux:
### Description:

added the possibility to runecraft combination runes, buy elemental runes (earth, fire, water, air, body, mind), added HUGE gold sink both for solos and runners. Accurate xp/hour to in game and slightly higher gp/xp than in game since OSB is inflated.

Could add a runecrafting sink tracker aswell if that is of interest.

### Changes:

Added combination runes to runecrafting and elemental runes to shop

### Other checks:


src/commands/Minion/rc.ts
```diff
@@ -73,6 +76,11 @@ export default class extends BotCommand {
 			boosts.push('5% for 60+ Agility');
 		}
 
+		if (msg.flagArgs.runners && !msg.author.isIronman) {
```
gc, line 79: as discussed in discord, we don't want  "runners".

gc (changes requested):

PR #2129, opened 2021-06-19
Correct offensive stat calculation for food

### LINKED ISSUE #2141: Food reduction from offensive stats
## 🐛 Bug Report

Looking at the food usage calculation function ("src/lib/minions/functions/calculateMonsterFood.ts", line 36), it seems as though the reduction of food based on the user's offensive stats are based on the attack styles of the monster rather than the user. This leads to strange and unrealistic results in the offensive stat food reduction.

For example, one could be killing mutated bloodvelds, and specifically +training a melee combat skill (irrelevant in the code). Bloodvelds look at your range gear set for defensive food reductions, and they attack with magical melee (using magic attack stat). The code calculates what appears to be the *user's* food reduction due to their offensive stats based on their magic attack stat in their range gear set, specifically because bloodvelds use magic attack.

This is problematic because the typical range gear setup heavily decreases magic attack, even into the negatives. The result of this is a 0% food reduction due to the user's offensive stats. This is just strange and unrealistic, because the user is training melee and theoretically using their range gear setup in this instance, neither of which have anything to do with magic attack.

Sorry if this is intended, but it seems unintended to me.

Lilylicious:
### Description:

Corrected the offensive stat calculation for food reduction.

### Changes:

Offensive stat percentage is now calculated once for the attack style the player is using. It is no longer divided by the amount of attack styles the monster is using. It is now capped at 95 like the defensive stat.

### Other checks:

Thoroughly tested, works fine. Only quirk is that melee always checks crush.

Closes #2141


Lilylicious:
This doesn't currently work because monsters are defined by gearsetuptypes, while we need GearStat types. We either need to define a specific offensive stat to use for each monster, or select a default stat for melee.

Lilylicious:
Fix tested this time around, and it works. All cases where melee stats are checked it uses crush attack, but that's just a matter of going through each monster and correcting what style is used.

PR #2138, opened 2021-06-20
EHP level requirements

Lilylicious:
### Description:

+as --ehp no longer selects monsters you lack requirements for.

### Changes:

Added a check to --ehp for level requirements. Probably only affects spiritual creatures right now, but still.

### Other checks:

-   [ x ] I have tested all my changes thoroughly.

### DIFF
--- a/src/commands/Minion/autoslay.ts
+++ b/src/commands/Minion/autoslay.ts
@@ -313,8 +313,9 @@ export default class extends BotCommand {
 			if (
 				ehpKillable?.levelRequirements !== undefined &&
 				!msg.author.hasSkillReqs(ehpKillable.levelRequirements)[0]
-			)
+			) {
 				return this.client.commands.get('k')?.run(msg, [null, usersTask.assignedTask!.monster.name]);
+			}
 
 			if (ehpMonster && ehpMonster.efficientName) {
 				if (ehpMonster.efficientMethod) msg.flagArgs[ehpMonster.efficientMethod] = 'force';
--- a/src/commands/Minion/autoslay.ts
+++ b/src/commands/Minion/autoslay.ts
@@ -318,9 +318,8 @@ export default class extends BotCommand {
 				}
 			});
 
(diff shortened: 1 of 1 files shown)


src/commands/Minion/autoslay.ts
```diff
@@ -307,6 +307,20 @@ export default class extends BotCommand {
 				return masterMatch && e.monsterID === usersTask.assignedTask!.monster.id;
 			});
 
+			const allMonsters = killableMonsters.filter(m => {
+				return usersTask.assignedTask!.monsters.includes(m.id);
```
themrrobert, line 311: You're checking if ANY of the possible task monsters have requirements you don't meet, instead you should just check if the efficient monster has requirements you can't meet. `ehpMonster.EfficientMonster`

src/commands/Minion/autoslay.ts
```diff
@@ -307,6 +307,20 @@ export default class extends BotCommand {
...
+			});
+
+			const cantEHP = allMonsters.some(m => {
```
themrrobert, line 314: and then here you only need to check the one monster, not the array.

themrrobert (changes requested):

src/commands/Minion/autoslay.ts
```diff
@@ -307,6 +307,15 @@ export default class extends BotCommand {
...
 			});
 
+			const ehpKillable = killableMonsters.find(m => m.id === ehpMonster?.efficientMonster);
+
+			// If we don't have the requirements for the efficient monster, revert to default monster
+			if (
+				ehpKillable?.levelRequirements !== undefined &&
+				!msg.author.hasSkillReqs(ehpKillable.levelRequirements)[0]
+			)
+				return this.client.commands.get('k')?.run(msg, [null, usersTask.assignedTask!.monster.name]);
```
gc, line 317: if an `if` statement is longer than 1 line, it needs to always have parens `{}`, its not enforced by eslint yet.

PR #2146, opened 2021-06-21
Correct ice burst costs

Lilylicious:
I guess no one tested bursts over barrage?


themrrobert (approved):
This is correct, i agree with these changes

PR #2148, opened 2021-06-21
Don't stack black mask and salve amulet

themrrobert:
### Description:
Black mask and salve i don't stack in the game, they shouldn't stack in the bot, it would be too op.

### Changes:
Black mask checks only roll if salve boost is not already applied.

### Other checks:

### DIFF
--- a/src/commands/Minion/k.ts
+++ b/src/commands/Minion/k.ts
@@ -182,7 +182,11 @@ export default class extends BotCommand {
 					timeToFinish = reduceNumByPercent(timeToFinish, 15);
 					boosts.push('15% for Black mask (i) on non-melee task');
 				}
-			} else if (isOnTask && msg.author.hasItemEquippedOrInBank('Black mask')) {
+			} else if (
+				isOnTask &&
+				(msg.author.hasItemEquippedOrInBank('Black mask') ||
+					msg.author.hasItemEquippedOrInBank('Black mask (i)'))
+			) {
 				timeToFinish = reduceNumByPercent(timeToFinish, 15);
 				boosts.push('15% for Black mask on melee task');
 			}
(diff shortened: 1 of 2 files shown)


src/commands/Minion/k.ts
```diff
@@ -172,18 +172,24 @@ export default class extends BotCommand {
...
+			// Add 15% slayer boost on task if they have black mask or similar
+			if (attackStyles.includes(SkillsEnum.Ranged) || attackStyles.includes(SkillsEnum.Magic)) {
+				if (isOnTask && msg.author.hasItemEquippedOrInBank('Black mask (i)')) {
+					timeToFinish = reduceNumByPercent(timeToFinish, 15);
+					boosts.push('15% for Black mask (i) on non-melee task');
+				}
+			} else if (
+				isOnTask &&
+				(msg.author.hasItemEquippedOrInBank('Black mask') ||
+					msg.author.hasItemEquippedOrInBank('Black mask (i)'))
```
gc, line 188: this `||` isnt needed.

src/commands/Minion/k.ts
```diff
@@ -172,18 +172,24 @@ export default class extends BotCommand {
 				boosts.push('15% for Dragon hunter crossbow');
 			}
 		}
-		// Add 15% slayer boost on task if they have black mask or similar
-		if (attackStyles.includes(SkillsEnum.Ranged) || attackStyles.includes(SkillsEnum.Magic)) {
-			if (isOnTask && msg.author.hasItemEquippedOrInBank('Black mask (i)')) {
+
+		// Black mask and salve don't stack.
+		const salveBoost = boosts.join('').toLowerCase().includes('salve amulet');
```
gc, line 177: this is some spaghetti. but im okay with it

gc (changes requested):

src/commands/Minion/k.ts
themrrobert: We discussed this in chat.

PR #2150, opened 2021-06-21
Craft bank usage and fletching filter

Lilylicious:
### Description:

+craft properly uses the Bank class to handle items.
+craft tells you what items it uses when leaving for a trip
Craftables filter now uses Bank class for its items
Fletching filter now works

### Changes:

Changed both +craft and Craftables to work very similar to +fletch and Fletchables. The main difference is in GP handling, due to tanning costs and the like.

### Other checks:

### DIFF
--- a/src/commands/Minion/craft.ts
+++ b/src/commands/Minion/craft.ts
@@ -3,12 +3,13 @@ import { CommandStore, KlasaMessage } from 'klasa';
 import { Activity, Time } from '../../lib/constants';
 import { FaladorDiary, userhasDiaryTier } from '../../lib/diaries';
 import { minionNotBusy, requiresMinion } from '../../lib/minions/decorators';
+import { ClientSettings } from '../../lib/settings/types/ClientSettings';
 import { UserSettings } from '../../lib/settings/types/UserSettings';
 import Crafting from '../../lib/skilling/skills/crafting';
 import { SkillsEnum } from '../../lib/skilling/types';
 import { BotCommand } from '../../lib/structures/BotCommand';
 import { CraftingActivityTaskOptions } from '../../lib/types/minions';
-import { bankHasItem, formatDuration, itemNameFromID, removeItemFromBank, stringMatches } from '../../lib/util';
+import { addBanks, formatDuration, itemNameFromID, stringMatches } from '../../lib/util';
 import addSubTaskToActivityTask from '../../lib/util/addSubTaskToActivityTask';
 
 export default class extends BotCommand {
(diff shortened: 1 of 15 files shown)


src/lib/settings/schemas/ClientSchema.ts
```diff
@@ -11,6 +11,7 @@ Client.defaultClientSchema
 	.add('farming_loot_bank', 'any', { default: {} })
 	.add('buy_cost_bank', 'any', { default: {} })
 	.add('magic_cost_bank', 'any', { default: {} })
+	.add('crafting_cost_bank', 'any', { default: {} })
```
gc, line 14: call it just `crafting_cost` (without the bank part)

src/commands/Minion/craft.ts
```diff
@@ -78,75 +76,77 @@ export default class extends BotCommand {
...
-		await msg.author.settings.update(UserSettings.Bank, newBank);
+
+		await msg.author.removeItemsFromBank(itemsNeeded);
+
+		if (gpNeeded > 0) await msg.author.removeGP(gpNeeded);
+
+		await this.client.settings.update(
+			ClientSettings.EconomyStats.CraftingCostBank,
+			addBanks([this.client.settings.get(ClientSettings.EconomyStats.CraftingCostBank), itemsNeeded.bank])
+		);
```
gc, line 127: This should use `updateBankSetting`, you were probably looking at a file which is still doing it in this way instead of with the function i made for this.

src/commands/Minion/craft.ts
```diff
@@ -78,75 +76,77 @@ export default class extends BotCommand {
...
+				`You don't have enough GP. For ${quantity}x ${craftable.name}, you're missing **${
+					gpNeeded - currentGP
+				}**.`
+			);
 		}
-		await msg.author.settings.update(UserSettings.Bank, newBank);
+
+		await msg.author.removeItemsFromBank(itemsNeeded);
+
+		if (gpNeeded > 0) await msg.author.removeGP(gpNeeded);
```
gc, line 122: This shouldn't be needed at all.  When you `removeItemsFromBank`, it also removes GP https://github.com/oldschoolgg/oldschoolbot/blob/master/src/extendables/User/Bank.ts#L180

gc (changes requested):

src/commands/Minion/craft.ts
Lilylicious: Normally yes, but on line 99 I'm removing GP from the `itemsNeeded`. It didn't seem to function with `userBank.has(itemsNeeded.bank)`

src/commands/Minion/craft.ts
```diff
@@ -78,75 +75,59 @@ export default class extends BotCommand {
...
 			quantity,
 			duration,
 			type: Activity.Crafting
 		});
 
+		let gpUsedString = '';
+
+		if (gpNeeded > 0) {
+			gpUsedString += ` and ${gpNeeded} GP`;
+		}
```
gc, line 123: This part is not needed and causes a weird output

image

gc (changes requested):

PR #2152, opened 2021-06-21
Fix aliases for clue filters

Lilylicious:
### Description:

While --filter="a filter" does work, no one was using it because it wasn't intuitive or something. This PR just removes spaces in the aliases to allow for using the filters with just --.

### Other checks:


gc (changes requested):
We discussed in discord, this isn't a good way to fix the issue.

Lilylicious:
Since --filter works, I'm closing this for the time being.

PR #2165, opened 2021-06-22
Adds --as to +slayertask and shows alternate monsters on receiving a task

kevinsmendes:
### Description:

- I noticed that whenever we got a new task, the alternated monsters available to kill didn't appeared on the return message. When issuing +st again, it would appear.
- Also, I found that a good QoL would be the addition of the parameter --as to the +st command. That way, we could just say "Minion, go do a new slayer task, I don't care what monster I get".

### Changes:

- Add getAlternateMonsterList and returnSuccess methods to slayertask.ts;
- Add flag --as/--autoslay to slayertask.ts, to allow autoslayer from a single command (+st --as is the same as +st then +as);
- Change de result message of a new task to include the alternate monster;
- When issuing --as/--autoslay to the slayertask.ts, the default output will be ignored for sucessfull comands (when a task is received/already in execution).

### Other checks:

image
image
image
image
image

Alternate Monsters being show on a new task being assigned
image


src/commands/Minion/slayertask.ts
```diff
@@ -32,13 +34,42 @@ export default class extends BotCommand {
...
+					return m!.name;
+				});
+			return alternateMonsters.length > 0 ? ` (${alternateMonsters.join('/')})` : '';
+		}
+		return '';
+	}
+
+	public async returnSuccess(msg: KlasaMessage, message: string, autoslay: boolean) {
+		if (autoslay) {
+			// await msg.channel.send(message);
```
themrrobert, line 54: Remove `// await` so the task assignment message comes through without problems. Magna might change this later, but I think people will want to see the message.

src/commands/Minion/slayertask.ts
```diff
@@ -32,13 +34,42 @@ export default class extends BotCommand {
 		});
 	}
 
+	public getAlternateMonsterList(currentTask: SlayerTaskTable, assignedTask: AssignableSlayerTask | null) {
```
themrrobert, line 37: Don't pass currentTask to this function. It adds an unneeded import dependency, and the purpose of the function can be entirely completed with just 'assignedTask.'

src/commands/Minion/slayertask.ts
```diff
@@ -32,13 +34,42 @@ export default class extends BotCommand {
...
 	@requiresMinion
 	async run(msg: KlasaMessage, [input]: [string | undefined]) {
 		const { currentTask, totalTasksDone, assignedTask } = await getUsersCurrentSlayerInfo(msg.author.id);
 		const myBlockList = msg.author.settings.get(UserSettings.Slayer.BlockedTasks);
 		const myQPs = msg.author.settings.get(UserSettings.QP);
-
 		const maxBlocks = calcMaxBlockedTasks(myQPs);
+
+		let returnMessage = '';
+
```
themrrobert, line 72: Maybe you could also add the attackable monsters' names to the 'busy' message: "Your minion is busy, but you can still manage your block list: +st blocks
You have 276 slayer points, and have completed 38 tasks in a row. XXX out of YYY MONSTER_NAME (Alternates) remaining"

src/commands/Minion/slayertask.ts
```diff
@@ -32,13 +34,42 @@ export default class extends BotCommand {
...
+		return '';
+	}
+
+	public async returnSuccess(msg: KlasaMessage, message: string, autoslay: boolean) {
+		if (autoslay) {
+			// await msg.channel.send(message);
+			try {
+				return this.client.commands.get('autoslay')!.run(msg, ['']);
+			} catch (e) {
+				return msg.channel.send('It was not possible to auto-slay this task. Please, try again.');
```
themrrobert, line 58: Remove comma from, 'Please, try again'

themrrobert (changes requested):

src/commands/Minion/slayertask.ts
kevinsmendes: The purpose of --as is to ignore all that and just train slayer. Having the first message will not be helpful for most people and will just create stress on the bot as multiple messages will be sent for little to no reasons. I left it there so it can be re-toggled as easily, if wanted.

src/commands/Minion/slayertask.ts
kevinsmendes: Thanks, changed.

src/commands/Minion/slayertask.ts
kevinsmendes: Don't think this is necessary.

src/commands/Minion/slayertask.ts
kevinsmendes: I want to give emphasis on try again, hence the comma.

PR #2174, opened 2021-06-23
Fix graceful check, simple switch

themrrobert:
### Description:
Graceful boost is applied if any one piece is worn.
New hasGracefulEquipped check didn't use every=true, so now it does, and it works fine.

**/update:** Max hood + 99 Agility hood do Not count as graceful, so I had to alter the tests accordingly.

### Changes:
Change hasGracefulEquipped: 
```                
.hasEquipped(..., false) 
```
 to : 
```
.hasEquipped(..., true) 
```

### Other checks:
-   [x] I have tested all my changes thoroughly. Even though it was a simple change, I still tested to make sure it worked as expected.

### DIFF
--- a/tests/hasGracefulEquipped.test.ts
+++ b/tests/hasGracefulEquipped.test.ts
@@ -30,11 +30,8 @@ describe('hasGracefulEquipped', () => {
 			legs: 'Hosidius graceful legs'
 		};
 		expect(hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Agility cape' }))).toEqual(true);
-		expect(
-			hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Agility cape', head: 'Agility hood' }))
-		).toEqual(false);
 		expect(hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Agility cape(t)' }))).toEqual(true);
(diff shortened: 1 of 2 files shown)


tests/hasGracefulEquipped.test.ts
```diff
@@ -32,9 +32,9 @@ describe('hasGracefulEquipped', () => {
 		expect(hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Agility cape' }))).toEqual(true);
 		expect(
 			hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Agility cape', head: 'Agility hood' }))
-		).toEqual(true);
+		).toEqual(false);
 		expect(hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Agility cape(t)' }))).toEqual(true);
 		expect(hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Max cape' }))).toEqual(true);
-		expect(hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Max cape', head: 'Max hood' }))).toEqual(true);
+		expect(hasGracefulEquipped(constructGearSetup({ ...base, cape: 'Max cape', head: 'Max hood' }))).toEqual(false);
```
gc, line 38: these tests are fairly redundant now - instead lets add a test which ensures this particular issue doesnt happen again.

themrrobert: Good call, updated.

PR #2178, opened 2021-06-23
make graceful revertible

coolbop32:
### Description:

make graceful revertible refunds 80% of the marks of grace

### Changes:

+create revert graceful gloves

### Other checks:

### DIFF
--- a/src/lib/data/createables.ts
+++ b/src/lib/data/createables.ts
@@ -1615,6 +1615,499 @@ const Createables: Createable[] = [
 			[itemID('Castle wars ticket')]: 75
 		}
 	},
+	/** Revert Graceful */
+	{
+		name: 'Revert graceful hood',
+		inputItems: {
+			[itemID('Graceful hood')]: 1
+		},
+		outputItems: {
+			[itemID('Mark of grace')]: 28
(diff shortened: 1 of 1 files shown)


gc:
I think these should be 'generated' rather than have every single graceful piece written out like this

something like this:
```js
for (const piece of gracefulTops.map(getOSItem)) {
    createables.push({{name: `Revert ${piece.name}`, input..., output...});
}
```

src/lib/data/createables.ts
```diff
@@ -1615,6 +1615,499 @@ const Createables: Createable[] = [
 			[itemID('Castle wars ticket')]: 75
 		}
 	},
+	/** Revert Graceful */
```
gc, line 1618: leaving again as a review so github knows changes are requested.

I think these should be 'generated' rather than have every single graceful piece written out like this

something like this:
```js
for (const piece of gracefulTops.map(getOSItem)) {
    createables.push({{name: `Revert ${piece.name}`, input..., output...});
}
```

gc (changes requested):

Lajnux:
Already implemented

PR #2184, opened 2021-06-24
Dinh's now prevents cannon/barrage

Lilylicious:
### Description:

This is consistent with in game.

"Furthermore, players wielding the bulwark will not be able to cast any combat spells; players attempting to do so will receive the message "Your bulwark gets in the way". When in defensive mode, the dwarf multicannon will not fire, but it will still rotate, with a filterable message stating "Your bulwark shuns your cowardliness; stopping your cannon from firing.""

### Other checks:


KyeDon:
Should look to give dinh's a use if this is being done! Boost for wildy activities maybe?

themrrobert:
Yea I'm not a fan of this. People will just stop using bulwark altogether. 

The minimal food reduction Dinh's provides  is not worth this penalty.

PR #2188, opened 2021-06-24
Cerb balancing changes

themrrobert:
### Description:
After discussions in #slayer we rebalanced items on cerb. Inq mace should be BIS, and we also decided ( with house paran/mylife/ and #slayer ) that arclight shouldn't be as strong vs cerb because there's no cost to using it, and it's so easy to obtain.

### Changes:
Cerb: 
- Added boosts for whip/tentacle/bludgeon
- Lowered boost for arclight
- Increased boost for Inquisitor's mace.

### Other checks:


coolbop32:
Whip and Tent whip should not give boosts. Cerb is weak to crush hasta and cudgel would make better sense as boosts. Arclight should also be the second best as it is 3rd in game behind scythe and mace.

PR #2196, opened 2021-06-25
Autofarm can now repeat

Lilylicious:
### Description:

+autofarm now allows repeat trips. +farm or +harvest does not.

### Other checks:


src/tasks/minions/farmingActivity.ts
```diff
@@ -169,6 +172,22 @@ export default class extends Task {
...
+					'',
+					res => {
+						user.log('continued trip of autofarming');
+						return this.client.commands.get('autofarm')!.run(res, []);
+					},
+					undefined,
+					data,
+					null
+				);
+			}
```
gc, line 190: This behaves kinda weirdly, it sends 2 messages after an autofarm now. I feel like it could be refactored a bit so a lot less code is needed, e.g.
```js
isAutofarming ? res => {
    user.log('continued trip of autofarming');
    return this.client.commands.get('autofarm')!.run(res, []);
} : undefined,
```

gc (changes requested):

PR #2200, opened 2021-06-25
SW point check and solo repeats

### LINKED ISSUE #1445: Soul Wars Zeal Check
Add a way of checking how much zeal you currently have

either a new command such as `+sw zeal`, add it to `+cl sw` or add it onto `+sw buy`
also a confirmation would be nice for purchases

currently the only way to check is by trying to buy/imbue an item, if you have under the required amount, it tells you how many you have. If you have over the required amount it auto buys with no confirmation so no way to check

Lilylicious:
* +sw no longer starts a solo, but instead shows commands and how many points you have
* SW trips now allow a repeat if only one user was on the trip. It sends +sw solo, so a +sw mass with a single user followed by a y sends a solo trip.
* SW buyables now allow multiple items. +sw buy 500 spoils of war

### Other checks:

closes #1445


src/tasks/minions/minigames/soulWarsActivity.ts
```diff
@@ -41,7 +43,23 @@ export default class extends Task {
 			user.incrementMinigameScore('SoulWars', quantity);
 			str += `${user} received ${points}x Zeal Tokens.`;
 		}
-
+		console.log(users.length);
```
gc, line 46: You have a console.log here.

gc (changes requested):

PR #2203, opened 2021-06-25
Fix duplicate names bank

Lilylicious:
### Description:

parseStringBank now allows multiple non-duplicate items to be returned from a single name in the string. 


Lilylicious:
Implemented basically the same way in a different commit by Magna

PR #2214, opened 2021-06-26
Clue continue, favalch in alch, black mask in fightcaves, and minor etc

### LINKED ISSUE #2209: Ring of the Gods (i) doesn't work as a normal Ring of the Gods for lizardman shamans.
## 🐛 Bug Report
just the title

Lilylicious:
### Description:

* You can now press Y/continueCharacter to open clue boxes you just got from a clue trip
* Solo CoX are repeatable, both CM and not.
* +alch now selects the highest value +favalch item if no item name is provided
* Black mask now boosts in fight caves even from bank, consistent with +k
* Chilli and Egg potatoes now eatable
* Ring of the Gods (i) now boosts shamans
* Grand exchange message remove notification now mentions if the reason was lack of mandatory words
* Slayer info when killing Jad now matches +k, giving info about slayer task streaks and points. Slayer XP per hour is now also shown.

### Other checks:

closes #2209


src/tasks/minions/minigames/raidsActivity.ts
```diff
@@ -108,5 +110,30 @@ export default class extends Task {
...
+
+					allUsers[0].log('continued trip of solo CoX');
+					return this.client.commands.get('raid')!.run(res, ['solo']);
+				},
+				undefined,
+				data,
+				null
+			);
+		}
 	}
```
gc, line 138: Isn't this going to cause 2 messages to be sent? you should do only sendtochannelid in masses, only handetripfinish in solos.

Also, please don't make PRs like this with lots of unrelated changes. Making 10 different PRs with small changes is a lot easier.

gc (changes requested):

src/tasks/minions/minigames/raidsActivity.ts
Lilylicious: Right you are on both counts! Future PRs won't have the same issue.

Lilylicious:
Closing this. Will create new PRs for individual features.

PR #2227, opened 2021-06-27
Fix nonexistent items appearing in +bank bug and Bank aliases

themrrobert:
### Description:
Currently, specifying a non-zero quantity in a bank query will make +bank ALWAYS show you have 1x of that item.
Example. This will show 1x Portrait even if you don't have any: 
```
+b 1 666
```

You might want to consider going back to this inside the `if (inputStr)` loop, unless you want `+b 5 cannonball` to only show 5 cannonballs.
```
bank.add(item.id, inputBank.amount(item.id))
```
Also fixed bank aliases.

### Changes:

- Added tests to prevent this issue from recurring.
- Changed the code used when parsing a filter string to use Math.max(0, ... instead of Math.max(1, ...)
- Also fixed an issue where item IDs were being interpreted as quantity, which isn't an issue now, but will be if/when ItemArray or parseStringBank is used with things like equip or spawn. It also helps when people are trading.
- Fixed Bank aliases.
- Recommend adding additional alias tests for bso for tmb/emb/hmb

### Other checks:

### DIFF
--- a/src/lib/util/parseStringBank.ts
+++ b/src/lib/util/parseStringBank.ts
@@ -1,10 +1,11 @@
 import { Bank, Items } from 'oldschooljs';
 import { Item } from 'oldschooljs/dist/meta/types';
+import { itemNameMap } from 'oldschooljs/dist/structures/Items';
 import { fromKMB } from 'oldschooljs/dist/util';
 
 import { MAX_INT_JAVA } from '../constants';
 import { filterableTypes } from '../data/filterables';
-import { stringMatches } from '../util';
+import { cleanString, stringMatches } from '../util';
 
 const { floor, max, min } = Math;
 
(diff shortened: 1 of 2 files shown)


src/lib/util/parseStringBank.ts
```diff
@@ -36,7 +39,11 @@ export function parseQuantityAndItem(str = ''): [Item[], number] | [] {
 		const item = Items.get(nameAsInt);
 		if (item) osItems.push(item);
 	} else {
-		osItems = Array.from(Items.filter(i => stringMatches(i.name, parsedName)).values());
+		osItems = Array.from(
+			Items.filter(
+				i => itemNameMap.get(cleanString(parsedName)) === i.id || stringMatches(i.name, parsedName)
+			).values()
+		);
```
themrrobert, line 46: This fixes item aliases.

src/lib/util/parseStringBank.ts
```diff
@@ -80,7 +87,7 @@ export function parseBank({ inputBank, inputStr, flags = {} }: ParseBankOptions)
 		for (const [item, quantity] of strItems) {
 			_bank.add(
 				item.id,
-				!quantity ? inputBank.amount(item.id) : Math.max(1, Math.min(quantity, inputBank.amount(item.id)))
+				!quantity ? inputBank.amount(item.id) : Math.max(0, Math.min(quantity, inputBank.amount(item.id)))
```
themrrobert, line 90: This fixes the bug where `+b 1 Twisted bow` will show 1 twisted bow even if you don't have any.

PR #2228, opened 2021-06-27
Kill inCatacombs

Lilylicious:
### Description:

Unlike +k, +kill did not pass whether or not it kills in catacombs or not to the worker, leading to never getting catacombs drops. 

### Changes:

By adding the cannon flagArg, you get a +kill outside of catacombs. Without it, you get inside catacombs.
`+kill 100k ankou --cannon` - no catacombs drops
`+kill 100k ankou` - catacombs drops

### Other checks:

### DIFF
--- a/src/commands/OSRS_Fun/kill.ts
+++ b/src/commands/OSRS_Fun/kill.ts
@@ -13,7 +13,7 @@ export default class extends BotCommand {
 			cooldown: 1,
 			oneAtTime: true,
 			description: 'Simulate killing OSRS monsters and shows the loot.',
-			usage: '<quantity:int{1}> <BossName:...str>',
+			usage: '<quantity:int{1}> <BossName:...str> [inCatacombs:boolean]',
 			usageDelim: ' ',
 			requiredPermissions: ['ATTACH_FILES'],
 			examples: ['+kill 100 vorkath', 'kill 100k bandos'],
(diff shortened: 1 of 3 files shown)


src/commands/OSRS_Fun/kill.ts
```diff
@@ -13,7 +13,7 @@ export default class extends BotCommand {
 			cooldown: 1,
 			oneAtTime: true,
 			description: 'Simulate killing OSRS monsters and shows the loot.',
-			usage: '<quantity:int{1}> <BossName:...str>',
+			usage: '<quantity:int{1}> <BossName:...str> [inCatacombs:boolean]',
```
themrrobert, line 16: Why are you specifying that the usage accepts a bool for inCatacombs if you're manually setting it later exclusively by the value of msg.flagArgs.cannon?

Please see my additional comments as well for a suggested rework.

src/commands/OSRS_Fun/kill.ts
```diff
@@ -59,7 +59,8 @@ export default class extends BotCommand {
 		const result = await Workers.kill({
 			quantity,
 			bossName,
-			limit: this.determineKillLimit(msg.author)
+			limit: this.determineKillLimit(msg.author),
+			inCatacombs: msg.flagArgs.cannon === undefined ? true : false
```
themrrobert, line 63: You aren't checking if the monster actually can be killed in the catacombs.

themrrobert (changes requested):
I personally think you should do it like this:
1) Remove the [inCatacombs] parameter
2) Default should be default, no flags.
3) You should accept thee flags: --ontask and --catacombs which would then be passed to the kill function.
4) Make sure you check the monster actually can be killed in the catacombs, because right now this will give catacombs drops to every monster, not just ones in the catacombs.

PR #2235, opened 2021-06-28
Open any, all, and until master clue

Lilylicious:
### Description:

Added three flagArgs to +open:

- --any will select the first type of openable it finds and open all of them. This is prioritized over --all if both are used.
- --all will open all openables owned by the user. 
- --master will stop opening clues if a master clue is dropped from one. This works in tandem with both previous flags, as well as opening clues regularly. It does not currently check openable types other than clues for master drops.

### Changes:

- Separated the functions for getting openable loot and sending channel messages for it. Kept the updating of settings/KC/banks to the channel message functions to avoid doing it too often.
- Added functions for any/all.
- Added master clue checking to the function to get loot from clue caskets.
- --all has a message prior to showing the loot. This is because if the user has a large amount of openables which may be the most common usecase for the function, it can take several seconds before the response is delivered.

### Other checks:


Lilylicious:
Has now been tested thoroughly by others as well, a few bugs fixed. No further bugs have been found.

PR #2237, opened 2021-06-29
Add Dagannoth Kings as a group, and the supporting infrastructure.

themrrobert:
### Description:

- Added a way to create groups of mosnters that are killable.
- It's easily extendable, so we could easily add the ability to assign weights to different monsters, for example cannoning Lizardmen also kills 'lizardman brute's.
- Increments the individual KC's of each monster in the group, as well as a KC for the group as a whole.
- Added Dagannoth Kings as the first implemented Monster Group
- Could be used in bso for some cool custom stuff.

### Changes:

- Added spawndks commaand for testing to spawn all the required gear, with --max giving max boosts as well.
- Added function setMonsterGroup to create the Monster entry based on the monsters in the group
- Added 'DagannothKings' as a slayer option to dagannoth tasks.
- Added ability to force all 3 combat styles by specifying all 4 offensive styles in the 'defaultAttackStyles' property.
- monsterActivity.ts checks if the monster if a group monster, and assigns individual kc if so.

### Other checks:

### DIFF
--- a/src/lib/minions/data/killableMonsters/bosses/groups/DagannothKings.ts
+++ b/src/lib/minions/data/killableMonsters/bosses/groups/DagannothKings.ts
@@ -1,0 +1,21 @@
+import { Monsters } from 'oldschooljs';
+
+import setMonsterGroup from '../../../../../util/setMonsterGroup';
+
+export const DagannothKingsMonsterGroup = [Monsters.DagannothRex, Monsters.DagannothPrime, Monsters.DagannothSupreme];
+
+setMonsterGroup({
+	id: 632267,
+	name: 'Dagannoth kings',
+	monsters: DagannothKingsMonsterGroup,
+	newItemData: {
+		id: 632267,
+		name: 'Dagannoth kings',
+		aliases: ['dagannoth kings', 'dks', 'dag kings']
+	},
+	randomStart: true
+});
+
+export const DagannothKings = Monsters.find(mon => mon.name === 'Dagannoth kings')!;
+
+export default DagannothKings;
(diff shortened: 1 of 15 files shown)


src/lib/minions/data/killableMonsters/bosses/groups/DagannothKings.ts
```diff
@@ -0,0 +1,21 @@
+import { Monsters } from 'oldschooljs';
+
+import setMonsterGroup from '../../../../../util/setMonsterGroup';
+
+export const DagannothKingsMonsterGroup = [Monsters.DagannothRex, Monsters.DagannothPrime, Monsters.DagannothSupreme];
+
+setMonsterGroup({
+	id: 632267,
```
themrrobert, line 8: This number came from 63 (my prefix) + 2267 (dagannoth rex's id)

src/lib/util/setMonsterGroup.ts
```diff
@@ -0,0 +1,36 @@
...
+
+		const randomizer = randomStart ? randInt(1, tables.length) - 1 : 0;
+
+		for (let i = 0; i < quantity; i++) {
+			loot.add(tables[(i + randomizer) % tables.length].kill(1, options));
+		}
+		return loot;
+	};
+}
+// Possibly change monsters to GroupMonsters[] => { weight: %, monster: SimpleMonster }
```
themrrobert, line 26: This is a suggestion for a potential upgrade if weighted distribution of kills is required.

themrrobert:
Magna didn't want it =(

PR #2242, opened 2021-06-29
Confirm commands

Lilylicious:
### Description:

Added --cf and --confirm to +capegamble, +bankbg, and +skillcape

Edit: Removed changes to bankbg and skillcape, modified capegamble to use the correct method of confirming


gc:
This isn't needed and makes no difference, unless there's a bug/problem im not aware of, when you do: `msg.confirm`, the `confirm` function already checks if theres `--cf` on the `msg`, see here: https://github.com/oldschoolgg/oldschoolbot/blob/master/src/extendables/Message/confirm.ts#L13

Lilylicious:
@gc Bug fixed, useless changes removed.

PR #2244, opened 2021-06-29
queueFn improvements

gc:
### Other checks:

### DIFF
--- a/src/extendables/User/Bank.ts
+++ b/src/extendables/User/Bank.ts
@@ -1,4 +1,5 @@
 import { User } from 'discord.js';
+import { randInt, sleep } from 'e';
 import { Extendable, ExtendableStore } from 'klasa';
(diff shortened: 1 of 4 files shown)


src/extendables/User/Bank.ts
```diff
@@ -159,31 +159,33 @@ export default class extends Extendable {
 	}
 
 	public async removeItemsFromBank(this: User, _itemBank: O.Readonly<ItemBank>) {
-		const itemBank = _itemBank instanceof Bank ? { ..._itemBank.bank } : _itemBank;
-
-		await this.settings.sync(true);
+		return this.queueFn(async user => {
```
themrrobert, line 162: You can use `this` within the function body, so not sure why you are changing the prototypes, calling convention for queue.add(), and passing user in as a variable. Let me know if i'm missing something, please :)

PR #2251, opened 2021-06-30
Potato cactus no longer shows in allotment patch

### LINKED ISSUE #2080: When farming potato's it shows as Potato Cactus in Checkpatch
## 🐛 Bug Report
When using the +farm potato command its planting potato's as expected. But when using the +checkpatch command it shows "Normal" potato's as Potato Cactus.

Schermafbeelding 2021-06-15 110105

Lilylicious:
### Description:

See title

closes #2080


themrrobert (changes requested):
Already talked to lily about this, we think there's probably a better way to solve the problem. It seems weird that checkpatches is doing string matching here.

themrrobert (approved):

PR #2253, opened 2021-06-30
Chop trips under 5

Lilylicious:
### Description:

Peky can only drop if the chop trip duration is above a constant. Due to incorrect brackets, this condition also applied to everything after it. Trips shorter than minPetDuration never gave loot and never handled the trip finish properly.

Also beaver drop code was in there twice, removed that.

### Other checks:


themrrobert (approved):
I have tested this also, and these changes are good.

gc:
Thanks!

PR #2256, opened 2021-07-01
Discordjs update

gc:
### Other checks:


gc:
NOTE: check message edit monitoring still works (e.g. g.e monitor)

gc:
also, need to merge the klasa branch and switch that

gc:
remove the changes to webhook.ts, lets not use any buttons/etc until after its merged and all that is good

gc:
also double check command perms are still ok

PR #2260, opened 2021-07-02
Change bank updates to use the Pqueue.

themrrobert:
### Description:
Updating anything I could find that updated the UserSettings.Bank directly, and changed to the User.[add,remove]Items[To/From]Bank() queued functions.

### Other changes
Also updated removeItemFromBank() to use a fresh copy of the bank in the queued function.

### Other checks:
-   [x] I have thoroughly tested this. I've tested every single page that I edited, and every location that was updated. Also had Anime, Mylife, and Titan help significantly with testing.


themrrobert:
Only did this so many times...

PR #2261, opened 2021-07-02
Seed Nest corrections

Lilylicious:
### Description:

Birds nests shouldn't roll the wyson table for seeds. This update changes birdhouses to give normal bird nest loot instead of nest  boxes. It also adds a chance to get clues from bird houses, and fixes the fallthrough mechanic of clues in the skilling clue function, since it was backwards.

### Other checks:


src/lib/simulation/birdsNest.ts
```diff
@@ -64,5 +64,4 @@ export const nestTable = new LootTable()
 	.every(birdsNestID)
 	.add(eggNest, 1, 3)
 	.add(ringNests, 1, 32)
-	.add(wysonSeedsNest, 1, 65 / 2)
-	.add(treeSeedsNest, 1, 65 / 2);
+	.add(treeSeedsNest, 1, 65);
```
themrrobert, line 67: Keep in mind thawt the stuff is 'simulation/' is global, and often used for more than just individual skills/activities.

I would suggest leaving 'nestTable' unchanged,  and making a birdhouseNestTable variable in the `src/lib/skilling/skills/hunter/birdHouseTrapping.ts` file.

themrrobert (changes requested):

src/lib/simulation/birdsNest.ts
Lilylicious: This was intentional, not just for bird house trapping. The nest table is used in adding skill clues to loot in the current codebase. This adds them to birdhouses, and makes both cases not include wyson boxes. Wyson uses a different implementation to give boxes.

src/tasks/minions/HunterActivity/birdhouseActivity.ts
```diff
@@ -56,6 +65,26 @@ export default class extends Task {
 				str = `${user}, ${user.minionName} finished collecting 4x full ${birdhouseToCollect.name}.`;
 			}
 
+			for (let i = 0; i < 4; i++) {
```
themrrobert, line 68: This should probably be a function though. 

Is there already a function for skilling clues? I know this isn't quite the same, but as long as we pass in the appropriate clue weightings and roll() chance it should be okay.  Just a thought, I don't think it's required here

themrrobert (approved):

PR #2275, opened 2021-07-04
Improve bankbg unlocks

imgidedin:
### Description:

- Users were complaining that the cost for re-applying an already unlocked bank background was too high, forcing them to be stuck with a background they could not like but had already spent too much unlocking it;
- This changes adds a new number[] field in the DB, smilar to BitFields, that stores the IDs of the unlocked Bank BGs. To unlock the bankbg, the user have to pay exactly as it is today. To re-apply the bank bg, the user will have to pay 10% of it's GPCost. The unlock does not allows users to apply bankbgs that the user is no longer possible to equip (changes in the user Patreon status or changes to bankbg/user collection log);
- The message that shows what is required for the bankbg will only show what the user is missing. Applies for CL requirements and item cost requirement.

### Changes:

- Make it so unlocked bankbgs are stored in the user db, allowing for the user to change bd without paying the huge cost of unlocking it again. This does only apply for GPCost and ItemCost.

### Other checks:

Show what user is missing from CL / Item Cost
image
image
Still requires default gold amount for unlock
image
Unlock message
image
Re-applying the bankbg
image

### DIFF
--- a/src/commands/Minion/bank.ts
+++ b/src/commands/Minion/bank.ts
@@ -2,7 +2,7 @@ import { MessageAttachment, MessageEmbed } from 'discord.js';
 import { chunk } from 'e';
 import { CommandStore, KlasaMessage } from 'klasa';
 
-import { Emoji } from '../../lib/constants';
+import { BitField, Emoji } from '../../lib/constants';
 import { UserSettings } from '../../lib/settings/types/UserSettings';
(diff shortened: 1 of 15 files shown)


gc (changes requested):
Please don't put multiple different things in 1 PR like this. You should make 1 PR for each of these 2 different things.

PR #2276, opened 2021-07-05
Add --smallbank BitField to allow for any BG to use small banks

kevinsmendes:
### Description:

- Only the default bank bg had the possibility to be shown as small and people actually liked that addition, but wanted to have their own bgs.

### Changes:

- Add `+b --smallbank` to enable or disable small banks on any bg.

### Other checks:

image
image
image
image


gc (changes requested):
Rename Allways to Always

src/lib/constants.ts
```diff
@@ -362,7 +363,8 @@ export const BitFieldData: Partial<Record<BitField, BitFieldData>> = {
 	[BitField.HasHosidiusWallkit]: { name: 'Hosidius Wall Kit Unlocked' },
 	[BitField.HasPermanentEventBackgrounds]: { name: 'Permanent Event Backgrounds' },
 	[BitField.HasPermanentTierOne]: { name: 'Permanent Tier 1' },
-	[BitField.PermanentIronman]: { name: 'Permanent Ironman' }
+	[BitField.PermanentIronman]: { name: 'Permanent Ironman' },
+	[BitField.AlwaysSmallBank]: { name: 'Allways Use Small Banks' }
```
gc, line 367: here

gc (changes requested):

PR #2301, opened 2021-07-08
Tempoross

Lilylicious:
### Description:

Added Tempoross as a solo minigame, along with all relevant uniques. XP rates relatively close to in game.

Infernal harpoon doubles your reward tokens but has significantly less boost compared to crystal harpoon.

Level: XP/hour with full boost from KC (10%) with no angler outfit
35: 22k
61 (dragon harpoon equipped): 44k
70: 51k
71 (crystal harpoon equipped): 66k
80: 75k
90: 86k
99: 96k (67k with no harpoon)

### Other checks:

-   [X] I have had the community test in the testing channels.


src/lib/data/collectionLog.ts
```diff
@@ -129,6 +129,20 @@ const wintertodtLog = resolveItems([
...
+	'Spirit angler boots',
+	'Soaked page',
+	'Fish barrel',
+	'Tackle box',
+	'Big harpoonfish',
+	'Tome of water (empty)',
+	'Tiny tempor',
+	'Dragon harpoon'
+]);
+
```
gc, line 145: You need to add the tempoross log to the minigame cl.

src/lib/data/filterables.ts
```diff
@@ -973,6 +974,11 @@ export const filterableTypes: Filterable[] = [
 		aliases: ['wintertodt', 'todt', 'wt'],
 		items: Object.values(wintertodt).flat(Infinity) as number[]
 	},
+	{
+		name: 'Tempoross',
+		aliases: ['temp', 'ross', 'tempo'],
```
gc, line 979: this doesnt have its actual name as an alias.

src/lib/simulation/tempoross.ts
```diff
@@ -0,0 +1,137 @@
...
+	.add('Uncut red topaz', [10, 15], 9)
+	.add('Uncut sapphire', [20, 30], 1)
+	.add('Uncut emerald', [15, 25], 1)
+	.add('Uncut ruby', [10, 20], 1)
+	.add('Uncut diamond', [8, 13], 1)
+	.add('Uncut dragonstone', [2, 3], 1)
+	.add('Coins', [4000, 6000], 9)
+	.add('Loop half of key', 9, 1)
+	.add('Tooth half of key', 9, 1)
+	.add('Coins', [25000, 50000], 1)
```
gc, line 52: if a number is 5 digits long (or more), you need to use a numeric separator: `25_000`

src/lib/simulation/tempoross.ts
```diff
@@ -0,0 +1,137 @@
...
+		.add('Fishing bait', [800, 1600], 100)
+		.add('Soaked page', [5, 9], 149)
+		.add('Fish barrel', 1, 20)
+		.add('Tackle box', 1, 20)
+		.add('Big harpoonfish', 1, 5)
+		.add('Tome of water (empty)', 1, 5)
+		.add('Tiny tempor', 1, 1)
+		.add('Dragon harpoon', 1, 1);
+
+	if (fishingLevel >= 81) {
```
gc, line 76: This should be more data-oriented, 

```ts
const someName = [
[79, (table) => table.add()],
[69, (table) => table.add()],
[59, (table) => table.add()]
];

for (const [level, fn] of someName) {
    if (fishingLevel >= level) {
        fn(table);
        break;
    }
}

File: src/tasks/minions/minigames/temporossActivity.ts [hunk 5]
```diff
@@ -0,0 +1,104 @@
...
+import itemID from '../../../lib/util/itemID';
+
+export default class extends Task {
+	async run(data: TemporossActivityTaskOptions) {
+		const { userID, channelID, quantity, rewardBoost, duration } = data;
+		const user = await this.client.users.fetch(userID);
+		const currentLevel = user.skillLevel(SkillsEnum.Fishing);
+		const channel = await this.client.channels.fetch(channelID);
+
+		let loot: ItemBank = {};
```
gc, line 21: this is coded weirdly: you define an empty ItemBank, then re-assign it below ,then you use the less-preferred `bankHasItem` function.

You should simply do: `const loot = getTemporossLoot()`, and then instead of `bankHasItem(loot, itemID('Tiny tempor'))`, just do `loot.has('Tiny tempor')` You shouldnt be using `ItemBank` anywhere anymore, unless its a special circumstance.

src/tasks/minions/minigames/temporossActivity.ts
```diff
@@ -0,0 +1,104 @@
...
+
+		loot = getTemporossLoot(rewardTokens, currentLevel, user.bank()).bank;
+
+		if (bankHasItem(loot, itemID('Tiny tempor'))) {
+			this.client.emit(
+				Events.ServerNotification,
+				`${Emoji.TinyTempor} **${user.username}'s** minion, ${
+					user.minionName
+				}, just received a Tiny tempor! Their Tempoross KC is ${
+					(await user.getMinigameScore('Tempoross')) + quantity
```
gc, line 35: instead of grabbing the minigame score like this here, increment their minigame score near the top, and in the return, it will tell you what their new KC is.

```ts
const {newScore} = await incrementMinigameScore();

`Your KC is now ${newScore}".
```

However, keep in mind for your pet announcement, to make it more realistic, you might consider doing this, so its possible to get a 1-kc pet for example, if they do 10 kills of tempoross, it will always say they got the pet on the 10th KC. So, instead, you can just say a random kc between their old kc and their new kc, making it possible to "get" the pet on any KC, and not just in increments of how many they kill.

e.g.:

```js
const {previousScore, newScore} = increment();
const kcTheyGotPet = randInt(previousScore, newScore);
`You got the pet on your ${formatOrdinal(kcTheyGotPet} KC.`;
```

gc (changes requested):

src/lib/data/createables.ts
```diff
@@ -601,6 +601,17 @@ const Createables: Createable[] = [
...
+	{
+		name: 'Infernal harpoon',
+		inputItems: {
+			[itemID('Dragon harpoon')]: 1,
+			[itemID('Smouldering stone')]: 1
+		},
+		outputItems: {
+			[itemID('Infernal harpoon')]: 1
+		},
+		requiredSkills: { Cooking: 85, Fishing: 75 }
```
gc, line 613: these shouldnt be titlecased.

src/tasks/minions/minigames/temporossActivity.ts
```diff
@@ -0,0 +1,103 @@
...
+			`${rewardTokens} reward pool rolls`,
+			true,
+			{
+				showNewCL: 1
+			},
+			user,
+			previousCL
+		);
+
+		if (!channelIsSendable(channel)) return;
```
gc, line 79: dont do this check, its handled already.

gc (changes requested):

PR #2321, opened 2021-07-11
Add buttons to trip handling and clues done/open

imgidedin:
### Description:

- Thought a good QoL would be to use the new buttons for trip handling and clue openings;
- That is only enabled for Tier 1 or above patreons. Non-patreons will still have the same system in place.

### Changes:

- Add a button to continue the trip;
- Adds a button to do a certain clue if the clue drops;
- Add a button to open clues when a clue trip is finished;

### Other checks:

image
image
image
image
image


src/lib/util/webhook.ts
```diff
@@ -90,11 +82,29 @@ export async function sendToChannelID(
...
+					});
+					if (componentSelection && componentSelection[selection.customID]) {
+						message.author = user;
+						componentSelection[selection.customID](message);
+					}
+					await message.edit({ components: [] });
+				} catch (e) {
+					await message.edit({ components: [] });
+				} finally {
+				}
```
gc, line 106: empty finally, maybe you meant to put the edit in here - (but still need a catch)

src/lib/util/webhook.ts
```diff
@@ -61,27 +62,18 @@ const queue = new PQueue({ concurrency: 10 });
...
 	client: KlasaClient,
 	channelID: string,
-	data: {
-		content?: string;
-		image?: Buffer | MessageAttachment;
-		embed?: MessageEmbed;
-	}
+	data: MessageOptions,
+	user: KlasaUser | undefined = undefined,
+	componentSelection?: Record<string, Function>
```
gc, line 67: Function isnt an appropriate type to use, it should show the args/return type.

gc (changes requested):
looks good, apart from these 2 little things, and apart from needing some extensive testing.

src/commands/Minion/k.ts
```diff
@@ -131,7 +131,7 @@ export default class extends BotCommand {
 
 		// Check requirements
 		const [hasReqs, reason] = msg.author.hasMonsterRequirements(monster);
-		if (!hasReqs) throw reason;
+		if (!hasReqs) return msg.channel.send(`${reason}`);
```
kevinsmendes, line 134: If anything related to a continuation trip or new slayer task ended up here, it would result in no message at all. Removed the throw and added a return msg to be more in line to what we have been doing lately.

PR #2323, opened 2021-07-11
Allow +st to show current task when minion is busy

kevinsmendes:
### Description:

- Allow +st to show current task when minion is busy.

### Changes:

- Add the current task to the message returned from +st when the minion is busy.

### Other checks:

Before:
image

After:
image

### DIFF
--- a/src/commands/Minion/slayertask.ts
+++ b/src/commands/Minion/slayertask.ts
@@ -128,6 +128,13 @@ export default class extends BotCommand {
 			const slayerStreak = msg.author.settings.get(UserSettings.Slayer.TaskStreak);
 			return msg.channel.send(
 				`Your minion is busy, but you can still manage your block list: \`${msg.cmdPrefix}st blocks\`` +
+					`${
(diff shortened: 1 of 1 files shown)


src/commands/Minion/slayertask.ts
```diff
@@ -128,6 +128,13 @@ export default class extends BotCommand {
 			const slayerStreak = msg.author.settings.get(UserSettings.Slayer.TaskStreak);
 			return msg.channel.send(
 				`Your minion is busy, but you can still manage your block list: \`${msg.cmdPrefix}st blocks\`` +
+					`${
+						currentTask
+							? `\nYour current task is to kill **${getCommonTaskName(
+									assignedTask!.monster
+							  )}**. You have ${currentTask.quantity.toLocaleString()} kills remaining.`
```
gc, line 135: You're showing `.quantity` here instead of `quantityRemaining`.

gc (changes requested):

PR #2328, opened 2021-07-12
Add special KCs to +kc

Lilylicious:
### Description:

Adds superiors, tithefarm, and hespori to +kc. Special handling for non-monster non-minigame KCs.

### Other checks:


src/extendables/User/Minion.ts
```diff
@@ -587,26 +588,46 @@ export default class extends Extendable {
...
+		outerloop: for (const [names, setting] of specialKCs) {
+			for (const name of names) {
+				if (stringMatches(name, kcName)) {
+					[specialName, specialSetting] = [name, setting];
+					break outerloop;
+				}
+			}
+		}
+
+		if (!mon && !minigame && !creature && !specialSetting) {
```
gc, line 618: I really dont think a nested loop is needed here, either do: a `.find()` or `.some()` inside the loop:

something like this
```ts
const specialKC = specialKCs.find(s => s[0].some(alias => stringMatches(alias, name))
```

src/extendables/User/Minion.ts
```diff
@@ -587,26 +588,46 @@ export default class extends Extendable {
...
 		}
 
 		const kc = mon
 			? this.getKC((mon as unknown as Monster).id)
 			: minigame
 			? await this.getMinigameScore(minigame!.key)
-			: this.getCreatureScore(creature!);
+			: creature
+			? this.getCreatureScore(creature!)
+			: await this.settings.get(specialSetting as string);
```
gc, line 628: Adding another ternary to this isn't great.. but it works okay, in the future we should refactor this mess to use early returns to avoid the need for so many ternaries.

gc (changes requested):

PR #2330, opened 2021-07-12
Non-strict string input in Dark Altar

Lilylicious:
DA was so strict about its input that people were consistently making blood runes when they intended souls. This simply makes it do souls if the input contains soul anywhere, otherwise bloods.

### Other checks:


gc (changes requested):
just do the check in the darkaltar command only, and set it to whichever, rather than changing several places/things.

Lilylicious:
Updated as requested

PR #2333, opened 2021-07-13
Collection Log Rework

kevinsmendes:
### Changes:

- Remove the old collection log system;
- Rewrote the entire collection system handling to use the new category way;
- All collection values are now exported as constants on the CollectionExports file;
- Logs can be easily attached to a role if needed (for discord roles);

### Other checks:

### WARNING: I was not able to test the roles.ts (rp roles), but the input to that function remains the same, I suspect the output will too.


src/commands/Minion/inbank.ts
```diff
@@ -0,0 +1,32 @@
...
+import { BotCommand } from '../../lib/structures/BotCommand';
+
+export default class extends BotCommand {
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			cooldown: 10,
+			aliases: ['ib'],
+			usage: '[collection:string]',
+			examples: ['+inbank bosses'],
+			description: 'Allows you to view all yout collection log items that you have in yout bank.',
```
Lilylicious, line 13: Spelling error: two instances of yout

PR #2361, opened 2021-07-17
Fix a lot of things and improve death mechanic to be as similar to OS…

kevinsmendes:
[DONE - PARTIALLY] Add Perdu and locked items to prevent players from losing them when dying (Requires LMS to buy parchment in game); --- Locked items are handled. A way to get locked items is not.
[DONE] Sort untradeables/0 value items by their alch value;
[DONE] Fix trips not continuing after inserting the correct character;
[DONE] Add smite mechanic (pure random change, 1:50);
[DONE] Add item protection;
[DONE] Handle special untradeable items (ornamented items) and other items like Abyssal tentacle);

### Other checks:

Death item protect item on (25+prayer) and having the Ava's (l) equipped
image


src/lib/util/calculateGearLostOnDeathWilderness.ts
```diff
@@ -0,0 +1,221 @@
...
+			originalItem: getOSItem(originalItem)
+		});
+	}
+
+	// 3 - Sort gear
+	const sortedGear = removableItems.sort((a, b) => b.sorter - a.sorter);
+
+	// 4 - Replicate the sorting to the main gear
+	const lostItemsBank = new Bank();
+	for (const { slot, originalItem } of sortedGear.slice(
```
Lilylicious, line 200: The ternary here for numSavedItems feels unnecessary and hard to read. Perhaps pull it out of the slice and instead define it beforehand?

PR #2363, opened 2021-07-17
[BSO] Scale tames food cost by growth stage

themrrobert:
### Description:
Scale tames' raw food intake by it's growth stage.

### Changes:
Baby's food: unchanged  (ex: 300)
Juvenile: 2/3 current.  (ex 200)
Adult: 1/3  current.  (ex 100)

### Other checks:

### DIFF
--- a/src/commands/bso/tames.ts
+++ b/src/commands/bso/tames.ts
@@ -159,11 +159,11 @@ ${allTames
 			return msg.channel.send("Your tame can't kill this monster fast enough.");
 		}
 
-		// Calculate food cost: 100% for baby, 50% for juvenile, 33% for adult
+		// Calculate food cost: 100% for baby, 67% for juvenile, 33% for adult
 		const baseFoodNeeded = monster.healAmountNeeded ?? 1;
-		// Food divisor: baby = 1, juvenile = 2, adult = 3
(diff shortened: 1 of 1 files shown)


gc:
This doesn't make sense to me. A fully grown dragon eats less food than a baby? I think it should be the opposite way around, unless im reading the code wrong.

AvariceBrah:
But if it's a dragon, that means its' serpentine. High metabolism when it's young but as it ages, it eats fillingly and spends 3 days sun basking and relaxing type shiz. Rages on nearby village on the fourth day.   

Just snake logic. Shrug I dunno just a way for that to make sense.

themrrobert:
> This doesn't make sense to me. A fully grown dragon eats less food than a baby? I think it should be the opposite way around, unless im reading the code wrong.

I'm fine with making an adult eat more than a baby if it eats a constant amount relative to time

But since the amount it eats is directly proportional to the heal amount needed, then an adult needs less food because its stronger, gets hurt less, and kills the monster faster.

Treebeards 1500 food cost for example is very high when not reduced at all.

themrrobert:
Can always revisit the idea later. I still think since food cost is based on the 'heal amount needed' that an adult should use less per kill than baby/juvenile. 

Maybe giving some other buff/defence is the answer.

PR #2365, opened 2021-07-17
Add purple item outline for new collection items on bankImage.ts

kevinsmendes:
### Description:

- Sometimes, purple numbers could be easily missed. This adds a purple outline when a purple number would be shown.

### Changes:

- Add purple outline to item that would show the purple quantity.

### Other checks:

image
image


gc (changes requested):
Put this code in canvasUtils in a standalone function, generalized to accept a certain color, and maybe a parameter for the size of the outline, or whatever is useful.

kevinsmendes:
Done. Moved it to its own function called drawImageWithOutline. Follows the same attributes as minimal drawImage, with outlineColor, outlineWidth and alpha as parameters.

PR #2378, opened 2021-07-18
Favalch in alch

Lilylicious:
### Description:

Adds favalch to alch trips with no item provided.

### Other checks:


src/commands/Minion/alch.ts
```diff
@@ -43,21 +44,33 @@ export default class extends BotCommand {
...
-			return msg.channel.send("You don't have any of this item to alch.");
+		const userBank = msg.author.bank();
+		let osItem = item?.find(i => userBank.has(i.id) && i.highalch && i.tradeable);
+
+		const [favAlchs] = msg.author.settings
+			.get(UserSettings.FavoriteAlchables)
+			.filter(id => userBank.has(id))
+			.map(getOSItem)
+			.filter(i => i.highalch > 0)
+			.sort((a, b) => b.highalch - a.highalch);
```
gc, line 55: this is duplicated in the agility alching isnt it? might be better to put this in a utility function if so, e.g. `getUsersFavAlchs` or something.

gc (changes requested):

Lilylicious:
Updated alch/laps to use a get function in the User extendable

src/lib/types/Augments.d.ts
```diff
@@ -222,6 +222,7 @@ declare module 'discord.js' {
 		queueFn(fn: (user: KlasaUser) => Promise<T>): Promise<T>;
 		bank(options?: GetUserBankOptions): Bank;
 		getPOH(): Promise<PoHTable>;
+		getUserFavAlchs(): Item[];
```
gc, line 225: if this is going to be an extendable, then it shouldnt have `User` in the name, my suggestion wasnt for an extendable but just a regular function taking the user as an arg.

gc (changes requested):

PR #2405, opened 2021-07-19
Favpatch for +af

Lilylicious:
### Description:

Added +favpatch command to favorite patch types much like +favalch favorites items to be alched. +autofarm would only plant in patches defined in +favpatch, if any are present.

### Other checks:

### DIFF
--- a/src/commands/Minion/favpatch.ts
+++ b/src/commands/Minion/favpatch.ts
@@ -1,0 +1,77 @@
+import { objectValues } from 'e';
+import { ArrayActions, CommandStore, KlasaMessage } from 'klasa';
+
+import { FarmingPatchTypes } from '../../lib/minions/farming/types';
+import { UserSettings } from '../../lib/settings/types/UserSettings';
+import { BotCommand } from '../../lib/structures/BotCommand';
+import { toTitleCase } from '../../lib/util';
+
+export default class extends BotCommand {
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			usage: '[patches:...string]',
+			description: 'Favorites a farming patch.',
+			examples: ['+favpatch herb'],
+			categoryFlags: ['minion', 'skilling']
+		});
+	}
+
+	async run(msg: KlasaMessage, [patches]: [string | undefined]) {
+		const currentFavorites = msg.author.settings.get(UserSettings.FavoritePatches);
+		let newFavorites = [...currentFavorites];
+		const input = patches?.split(', ').map(x => toTitleCase(x).replace(' ', ''));
+
+		if (!patches || patches.length === 0) {
+			if (currentFavorites.length === 0) {
+				return msg.channel.send(
+					`You have no favorited farming patches. Valid options are: ${objectValues(FarmingPatchTypes)}`
+				);
+			}
+			return msg.channel.send(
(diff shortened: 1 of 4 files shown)


src/commands/Minion/favpatch.ts
```diff
@@ -0,0 +1,77 @@
...
+			if (removed.length !== 0) str += `You removed: ${removed.toString().split(',').join(', ')}\n`;
+			if (unrecognized) str += `Unrecognized patches: ${unrecognized.toString().split(',').join(', ')}\n`;
+
+			await msg.author.settings.update(UserSettings.FavoritePatches, newFavorites, {
+				arrayAction: ArrayActions.Overwrite
+			});
+			return msg.channel.send(str);
+		}
+
+		return msg.channel.send('No change in fav patches detected. Please report this.');
```
gc, line 75: I don't like this, it should either be here not at all, or throw/log an error

src/commands/Minion/favpatch.ts
```diff
@@ -0,0 +1,77 @@
...
+			return msg.channel.send(
+				`Your current favorite farming patches are: ${currentFavorites.toString().split(',').join(', ')}.`
+			);
+		}
+
+		if (input![0] === 'clear') {
+			await msg.author.settings.update(UserSettings.FavoritePatches, [], {
+				arrayAction: ArrayActions.Overwrite
+			});
+			return msg.channel.send('Your favorite patches are cleared');
```
gc, line 39: The grammar on this sounds weird, use: `Cleared your favorite patches.`

src/commands/Minion/favpatch.ts
```diff
@@ -0,0 +1,77 @@
...
+		const input = patches?.split(', ').map(x => toTitleCase(x).replace(' ', ''));
+
+		if (!patches || patches.length === 0) {
+			if (currentFavorites.length === 0) {
+				return msg.channel.send(
+					`You have no favorited farming patches. Valid options are: ${objectValues(FarmingPatchTypes)}`
+				);
+			}
+			return msg.channel.send(
+				`Your current favorite farming patches are: ${currentFavorites.toString().split(',').join(', ')}.`
```
gc, line 31: all the `.toString().split()` in these are really weird, youre turning an array into a string, then back into an array, just remove the `.toString().split()`

src/commands/Minion/favpatch.ts
```diff
@@ -0,0 +1,77 @@
...
+			});
+			return msg.channel.send('Your favorite patches are cleared');
+		}
+
+		const patchesProvided: FarmingPatchTypes[] = input
+			?.map(patch => FarmingPatchTypes[patch as keyof typeof FarmingPatchTypes])
+			.filter(x => x !== undefined) as FarmingPatchTypes[];
+
+		const unrecognized = input?.filter(
+			patch => FarmingPatchTypes[patch as keyof typeof FarmingPatchTypes] === undefined
```
gc, line 47: maybe we can just change the typing of `patch` so these casts arent necessary.

gc (changes requested):

src/commands/Minion/favpatch.ts
Lilylicious: They're specifically there to convert "patch1,patch2,patch3" into "patch1, patch2, patch3", adding a space after commas. I wanted to avoid regex.

PR #2407, opened 2021-07-19
Boosts Dungeoneering

kevinsmendes:
### Description:

Requested by:
image
image

### Changes:

- Removed negative boost from soloing;
- Adding 10% boost for each player that joins the party;
- Changed the quantity of floors done to do the maximum allowed after all boosts/effects is applied, instead of always doing the minimum possible and returning earlier;
- Changed Scroll of Mystery to roll for every floor done and not once per trip.

### Other checks:

### DIFF
--- a/src/commands/Minion/dung.ts
+++ b/src/commands/Minion/dung.ts
@@ -319,8 +319,8 @@ export default class extends BotCommand {
 		duration = reduceNumByPercent(duration, 20);
 
 		if (users.length > 1) {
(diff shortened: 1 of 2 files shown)


gc:
Did you do XP/HR measurements?

kevinsmendes:
Should be 20% faster at solo, due to the removal of the penalty, which was highly requested. Limited trips to 5 players max (out of 12). A 5 man team should around 50% faster than it is now.

PR #2425, opened 2021-07-22
Add +use command, allow players to use a skill lamp, genie lamp, book of knowledge or dark relic

kevinsmendes:
### Description:

- Adds an use for Dark Relics and future items, like Genie Lamp, Books of Knowledge or any other possible item that RS comes up with.
- Allow Creators (Magna) to easily lock/unlock a skill from receiving XP (SotW or new skill releases) by issuing +use --add or +use --remove;
- Usage is simple, very similar to lamps, but allow multiple of the same item to be used at once, example: +use 10 dark relic, agility;
- If a skill is NOT defined, a list with all skills will appear, prompting the player to select which one they want to select.

### Changes:

- Remove lamp.ts
- Add use.ts.

### Other checks:

image
image
image
image
image


gc:
Not sure if these should be in the lamp command for consistency. So all xp-giving consumables are in a single command.

PR #2433, opened 2021-07-23
Add items bought from tokkulshop.ts to the collection log

kevinsmendes:
### Description:

- Tokkulshop is not adding items bought to the user collection log. Not sure if intended or not.

### Changes:

- Change it so, when items are added to the bank, also adds to the collection log.

### Other checks:

image
image


gc:
Items bought from a shop arent "collected", and so shouldnt go in the CL.

Alexsuperfly:
If thats what you want then you should remove all the other shops from adding to CL.  As it stands the tokkul shop is the outlier.

PR #2440, opened 2021-07-24
Corrupt gauntlet

coolbop32:
### Description:

Update time to be more accurate.

### Changes:

Made it so you can make the new weapons
Made skilling reqs same as in game
Made time more accurate

### Other checks:


gc:
Does this need to be updated to the new osjs version?

coolbop32:
> Does this need to be updated to the new osjs version?

yes

gc (changes requested):
This should be updated, and tested to make sure its alright (after having updated it)

coolbop32:
> This should be updated, and tested to make sure its alright (after having updated it)

updated

PR #2453, opened 2021-07-27
Change sellto.ts to use a new groupConfirm.ts system.

imgidedin:
### Description:

- Add buttons to sellto. Both users needs to confirm the trade. A button to cancel the trade is added, that can be clicked by both the seller and buyer.

### Changes:

- Add new groupConfirm to KlasaMessage augument;
- Change old confirm/buy to the groupConfirm system.

### Other checks:

Trade
image

Trade canceled due to time
image

Trade canceled because the buyer or seller wanted to
image

When someone accepts it
image

Success Trade
image


themrrobert:
I don't think we should do this. A lot of people that either have older phones, and some people even with modern phones and up-to-date software, are unable to use these new confirmation buttons.

We need to be backwards-compatible for at least a while. This is too new, and not fully supported.

kevinsmendes:
Added backwards compatibility. People will not be able to type `accept` to confirm their trade (including the seller) or `cancel` to cancel it (including the seller too).

PR #2457, opened 2021-07-27
Fix GGS combat EXP

### LINKED ISSUE #2423: Grotesque Guardians combat xp
Currently GGs only give combat xp for a single guardian, though they give the correct slayer xp. Combined hp for both Dawn and Dusk are 900, and should result in 3600 combat xp per kill. Currently it's only half that.

image

kevinsmendes:
### Description:

Closes #2423 

### Changes:

- Grotesque Guardians only give combat XP for 1 of the brothers;
- Changed it so, when adding XP for the monster, if the monster is GGS, it will double the Qty killed (it does not affect slayer XP).

### Other checks:

Before:
image

After
image

### DIFF
--- a/src/lib/minions/data/killableMonsters/vannakaMonsters.ts
+++ b/src/lib/minions/data/killableMonsters/vannakaMonsters.ts
@@ -560,7 +560,8 @@ export const vannakaMonsters: KillableMonster[] = [
 		],
 		healAmountNeeded: 20 * 12,
 		attackStyleToUse: GearStat.AttackSlash,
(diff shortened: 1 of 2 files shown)


themrrobert (changes requested):
Instead of this, update the miscHpHap in lib/minions/function/index.ts
like this:
```const miscHpMap: Record<number, number> = {
	9415: NIGHTMARES_HP,
	3127: 250,
	[Monsters.GrotesqueGuardians.id]: 900
};```

PR #2459, opened 2021-07-27
Blocks givetgb to self

kevinsmendes:
### Other checks:

image

### DIFF
--- a/src/commands/bso/givebox.ts
+++ b/src/commands/bso/givebox.ts
@@ -27,7 +27,7 @@ export default class extends BotCommand {
 		const difference = currentDate - lastDate;
 
 		if (difference < Time.Hour * 24 && msg.author.id !== '157797566833098752') {
-			const duration = formatDuration(Date.now() - (Number(lastDate) + Time.Hour * 24));
+			const duration = formatDuration(Date.now() - (lastDate + Time.Hour * 24));
 			return msg.channel.send(`You can give another box in ${duration}.`);
 		}
 		await msg.author.settings.update(UserSettings.LastGivenBox, currentDate);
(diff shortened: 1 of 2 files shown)


src/commands/bso/givetgb.ts
```diff
@@ -13,6 +13,7 @@ export default class extends BotCommand {
 	}
 
 	async run(msg: KlasaMessage, [user]: [KlasaUser]) {
+		if (user.id === msg.author.id) throw "You can't give boxes to yourself!";
```
gc, line 16: its a minor thing but, we always `return msg.channel.send` now and not throw, for things like this.

gc (changes requested):

src/commands/bso/givetgb.ts
kevinsmendes: Fixed it and also fixed the place I copied that from.

src/commands/bso/givebox.ts
```diff
@@ -18,14 +18,16 @@ export default class extends BotCommand {
...
+			return msg.channel.send("You can't give boxes to yourself!");
+		}
 		if (user.isIronman) return;
 		const currentDate = Date.now();
 		const lastDate = msg.author.settings.get(UserSettings.LastGivenBox);
 		const difference = currentDate - lastDate;
 
 		if (difference < Time.Hour * 24 && msg.author.id !== '157797566833098752') {
-			const duration = formatDuration(Date.now() - (lastDate + Time.Hour * 24));
+			const duration = formatDuration(Date.now() - (Number(lastDate) + Time.Hour * 24));
```
kevinsmendes, line 30: Had to add this because it was saying: `ESLint: Operands of '+' operation must either be both strings or both numbers.(@typescript-eslint/restrict-plus-operands)`, even when they are both numbers.

PR #2469, opened 2021-07-29
Allow users to type confirm/cancel for confirmation prompts

kevinsmendes:
### Description:

- Users cant type confirm anymore, and the only way to do things that require confirmation, is to use --cf

### Changes:

- Add a race promise to also check for typed messages

### Other checks:

It ignores, as it should, messages from other players
image


gc:
Personally, I really think people should just update their apps. I think its just a small minority with maybe broken/outdated versions, if they ask us for help because they cant see them, we can just tell them to reinstall/update.

Alexsuperfly:
its not just for backwards compatibility, I don't like having to grab my mouse to push the button then move it back to the keyboard to type again, if I forgot to add a `--cf` to the original command.  I think this should be added for that use case too.

PR #2479, opened 2021-07-30
Fix skilling clue distribution

themrrobert:
### Description:
Fixed skilling clues so that it distributes them correctly when a successful roll is applied. Previously you could get multiple clues per roll, or even zero clues per roll.

### Changes:
Changed the code to use a standard weighted random choice, and only allows 1 clue to be obtained per successful roll.

Make sure you remove the marked/commented `if` block when merging to BSO.

### Other checks:


src/lib/minions/functions/addSkillingClueToLoot.ts
```diff
@@ -23,6 +23,9 @@ export default function addSkillingClueToLoot(
 	const userLevel = user.skillLevel(skill);
 	const chance = Math.floor(clueChance / (100 + userLevel));
 	let nests = 0;
+	let cluesTotalWeight = 0;
+	clues.forEach(c => (cluesTotalWeight += c[1]));
```
gc, line 27: no `forEach`! your most convenient way to do this is something like: `addArrayOfNumbers(clues.map(c => c[1])`

gc (changes requested):

PR #2487, opened 2021-07-31
Allow bosses based on BossInstance to be killed more than one per trip

kevinsmendes:
### Description:

- Monsters based on BossInstance had qty fixed to one.

### Changes:

- Allow users to specify a qty to kill. If not, defaults to max.
- Add quantity, allowMoreThan1Solo and allowMoreThan1Group;
- Changed the location where the user required item is calculated;
- Changed igneActivity to allow multiple kills;
- Changed all bosses to properly account for multiple kills;

### Other checks:

Tried to use 3 as qty, but the limit per maxtrip is 2
image

Specified 1, so only 1 was killed
image

Nothing specified, kills max
image

Multiple vasa
image
image

Only 1 KG was killed
image

### DIFF
--- a/src/commands/bso/igne.ts
+++ b/src/commands/bso/igne.ts
@@ -15,6 +15,7 @@ import { formatDuration } from '../../lib/util';
 export default class extends BotCommand {
 	public constructor(store: CommandStore, file: string[], directory: string) {
 		super(store, file, directory, {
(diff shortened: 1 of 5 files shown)


gc (commented):
Seems good - we just need some thorough testing.

PR #2583, opened 2021-08-10
Fixed bug where `=spawnlamp` random XP was wrong range.

themrrobert:
### Description:
With the spawnlamp change, it would end up choosing random XP that actually spanned 3 levels:

- The entire level, 
- the entire level before it, 
- and 1 xp into the next level.

### Changes:
Fixes the XP range so that it's only the entirety of the chosen level.

### Other checks:


gc:
Thanks

PR #2622, opened 2021-08-14
Change the wrong item message on getOSItem

kevinsmendes:
### Description:

image

### Changes:

- Change the message to `The item **itemName** does not exist.`

### Other checks:


gc:
Should only do this in dev.

PR #2631, opened 2021-08-16
Music Cape

6enny6:
- Changed achievement diary cape clue boost to 5%
- Added 5% clue boost for quest cape
- Added 5% master clue boost for music cape (ty gid)
- Allowed the above capes boost's to work from bank (ty gid)
- Added music cape to cl capes

edit: 
- removed music cape from buyables so u can do whatever
- need to remove chompy chick cl part as saw u already fixed that

### DIFF
--- a/src/commands/Minion/mclue.ts
+++ b/src/commands/Minion/mclue.ts
@@ -111,6 +111,11 @@ export default class extends BotCommand {
 			duration *= 0.95;
 		}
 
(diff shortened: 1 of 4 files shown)


gc:
I have an idea for music cape which is more accurate to ingame than this (requires doing atleast 1 of every possible activity in the bot), and so it should be removed from this PR.

6enny6:
sorry but i dont think anyone wants that, sounds very tedious considering there is no list of all activities and half of osrs content is still missing

PR #2633, opened 2021-08-17
Prisma [WIP]

gc:
Just making this so I can easily read the diff.


src/lib/events.ts
```diff
@@ -14,13 +14,13 @@ FROM users
 WHERE ${maxFilter}
 ${ironman ? 'AND "minion.ironman" = true' : ''};`;
 
+// TODO VALIDATE/TEST THIS
```
themrrobert, line 17: I didn't test it, but looks good to me. Based on the prisma queryRaw docs, should return individual arrays, and psql names the column, `count` (unlike mysql) so looks to me like it should work. Still probably worth a test

src/lib/settings/prisma.ts
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	const taskName = taskNameFromType(activity.type);
+	const task = client.tasks.get(taskName);
+
+	if (!task) {
+		throw new Error('Missing task');
+	}
+
+	client.oneCommandAtATimeCache.add(_activity.user_id);
```
themrrobert, line 45: Should probably use one or the other for consistency:
Either `_activity.user_id` or `activity.userID` 
(see line 60 in the finally when the cache is deleted)

themrrobert (changes requested):

PR #2634, opened 2021-08-17
Master allow until flag k

imgidedin:
### Description:

- Allow trips to instantly finish when an item is dropped.

### Changes:

- When --clues or --until flag is used, the +k will pre-calculate the loot, just like the monsterActivity would;
- The activity will still require full supplies for the entire trip, but will return the supplies not used if ending early.

### Other checks:

### DIFF
--- a/src/commands/Minion/k.ts
+++ b/src/commands/Minion/k.ts
@@ -398,69 +398,89 @@ export default class extends BotCommand {
 		const displayDuration = duration;
 		const displayQuantity = quantity;
 		let factor = 1;
-		let seededLoot: { loot: ItemBank; superiorCount: number; returnLootCost: ItemBank } | undefined = undefined;
+		let seededLoot:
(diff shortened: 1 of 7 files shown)


themrrobert:
What about making a new command for this?

The new command would handle all the prerolling, and then if it 'hits' it calls a new, prerolledMonsterActivity, and if it doesn't hit, it just calls the normal 'k' command which generates the normal monsterActivity.

This is because I don't see this really being needed in BSO, and it's going to be a monster pain to merge and maintain going forward.

But if we separate the code the way described above, it still avoids most of the code duplication without cluttering the core k command + monsterActivity?

Like `+ku` or `+kuntil`

kevinsmendes:
> What do you guys think?

Moved everything complicated into its own function called `killUntilCalculation`. That should make it MUCH easier for merging.

PR #2643, opened 2021-08-17
Fix non-existent-item unit test

themrrobert:
### Description:

Fixed util's Unit test for non-existent-item message.

### Other checks:


themrrobert:
I'll let you fix the unit test. Windows has one version and Linux/Ubuntu has another version, so I'll let you decide how you want to do it.

PR #2665, opened 2021-08-18
Slayer Helmets CL

6enny6:
- Added a custom CL for recoloured Slayer helmets
- Added new slayer helms (from cb diary) to similaritems


src/lib/data/Collections.ts
```diff
@@ -824,6 +825,12 @@ export const allCollectionLogs: ICollection = {
 				items: slayerCL,
 				roleCategory: ['slayer']
 			},
+			SlayerHelms: {
+				alias: ['slayerHelms', 'helms', 'slayerhelmets'],
```
gc, line 829: aliases should always be lowercase, and put spaces in them.

gc (changes requested):

PR #2672, opened 2021-08-19
Fix tags, and prevent 'an error has occured' any time a command is mis-typed

themrrobert:
### Description:
Because of an error in modifying the tags command, tags stopped working, and any message that started with the bots prefix but didn't match a command would return, 'An error has occured.' This fixes both problems.
image
image

### Changes:

- Replaced the show() command which is called by the events/tags.ts class.
- Updating the settings calls to the new format.

### Other checks:


gc:
I think this was fixed already.

PR #2709, opened 2021-08-26
Add LMS Gambling +lmsg

imgidedin:
### Description:

- Was asked to make something like this a few times.

### Changes:

- Add a new `+lmsg` command, that allow users to create a LMS where people can join and compete against others (Same random system as +lms);
- An entrance fee can be set. The entrace fee is added to the prize pool. The winner takes it all.
- A minimum of 2 and maximum of 40 players is set.
- Users can change the min/max by adding another two parameters (`+lmsg 100m 10 20` means it will require 100m to enter, minimum users is 10 and auto starts at 20).

- Add a new folder to commands called Gambling;
- Add a new system for NonBusyActivity, where the minion is not considered Busy when doing;
- Currently, only add lmsGamblingActivity to the NonBusyActivity.

### To Consider:

- Currently, it can be used anywhere. Maybe force it to be used on a certain channel only?

### Other checks:

### DIFF
--- a/src/commands/Gambling/lmsgambling.ts
+++ b/src/commands/Gambling/lmsgambling.ts
@@ -3,14 +3,14 @@ import { objectEntries, objectValues, sleep, Time } from 'e';
 import { CommandStore, KlasaMessage } from 'klasa';
 
 import { toKMB } from '../../../.yalc/oldschooljs/dist/util';
-import { Activity } from '../../lib/constants';
+import { Activity, channelsPlayingLms } from '../../lib/constants';
 import { ironsCantUse, requiresMinion } from '../../lib/minions/decorators';
 import { UserSettings } from '../../lib/settings/types/UserSettings';
 import { BotCommand } from '../../lib/structures/BotCommand';
 import { LmsGamblingActivityTaskOptions } from '../../lib/types/minions';
 import { addArrayOfNumbers } from '../../lib/util';
 import addSubTaskToActivityTask from '../../lib/util/addSubTaskToActivityTask';
-import LastManStandingCommand, { channelsPlayingLms } from '../OSRS_Fun/LastManStanding';
+import LastManStandingCommand from '../OSRS_Fun/LastManStanding';
 
 export default class extends BotCommand {
 	public constructor(store: CommandStore, file: string[], directory: string) {
@@ -47,6 +47,7 @@ export default class extends BotCommand {
 			{ id: msg.author.id, name: `${msg.author.username}#${msg.author.discriminator}`, user: msg.author }
 		];
 		let started = false;
+		let canceled = false;
 
 		let contentMsg = () => {
 			const prize = usersIn.length * price;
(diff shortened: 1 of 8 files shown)


AvariceBrah:
I suggest adding a second version tbh. 

Reasoning: irons are allowed to play lms in osrs (not competitive mode) and use the lms points at Justine's shop. 

Logic; this would go against the current state (and spirit) of last bot(man?) Standing. 

Solution; 

1.create a competitive mode which is what you've already produced ; complete. 

2. Call it competitive lms. =lmsc or =clms

3. Create a near clone, that doesn't require bets (aside from the 100k coffee minimum) and does not payout gp, but instead rewards lms points.

4. Add lms shop. Maybe excluding ornate maul handle due to economical reasons. 

Conclusion; a fair inclusion for both kinds of players. Regular lms could even be added to Osb as a regular thing with competitive being a restricted command there. 

Lmk what ya think.

eggaroonie:
> I suggest adding a second version tbh.
> 
> Reasoning: irons are allowed to play lms in osrs (not competitive mode) and use the lms points at Justine's shop.
> 
> Logic; this would go against the current state (and spirit) of last bot(man?) Standing.
> 
> Solution;
> 
> 1. Create a competitive mode which is what you've already produced ; complete.
> 2. Call it competitive lms. =lmsc or =clms
> 3. Create a near clone, that doesn't require bets (aside from the 100k coffer minimum) and does not payout gp, but instead rewards lms points.
> 4. Add lms shop. Maybe excluding ornate maul handle due to economical reasons.
> 
> Conclusion; a fair inclusion for both kinds of players. Regular lms could even be added to Osb as a regular thing with competitive being a restricted command there.
> 
> Lmk what ya think.

This isn't meant to be like an actual implementation of LMS. The bot has a command that will take words you input or random users who have recently spoken and create a fake [hunger games simulator](https://brantsteele.net/hungergames/reaping.php) style game out of them. It's not an actual copy of the in-game LMS minigame, and is not meant to give the normal rewards.

AvariceBrah:
> > I suggest adding a second version tbh.
> > 
> > Reasoning: irons are allowed to play lms in osrs (not competitive mode) and use the lms points at Justine's shop.
> > 
> > Logic; this would go against the current state (and spirit) of last bot(man?) Standing.
> > 
> > Solution;
> > 
> > 1. Create a competitive mode which is what you've already produced ; complete.
> > 2. Call it competitive lms. =lmsc or =clms
> > 3. Create a near clone, that doesn't require bets (aside from the 100k coffer minimum) and does not payout gp, but instead rewards lms points.
> > 4. Add lms shop. Maybe excluding ornate maul handle due to economical reasons.
> > 
> > Conclusion; a fair inclusion for both kinds of players. Regular lms could even be added to Osb as a regular thing with competitive being a restricted command there.
> > 
> > Lmk what ya think.
> 
> This isn't meant to be like an actual implementation of LMS. The bot has a command that will take words you input or random users who have recently spoken and create a fake [hunger games simulator](https://brantsteele.net/hungergames/reaping.php) style game out of them. It's not an actual copy of the in-game LMS minigame, and is not meant to give the normal rewards.







**Ah, I was under impression of the assumption. My bad famm. Carry on!**

PR #2731, opened 2021-08-29
Rework alching & favalch

imgidedin:
### Description:

- Alching is limited to 1 item at time.
- Fav items is limited to adding/removing 1 item at time.

### Changes:

- Reworks the alching system to allow for multiple items to be alched, be it using `+alch` or on agility laps.
- Allow `+favalch` to receive multiple items as parameters.
- Allow `+favalch` now returns an image with the items (and their ID) by default. `--text` was added to allow for a text to be exported.

### Other checks:

image
image
image

### DIFF
--- a/src/commands/Minion/alch.ts
+++ b/src/commands/Minion/alch.ts
@@ -1,16 +1,16 @@
 import { Time } from 'e';
-import { CommandStore, KlasaMessage } from 'klasa';
-import { Bank, Util } from 'oldschooljs';
+import { CommandStore, KlasaMessage, KlasaUser } from 'klasa';
+import { Bank } from 'oldschooljs';
 import { Item } from 'oldschooljs/dist/meta/types';
 
 import { Activity } from '../../lib/constants';
 import { minionNotBusy, requiresMinion } from '../../lib/minions/decorators';
 import { ClientSettings } from '../../lib/settings/types/ClientSettings';
-import { SkillsEnum } from '../../lib/skilling/types';
 import { BotCommand } from '../../lib/structures/BotCommand';
 import { AlchingActivityTaskOptions } from '../../lib/types/minions';
 import { formatDuration, updateBankSetting } from '../../lib/util';
 import addSubTaskToActivityTask from '../../lib/util/addSubTaskToActivityTask';
+import { parseStringBank } from '../../lib/util/parseStringBank';
 import resolveItems from '../../lib/util/resolveItems';
 
 const unlimitedFireRuneProviders = resolveItems([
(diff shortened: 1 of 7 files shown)


src/commands/Minion/alch.ts
```diff
@@ -26,11 +26,94 @@ const unlimitedFireRuneProviders = resolveItems([
...
+	items?: { item: Item; qty: number }[];
+	qtyLimit?: number;
+	calculateByRunesOwned?: boolean;
+}) {
+	let { items, duration, user, qtyLimit, calculateByRunesOwned } = params;
+	const userBank = user.bank().clone();
+
+	if (!items) items = [];
+
+	if (items.length === 0) {
```
themrrobert, line 41: This should be able to be reworked with a map.

Something like
```ts
const favAlchs = user.getUserFavAlchs().map(fa => {
	if (userBank.amount(fa.id))
		return {
			item: fa,
			qty: qtyLimit
				? qtyLimit < userBank.amount(fa.id)
					? qtyLimit
					: userBank.amount(fa.id)
				: userBank.amount(fa.id)
		};
}).filter(fa => fa !== undefined);
const allItemsToAlch = items.length === 0 ? favAlchs : items;
```

Then use the allItemsToAlch array

src/commands/Minion/alch.ts
```diff
@@ -26,11 +26,94 @@ const unlimitedFireRuneProviders = resolveItems([
...
+		for (const i of user.getUserFavAlchs() as Item[]) {
+			// Ignore alchs that are already on the alch items
+			if (items.find(a => a.item.id === i.id)) continue;
```
themrrobert, line 44: This code block is only run if `items.length === 0`, so this shouldn't be here unless getUserFavAlchs() can return duplicate items, in which case that should probably be fixed.

src/commands/Minion/alch.ts
```diff
@@ -26,11 +26,94 @@ const unlimitedFireRuneProviders = resolveItems([
...
+			items.push({
+				item: i,
+				qty: qtyLimit
+					? qtyLimit < userBank.amount(i.id)
+						? qtyLimit
+						: userBank.amount(i.id)
+					: userBank.amount(i.id)
+			});
+		}
+		if (items.length === 0) return false;
```
themrrobert, line 54: This would have to be moved outside the if() block and changed to count the allItemsToAlch() array instead.

src/commands/Minion/alch.ts
```diff
@@ -26,11 +26,94 @@ const unlimitedFireRuneProviders = resolveItems([
...
+		if (maxCasts > maxRuneCasts) {
+			maxCasts = maxRuneCasts;
+		}
+	}
+
+	const alchBank = new Bank();
+	let totalValue = 0;
+
+	let castsLeft = maxCasts;
+	for (const i of items) {
```
themrrobert, line 85: Should probably add some sorting so the highest alch value items are alch'd first.

src/commands/Minion/alch.ts
```diff
@@ -41,85 +124,91 @@ export default class extends BotCommand {
...
-		await msg.confirm(
-			`${msg.author}, please confirm you want to alch ${quantity} ${osItem.name} (${Util.toKMB(
-				alchValue
-			)}). This will take approximately ${formatDuration(duration)}, and consume ${quantity}x Nature runes.`
-		);
+		if (!ignoreConfirm) {
+			await msg.confirm(
+				`Are you sure you want to use **${result.runes}** to alch **${
+					result.items
+				}** and convert it to <:RSGP:369349580040437770> **${result.value.toLocaleString()}** coins? This will take approximately ${formatDuration(
```
themrrobert, line 181: Do we have an emoji constant for this?

src/commands/Minion/favalch.ts
```diff
@@ -1,52 +1,79 @@
...
-		const [item] = items;
-
-		if (!item.highalch) {
-			return msg.channel.send("That item isn't alchable.");
+		const dupeCheck: string[] = [];
+		const cantAlchItems: string[] = [];
+		const removed: string[] = [];
+		const added: string[] = [];
+		for (const [item] of items) {
+			if (dupeCheck.includes(item.name)) continue;
```
themrrobert, line 50: Some items have the same name. If you want to dupe check, should probably use ID, both for memory and mechanical reasons.

src/commands/Minion/favalch.ts
```diff
@@ -1,52 +1,79 @@
...
 
-		const [item] = items;
-
-		if (!item.highalch) {
-			return msg.channel.send("That item isn't alchable.");
+		const dupeCheck: string[] = [];
+		const cantAlchItems: string[] = [];
+		const removed: string[] = [];
+		const added: string[] = [];
+		for (const [item] of items) {
```
themrrobert, line 49: Again you can refactor this with a map(item => { 
//do stuff; 
return addItem ? objToAdd : undefined;
}).filter(item => item !== undefined);

themrrobert (changes requested):
I didn't look too closely at the stuff in the last 1/3 of the diff, but I think this is a good rework.

src/commands/Minion/alch.ts
imgidedin: Why? The reason it is like that, is so we don't run that when the user just wants to alch an item. Why go over the user favs if he is trying to force alch something. This will only run if the user does `+alch` or when the user is doing alch agility.

WeeGym:
Addition to this suggestion. 
Allow an alch filter such as --alch --hv (highest gp value for stack) or --alch -- hq (highest quantity of items in a stack) 

Also --favalch --clear or something similar to delete list to start from scratch

PR #2741, opened 2021-08-31
Allow Elder Runes to be dropped from Eternal Implings

imgidedin:
### Description:

- Eternal Implings drops all runes but elder runes.

### Changes:

- Make a new version of the RuneTable, just for the impling, to add Elder Runes to the HighRunesTable.

### Other checks:

image


gc:
Did this in a different way

PR #2752, opened 2021-09-02
Fixed Webhook messages not splitting. (Affects Igne return messages)

themrrobert:
### Description:
Finally figured out why the Igne return messages weren't splitting when > 2,000 chars when it always worked in testing.... it uses a webhook where available, and it definitely exists here.

**This patch fixes that, finally people will be able to see their loot again in large teams!**

### Changes:

1. Updated the webhook.ts interaction code to use a refactored version of the `channel.send()` code. 
_(I tried just tacking on WebhookClient onto the Extendable, but they weren't compatible, also tried making a new extendable, but there was no access to the base API from the WebhookClient object, and lastly I tried to refactor the splitting into a separate function, but since it's recursive it would have meant a rewrite, so I ended up having to modify the webhook.ts code._
2. Updated the standard `channel.send()` code to also move embeds to the end as should have been done initially.

### Other checks:

-   Tested deleting the webhook and ensuring the auto-delete + failover still works.
-   Tested the igne return messages.
-   Tested regular message splits to make sure they still work properly.

### DIFF
--- a/src/extendables/TextDMChannel.ts
+++ b/src/extendables/TextDMChannel.ts
@@ -67,16 +67,17 @@ export default class extends Extendable {
 			const split = Util.splitMessage(input.content, { maxLength });
 			const newPayload = { ...input };
 			// Separate files and components from payload for interactions
-			const { components, files } = newPayload;
+			const { components, files, embeds } = newPayload;
 			delete newPayload.components;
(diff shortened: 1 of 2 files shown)


gc:
Nice work 👍

PR #2758, opened 2021-09-03
Add new settings command

imgidedin:
### Description:

Requested partially by #bso-vote: 
image

- Allow users to edit certain bitfields they own;
- Users can enable or disable the gorajan bonecrusher (Defaults to enabled).

### Changes:

- Add a new command called `=settings` that can be used to enable or disable user BFs, without requiring mods to do so.
- Add a new BitField called DisableGorajanBonecrusher.

### Other checks:

--help or no parameter will show this
image
image
image
image


gc:
This seems like something that should be going to OSB first, small bank is in both bots isnt it? plus we might want to add osb things to it anyway

PR #2770, opened 2021-09-03
Fix giveaways after bot restart

themrrobert:
### Bug Description:

- If the bot is restarted after a giveaway is started, then it fails to assign a winner after restarting. 
- Also, the message sent to the user shows `[ Object object ]`  instead of the giveaway bank's contents.

### Changes:

- Fetches the users who reacted after a bot restart; the reactions were pulled, but the users weren't cached yet.
- Converted the Giveaway loot bank into a Bank() object for the `.toString()` method for the creator's PM.

### Other checks:

-   Gidedin helped test + develop this fix


gc:
Good catch.

PR #2776, opened 2021-09-04
Buffs elder log clues when the user is 120 woodcutting

imgidedin:
### Description:

Requested by #bso-vote
image

### Changes:

- Checks if the user is 120+ wc on the woodcuttingActivity and modifies the cloned log object to increase the clue chance.
- The odd of receiving a clue is as close as possible to fishing Raw Rocktails.

- Elder logs
image

- Rocktails
image

### Other checks:

### DIFF
--- a/src/tasks/minions/woodcuttingActivity.ts
+++ b/src/tasks/minions/woodcuttingActivity.ts
@@ -21,7 +21,8 @@ export default class extends Task {
 		let xpReceived = quantity * log.xp;
 		if (logID === itemID('Elder logs')) {
 			const userWcLevel = user.skillLevel(SkillsEnum.Woodcutting);
-			if (userWcLevel >= MAX_LEVEL) clueChance = 20_434;
+			// Bring it as close as possible to Rocktails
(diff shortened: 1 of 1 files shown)


src/tasks/minions/woodcuttingActivity.ts
```diff
@@ -15,10 +15,16 @@ export default class extends Task {
 		const { logID, quantity, userID, channelID, duration } = data;
 		const user = await this.client.fetchUser(userID);
 
-		const log = Woodcutting.Logs.find(Log => Log.id === logID)!;
+		const log = { ...deepClone(Woodcutting.Logs.find(Log => Log.id === logID))! };
```
gc, line 18: I'm really not a fan of this.  just do `let clueChance = log.clueChance;` and edit that variable instead.

gc (changes requested):

src/tasks/minions/woodcuttingActivity.ts
imgidedin: Resolved.

PR #2777, opened 2021-09-04
Increase buy limit by 20x, from 250K to 5M.

imgidedin:
### Description:

Requested by #bso-vote (not many votes, but enough to make it worth)
image

### Other checks:

image


gc:
This is not a good idea for other buyable items besides feathers, what we really need is a buy limit system and just assign a big limit to feathers. I started to code it a while ago and it was just an array of transactions (not stored in the db) done, e.g. `{user: '12349', item: 1234, quantity: 123, time: timestamp}`, and then each item would have like `[Time.Hour, 1000]` = you can only buy 1000 per hour.

PR #2779, opened 2021-09-05
Allow users to claim all the hats they are missing from CL or Bank in a single command

imgidedin:
### Description:

Requested by #bso-vote
image

### Changes:

- Add a claim command do +ch, that will sell the user all the hats they are missing from CL OR BANK.
- Add some more aliases to the chompy cl to match the command to hunt chompies.

### Other checks:

image
image


src/commands/Minion/chompyhunt.ts
```diff
@@ -52,14 +52,45 @@ export default class extends BotCommand {
...
+
+		const missingHats = chompyHats.filter(c => (!userBank.has(c[0].id) || !userCl[c[0].id]) && chompyScore >= c[1]);
+
+		if (missingHats.length === 0) return msg.channel.send('There is nothing to claim.');
+		const missingHatsBank = new Bank();
+		for (const mh of missingHats) missingHatsBank.add(mh[0].id, 1);
+
+		const totalPrice = addArrayOfNumbers(missingHats.map(m => m[1] * 44));
+
+		if (msg.author.bank({ withGP: true }).amount(995) < totalPrice) {
```
gc, line 78: You shouldn't need to do `995` here, just `Coins` should work.

gc (changes requested):

PR #2782, opened 2021-09-05
Allow users to use birdhouses from bank on subsequent runs by default

imgidedin:
### Description:

- Users are forced to do --nocraft on birdhouse runs even if they already have birdhouses in their bank, when they want to 'replant' a birdhouse.
- This update automatically uses crafting or what is in bank and add the flag --craft, to force users to craft if they want that and have birdhouses in their bank.

### Changes:

- Changes how the birdhouses calculates what is required for the run.

### Other checks:

### DIFF
--- a/src/commands/Minion/birdhouse.ts
+++ b/src/commands/Minion/birdhouse.ts
@@ -187,12 +187,21 @@ export default class extends BotCommand {
 
 		let removeBank = new Bank();
 		let gotCraft = false;
-		if (!prevBirdhouse || msg.flagArgs.nocraft) {
-			for (const [item, quantity] of birdhouse.houseItemReq.items()) {
-				if (userBank.amount(item.name) < quantity * 4) {
-					return msg.channel.send(`You don't have enough ${item.name}s.`);
-				}
-				removeBank.add(item.id, quantity * 4);
+
+		const birdHouses = 4;
+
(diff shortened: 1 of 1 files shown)


imgidedin:
If anyone needs this done, please, talk with Fishy. He is against this.

Lajnux:
As stated in the #Developers channel this PR is great for BSO but it got no added benefits in OSB. 
So this PR should be made targeting the BSO branch and leave the master as is. If there are any other problems with how birdhouses work in OSB this should be done with another PR fixing those specific problems rather than reversing a functionality that been working flawless for 9 months straight. 

Reason why it's good for BSO is:
In BSO the experience gap for some custom content and beneficial good for a player running custom birdhouses is added incentive for players trading birdhouses between eachother. 

Reason why it's bad for OSB is:
The amount of birdhouses that was traded between players since hunter release in big quantities is none. There only been players trading a few birdhouses over to new players so they can do their first ever run in whatever tier they can start with normally Normal/Oak birdhouses. Therefor would this update for OSB only lead to that players that store birdhouses to give away to new players would either have to drop them or give them all away at once rather than handful to new players. Otherwise the birdhouses would be automatic consumed unless the player goes through the tideous way of having to add a flag --craft on all their commands. This also prevents collections in OSB depending on the users hunter level.

PR #2791, opened 2021-09-06
Add Channel/Roles/Support server override from config.ts

imgidedin:
### Description:

- Allowed Config.ts to override Channels, Roles and SupportServer consts;
- Moved some stuff around in the bot, to use constants instead of forced strings;
- Added all current badges to the BadgeEnum;

- Fix some `+rp roles` errors, as it was required to properly test the command.

### Other checks:


src/commands/roles.ts
```diff
@@ -45,7 +45,9 @@ export default class extends BotCommand {
 			!msg.guild ||
 			msg.guild.id !== SupportServer ||
 			!role ||
-			['342983479501389826', Roles.Moderator, '590527706974781470', '346238402737340416'].includes(role.id) ||
+			// 346238402737340416 = Bot Role
+			// 590527706974781470 = Skira user?
+			[Roles.Moderator, '590527706974781470', '346238402737340416'].includes(role.id) ||
```
gc, line 50: These comments are really weird.

src/lib/constants.ts
```diff
@@ -482,3 +493,19 @@ export const lastTripCache = new Map<
...
+		}
+	}
+}
+if (DISCORD_SETTINGS.Channels) {
+	for (const [_key, _value] of objectEntries(DISCORD_SETTINGS.Channels)) {
+		if ((Channel as any)[_key]) {
+			(Channel as any)[_key] = _value;
+		}
+	}
+}
```
gc, line 511: Cant you just do this in the declaration? `SupportChannel = Roles.SupportChannel ?? '123123123';`

gc (changes requested):

src/lib/constants.ts
imgidedin: No, you cant do that with enums.

src/commands/roles.ts
imgidedin: I thought the same thing. I have no idea why skyra is there, but at least we know what that means. I have no idea if that means skyra user or some sort of hidden role.

src/lib/constants.ts
gc: we need to do something a bit better, this looks unholy - just change it to be a plain object if needed.

imgidedin: Done, changed it from enum to an object and changing it on object init.

PR #2803, opened 2021-09-07
Add +kcgains command

imgidedin:
### Description:

- Allow users to compare their killing habits of certain monsters with other users, similar to `+xpgains`;
- Up to 10 users will show;

### Changes:

- Add date sorting to same xpgain for +xpgains;
- Add user badges to xpgains and kcgains.

### Other checks:

### DIFF
--- a/src/commands/Minion/xpgains.ts
+++ b/src/commands/Minion/xpgains.ts
@@ -31,12 +31,12 @@ export default class extends BotCommand {
 			: undefined;
 
 		const res = await this.client.orm.query(
-			`SELECT user_id as user, sum(xp) as total_xp, max(date) as lastDate
+			`SELECT user_id AS user, sum(xp) AS total_xp, max(date) AS lastDate
 FROM xp_gains
 WHERE date > now() - INTERVAL '1 ${interval}'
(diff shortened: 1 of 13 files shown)


src/commands/Minion/xpgains.ts
```diff
@@ -30,25 +31,30 @@ export default class extends BotCommand {
...
+			`SELECT user_id as user, sum(xp) as total_xp, max(date) as lastDate
 FROM xp_gains
-INNER JOIN "new_users" "new_user" ON xp_gains.user_id = "new_user"."id"
 WHERE date > now() - INTERVAL '1 ${interval}'
 ${skillObj ? `AND skill = '${skillObj.id}'` : ''}
-GROUP BY "new_user".username
-ORDER BY total_xp DESC
+GROUP BY user_id
+ORDER BY total_xp DESC, lastDate asc
 LIMIT 10;`
```
gc, line 40: uppercase on the SQL operators please. (asc), i see i didnt uppercase the `as` but if you could uppercase that too.

gc (changes requested):

PR #2843, opened 2021-09-11
Add aliases for +lb minigame

imgidedin:
### Description:

- Some minigames have names way too big or complex.
- Allows addition of aliases to the Minigame object, for easy use.

### Changes:

- Add optional `aliases` parameter to `interface Minigame`.
- Add aliases to all minigames but gauntlet. No idea what to put there.

### Other checks:

image
image

### DIFF
--- a/src/extendables/User/Minigame.ts
+++ b/src/extendables/User/Minigame.ts
@@ -7,7 +7,7 @@ import { MinigameTable } from '../../lib/typeorm/MinigameTable.entity';
 
 export interface Minigame {
 	name: string;
-	aliases?: string[];
+	aliases: string[];
 	key: MinigameKey;
 	column: string;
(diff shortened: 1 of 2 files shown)


mylife212:
`ng` for normal gauntlet and `cg` for corrupted gauntlet would be good aliases

src/extendables/User/Minigame.ts
```diff
@@ -7,6 +7,7 @@ import { MinigameTable } from '../../lib/typeorm/MinigameTable.entity';
 
 export interface Minigame {
 	name: string;
+	aliases?: string[];
```
gc, line 10: i think this shouldnt be optional

gc (changes requested):

PR #2869, opened 2021-09-16
Allow --each flag to be used with sellto

imgidedin:
### Description:

- Something that is used a lot, is whe users want to buy something for a certain price each, like, 25_000 gold ore for 5000 each. Often, certain trades would need users doing math and calculators. This PR aims to solve that by allowing users to apply a ONE price for each items in the trade.

### Changes:

- Add the flag `each` to `sellto`.
- Players can use it like this: `+sellto @gidedin 50 1000 feather, 500 flax, 600 bowstring` and the final price will be `105_000`, as it will add the amount of all items (`1000` + `500` + `600`) and multiply that by `50`.

### Other checks:

image

### DIFF
--- a/src/commands/Minion/sellto.ts
+++ b/src/commands/Minion/sellto.ts
@@ -40,9 +40,6 @@ export default class extends BotCommand {
 			rawPrice = rawPrice.replace('#', bankToSell.amount(item[0].id).toString());
 		}
 		let price = evalMathExpression(rawPrice ?? '1') ?? 1;
-		if (price < 1 || price > 100_000_000_000) {
-			return msg.channel.send('Invalid price.');
-		}
 
 		// If the each flag is set, multiply the value for each of the items being sold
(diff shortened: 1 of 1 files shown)


src/commands/Minion/sellto.ts
```diff
@@ -39,9 +39,24 @@ export default class extends BotCommand {
...
 		}
-		const price = evalMathExpression(rawPrice ?? '1') ?? 1;
+		let price = evalMathExpression(rawPrice ?? '1') ?? 1;
+
+		// If the each flag is set, multiply the value for each of the items being sold
+		if (msg.flagArgs.each) {
+			price *= bankToSell
+				.items()
+				.map(i => i[1])
+				.reduce((a, b) => a + b);
```
gc, line 49: use `sumArr` insetad of the reduce

gc (changes requested):

PR #2883, opened 2021-09-19
Add KK and Nex CL to filterables.ts

mylife212:
### Description:

Added Kalphite King and Nex to the bank filters

### Changes:

- Added nexCL to the collectionsExport list in the filterables.ts file
- Did the same with kalphiteKingCL
- Added a bank filter for both of these collection logs.

### Other checks:

### DIFF
--- a/src/lib/data/filterables.ts
+++ b/src/lib/data/filterables.ts
@@ -27,8 +27,8 @@ import {
 	cluesMasterRareCL,
 	cluesMediumCL,
 	cluesSharedCL,
-	nexCL,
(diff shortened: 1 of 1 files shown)


gc:
This needed some changes, but I fixed it in a new commit instead myself to save you from having to make changes, i made the bot automatically add a filter for every CL.

mylife212:
What changes would you recommend? I plan to add a few more bank filters in the near future (for stuff like raw food, not items found in CLs), and dont want to make the same mistake then. Thanks in advance

PR #2891, opened 2021-09-22
add cox drops to tmb

coolbop32:
### Description:

add cox drops back to tmb

### Changes:

add cox drops back to tmb

### Other checks:


Lajnux:
since it haven't been merged for such a long time I assume it won't and therefor closing it.

PR #2932, opened 2021-09-29
Add anti-venom+ and anti-venom usage to hydra

### LINKED ISSUE #2286: Allow using anti-venoms on Alchemical Hydra
Currently you need strictly Antidote++ to kill Alchemical Hydra. You should be able to use Anti-venom and Anti-venom+ too.

Extra suggestion: The amount used could be maybe decreased, currently it's using 3 whole potions per trip while 3-4 doses should be enough. (12 minutes per dose.) Maybe lower to 1 potion per trip? If the increased use is just OSB itemsink then ignore this part.

J4stMart:
### Description:

Added anti-venom+(4) and anti-venom(4) usage to hydra

Potions at hydra will be used in the following order: Anti-venom+(4) -> Anti-venom (4) -> Antidote++(4)
As this will be nicest order concerning herblore xp (although it can be argued that you'd use antidotes first in-game because they are cheaper)

Made a clear difference between itemsRequired and itemCost. 
Where itemsRequired are items that you need to own, but don't get consumed
And itemsCost are items that you need to own, but will get consumed before a trip

Closes #2286 

### Changes:

Changed monster.itemCost to be a Consumable
Added alternative consumables to Consumable (to keep backwards compatibility, but also allow anti-venoms at hydra)
Changed qtyPerMinute of anti-dote to 0.021 (1/48), because it's a 4 dose potion so a full potion lasts 48 minutes.
Changed +monster, to better display item costs
Removed itemsRequired from monsters with itemCost, as it's redundant to have the same items twice in the code

### Other checks:

-   [ X ] I have tested all my changes thoroughly.

### DIFF
--- a/src/commands/Minion/k.ts
+++ b/src/commands/Minion/k.ts
@@ -299,20 +299,7 @@ export default class extends BotCommand {
 		let pvmCost = false;
 
 		if (monster.itemCost) {
-			consumableCosts.push({
-				itemCost: monster.itemCost.clone(),
-				qtyPerKill: 1
-			});
-		} else {
-			switch (monster.id) {
-				case Monsters.Hydra.id:
-				case Monsters.AlchemicalHydra.id:
-					consumableCosts.push({
-						itemCost: new Bank().add('Antidote++(4)', 1),
-						qtyPerMinute: 0.067
-					});
-					break;
-			}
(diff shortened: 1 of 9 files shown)


themrrobert (approved):
Looks good to me, nice solution.

PR #2934, opened 2021-09-29
allow gorajan bonecrusher to be toggled off

Kefkef123:
### Description:

Adds an rp option to toggle the functionality of the gorajan bonecrusher. Please let me know if you'd like to see this implemented differently, thanks!

### Other checks:

-   [V] I have tested all my changes thoroughly.
image

### DIFF
--- a/src/commands/Owner/rp.ts
+++ b/src/commands/Owner/rp.ts
@@ -114,6 +114,25 @@ ${
 				}
 				return msg.channel.send(`Time Remaining: ${formatDuration(diff)}`);
 			}
+			case 'togglegorajanbonecrusher' : {
+				let newValue = "";
+				let bitFields = [...msg.author.settings.get(UserSettings.BitField)];
+
+				if(msg.author.settings.get(UserSettings.BitField).includes(BitField.DisabledGorajanBoneCrusher)) {
+					bitFields = bitFields.filter(field => field !== BitField.DisabledGorajanBoneCrusher);
+					newValue = "on";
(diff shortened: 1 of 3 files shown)


src/commands/Owner/rp.ts
```diff
@@ -114,6 +114,25 @@ ${
 				}
 				return msg.channel.send(`Time Remaining: ${formatDuration(diff)}`);
 			}
+			case 'togglegorajanbonecrusher' : {
+				let newValue = "";
+				let bitFields = [...msg.author.settings.get(UserSettings.BitField)];
+
+				if(msg.author.settings.get(UserSettings.BitField).includes(BitField.DisabledGorajanBoneCrusher)) {
```
themrrobert, line 121: You just pulled the setting 2 lines above this, so it should be:

```ts
if(bitFields.includes(BitField.DisabledGorajanBoneCrusher)) {
```

themrrobert (changes requested):
I don't believe =rp is the correct place for this.

Please expand into a new command, perhaps, 'playersettings.'

Keep in mind that it would be good if this new command was easily extensible to manage other flags which are often obscure or hard to remember, like the farming contract reminders or random events.

Kefkef123:
Closing this issue, I agree that there could be a better place for these kinds of settings, but that wouldn't neccesarily fit into this PR. I might be able to continue #2758, first implementing that into osb, and then having this bonecrusher functionality on top of that.

PR #2950, opened 2021-10-02
fixed hellfire bow not counting as Twisted bow

mylife212:
### Description:

There were reports of Hellfire bow (including dyed) not counting as a Twisted bow. This *should* fix it, though I'm not too familiar with it all. Probably a better solution 

### Changes:

- Added Twisted bow to similaritems.ts
- Added Hellfire bow and Hellfire bownana as similar items to the above.

### Other checks:


Bami420:
Not sure If this is still work in progress. Sorry for my impatience  if It is, but hellfire bownana still doesn’t allow me to start solo cm raids. Tried from bank and equipped in range gear. It does seem to allow me to start masses however. I can send pictures if that helps you at all.

PR #2963, opened 2021-10-04
Add Buying Multiple Items from Mahogany Homes

TastyPumPum:
### Description:
Adds ability to purchase multiple items from Mahogany Homes, useful for buying several Builders Supply Crates
### Changes:
Added an optional parameter to mh buy for the user to set quantity, and defaults to 1 if the quantity is not specified.
### Other checks:


gc:
Needs testing.

PR #2969, opened 2021-10-05
Open correct number of coffins in sepulchre

rtomek:
### Description:

Fixed number of coffins per floor on sepulchre

### Changes:

Modified sepulchreActivity.ts to pull number of coffins per floor

### DIFF
--- a/src/tasks/minions/minigames/sepulchreActivity.ts
+++ b/src/tasks/minions/minigames/sepulchreActivity.ts
@@ -25,7 +25,8 @@ export default class extends Task {
 					loot.add(GrandHallowedCoffin.roll());
 				}
 
-				const numCoffinsToOpen = floor.numCoffins;
+				const openAllCoffins = false;
+				const numCoffinsToOpen = openAllCoffins ? floor.numCoffins : 1;
 				numCoffinsOpened += numCoffinsToOpen;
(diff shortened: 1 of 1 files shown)


Lajnux:
Hey! Love to see a new coder among us, however even tho the change you made is right since some floors got more coffins.
The reason why only 1 coffin is opened is because you just don't have time to loot more coffins when running sepulchre in game. And it's currently balanced around very premium xp rates. More so than most other content in the bot. 
So with that said it currently doesn't make sense to loot more coffins than 1 per floor even if there is more coffins in some floors.

rtomek:
Got it, that makes sense. We should update the wiki with this info, then, since the sepulchre description makes it seem like the bot is broken. I also updated that section to be a little more verbose that it is intentional rather than an oversight. If you don't find it helpful, then you can feel free to close this PR.

It might be worth looking at lockpick chances. With each floor taking longer to complete, the chance to get lockpick per minute barely changes from 1 floor to 4 floors, with 4 floors being optimal. At four floors, it's still 1 per ~1800 minutes w/o bonuses, and w/bonuses it's more like 1 per 1400 minutes. Chance per minute at lockpick actually gets worse at the 5th floor, but it does add a grand hallowed coffin roll.

PR #2987, opened 2021-10-09
fix salve ammy boosts

Lajnux:
### Description:

changed salve amulet boosts for vorkath to 20 and 16% and adjusted base kill time so that the overall kill time is similar.

### Changes:

changed salve amulet boosts for vorkath to 20 and 16% and adjusted base kill time so that the overall kill time is similar.

### Other checks:


gc:
Please show what the kills per hour is before/after.

PR #3066, opened 2021-10-25
[BSO] Fix Gorajan Shard rates, and make messaging more clear.

themrrobert:
### Description:

When adding the G.Shard boost messaging, the rates were inadvertently changed. 

This code replaces the previous rates ( 2,000 base, 1,200 @ 99, 500 @ 500m ), and also makes the messaging clearer, with the actual rate shown.

### Changes:

- Fixes the rates to match the published rates in #bso-news
- Makes the messaging clearer
- Use the actual item name, 'Gorajan shards' as it's always plural.

### Other checks:

image


themrrobert:
magna didn't want to use the old drop rates

PR #3084, opened 2021-10-30
Phosani

gc:
### Other checks:


src/commands/Minion/nightmare.ts
```diff
@@ -184,6 +255,21 @@ export default class extends BotCommand {
 			} else {
 				effectiveTime *= 0.96;
 			}
+			if (isPhosani) {
+				effectiveTime = reduceNumByPercent(effectiveTime, 25);
```
themrrobert, line 259: Just a suggestion :)
I think this should be maybe 30-33%, a relatively slow phosani killl is 12 minutes. In OSRS I got a 9:10 on my 19th phosani kill. So 3 kills in 40 minutes with tablet seems a little slow. I also had none of the boosts below in game, just using godsword and bludgeon.

src/commands/Minion/nightmare.ts
```diff
@@ -193,27 +279,34 @@ export default class extends BotCommand {
...
-				});
-				totalCost.add(foodRemoved);
+		let soloFoodUsage: Bank | null = null;
+		for (const user of users) {
+			const [healAmountNeeded] = calculateMonsterFood(NightmareMonster, user);
+			const cost = this.perUserCost(user, isPhosani, quantity);
+			if (!user.owns(cost)) {
+				return msg.channel.send(`${user} doesn't own ${cost}.`);
 			}
+			let healingMod = isPhosani ? 1.5 : 1;
```
themrrobert, line 294: Sang staff should give a slight healing buff, maybe 1.4 instead of 1.5. (it should be a bigger buff for regular nightmare)

(Phosani only:) Also, with high KC, you should use less food, because most damage can be mitigated, only the final phase has unavoidable damage.  (Husks can be prayed against, but are not the priority, so you can take a small amount of damage from them, also.) That said, mistakes are common, especially with less experience. 

[Normal Nightmare prayers only block a % of  the nightmares damage, while Phosani, they block 100% because there's so many more of them]

src/lib/skilling/skills/herblore/mixables/potions.ts
```diff
@@ -451,6 +451,21 @@ const Potions: Mixable[] = [
...
+	},
+	{
+		name: 'Sanfew serum(4)',
+		aliases: ['sanfew', 'sanfew serum'],
+		id: itemID('Sanfew serum(4)'),
+		level: 65,
+		xp: 192,
+		inputItems: resolveNameBank({
+			'Super restore(4)': 1,
+			'Unicorn horn dust	': 1,
```
themrrobert, line 463: [BUG] This tab needs to be removed from the end of 'Unicorn horn dust'

themrrobert (changes requested):
Looks pretty good.

1 Little bug and 2 comments/suggestions

PR #3087, opened 2021-11-02
Adds POH to +monster

LuxrayElite:
### Description:

mainly a QOL update to include POH (pools) to be listed as a boost on the +monster command, only runs if the monster has pohboosts attribute

### Changes:

- Adds calculation of POH boosts to the trip timer
- Displays all potential boost (should probably remove similar items eg Variation of Rejuvv Pools)
- Displays current POH boost (Doesn't display anything similar to regular items when blank)
- only runs the commands when pohboosts is populated
- POH Boosts follow item boosts

### Other checks:

### DIFF
--- a/src/commands/Minion/monster.ts
+++ b/src/commands/Minion/monster.ts
@@ -94,12 +94,10 @@ export default class MinionCommand extends BotCommand {
 		
 		if (monster.pohBoosts) {
 			const POHBoosts=[];
-			const usedslots=[];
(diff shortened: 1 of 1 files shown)


LuxrayElite:
Figured out how to do linting on my end :S Apologies, applied all suggested corrections

LuxrayElite:
dupe of lily's PR from 5 months ago which did a better job :L 

Will contact lily about conflicts in their old PR

PR #3093, opened 2021-11-03
Fix "Mass failed to start, Unknown error" bug.

themrrobert:
### Affects:
- King Goldemar
- Ignecarus

### Description:
- Starting a mass sometimes fails with reason, "Unknown error"
- Logging revealed that the ReactionManager.get(ReactionEmojis.Join) was undefined
- Testing hinted that it could be a race condition
- Investigation found some async commands that were not await'ed leading to inconsistent results

### Changes:
- Change the collector.once('end') function to async
- Call start() before removing the emojiis, and await start()

### Other checks:

### DIFF
--- a/src/lib/structures/Mass.ts
+++ b/src/lib/structures/Mass.ts
@@ -173,9 +173,9 @@ export class Mass {
 				}
 			});
 
-			collector.once('end', () => {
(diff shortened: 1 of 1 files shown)


gc:
Good job.

PR #3135, opened 2021-11-12
Revert Blowpipe

6enny6:
Added `+revert toxic blowpipe`
Currently after unloading/uncharging a blowpipe you are stuck with 'toxic blowpipe' which is untradeable


LuxrayElite:
Currently allows reverting of full BP without refund, unsure how you'd go about this without getting messy

PR #3137, opened 2021-11-12
[BSO] Spawnbox Grammar Fix

6enny6:
Fixed spawnbox grammar
‘The BSO Channel’ doesnt really tell you where you can actually use it

EDIT: Pleeease could we get a command to disable commands by channel rather than globally
ie: `+rp disable dice #bso-general `


Lajnux:
the normal spawnbox no longer exist

PR #3140, opened 2021-11-13
10% boost for Zul-andra teleports when killing Zulrah

Superwhammybar:
Unsure if this was the best way to do it, I couldn't see any examples of consumableCosts to speed up kills that were optional. Happy to re-write if an example can be provided.

### Description:

Added in a case for using zul-andra teleports to speed up Zulrah kills. Only used if there are enough to cover all kills. 10% was found from 25s of travel time (house/quest cape -> fairy ring) compared to 1s from teleport.

### Changes:

Added clause in k.ts to check if the target monster is Zulrah and if the user has enough teleports to cover the kills.

### Other checks:

### DIFF
--- a/src/commands/Minion/k.ts
+++ b/src/commands/Minion/k.ts
@@ -449,6 +449,17 @@ export default class extends BotCommand {
 		}
 
 		// Boosts that don't affect quantity:
+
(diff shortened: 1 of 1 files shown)


Superwhammybar:
Needs more thought about values and optional usage before merging.

PR #3154, opened 2021-11-15
Updated herblore/secondaries/skilling filters

6enny6:
Cleaned up a few filters
Added bird nest/cactus spine/cave nightshade to secondaries


LuxrayElite:
Few feedback suggestions/ additional changes:

Herblore filterable: remove secondaries in quotes add  ...secondaries
Skilling filterable: same as above

example: 
image


Perhaps a very specific fix but change 'bird nest' to item id: 5075

This prevents the bird egg nests (which some people (myself)) from still having and aren't useable.

Tested and working

Lux~

PR #3172, opened 2021-11-19
Update buyables with Arceuus Signet

### LINKED ISSUE #3171: Add Arceuus Signet
## 🚀 Feature Proposal

Add Arceuus signet. Lately the Kourend Favour was added and the only way to obtain an arceuus signet is with 100% arceuus favour. The signet is required for POH spellbook altar. Although it has no benefit yet, it may in the future. The +POH items also claims you can build it but in reality you cannot without a signet so it’s providing false information.

TastyPumPum:
Updated the buy command to check for favour, update buyables with the arceuus signet.

Closes #3171 

Edit - Also tested the POH building of Dark Altar & Occult Altar which this now allows.


TastyPumPum:
bump should be a relatively easy review and would be nice to have, have tested everything I could think of

Issue #2572, opened 2021-08-10
[BSO] Add Mysterious ore - Available at 120 mining, buff for 500m


mylife212:
## 🚀 Feature Proposal
Add Mysterious Ore
- 120 mining req, 500m = Unlocks dwarven ore as a rare drop. Slight speed boost?
- This would unlock the Master Mining guild. Contains the coveted Mysterious ore, which unfortunately is in a vein the dwarves are too short to mine. Thankfully, you aren’t!
- Mysterious ore is basically a hard shell that contains various goodies
- Can mine 30? ore/hr, the Dwarves will break the ore open and give you the loot inside as a thank you. There is a small chance you crack the ore while mining, rendering it useless. Acts as a failed roll. 
- Exp/hr should be less than iron. 600k/hr?
EXP/HR, ORE/HR SUBJECT TO CHANGE

Loot table - Not final but to give an idea of to expect. I'm sure Magna can figure out what would make a good rate.
Always: Generic rock and...

Common 
250-750 Coal
50-150 Amethyst
250-750 Gold ore
1-50 Golden nugget
1k-5k Pure essence
20-50 Runite ore
Uncommon
50-120 Runite ore
1x Roll of Mbox table 
2x Clue scroll (master) 
Rare
Grandmaster clue (1/150)
Cut Onyx (1/100)
Corrupted ore  [from gauntlet] (1/1m) - Troll drop
ONLY IF 500M MINING+MASTERY CAPE
Dwarven ore (1/300?) Note - This would be 1 Dwarven ore every 10 hrs at 30 ore/hr. 

Goal is to give Mining some endgame content that can give variety for end-game moneymaking that isn't just geckos.

mylife212:
This also had a really good voting turn-out in #bso-vote ! 
![image](https://user-images.githubusercontent.com/63373565/128811529-2a57ef9c-73a4-43cb-bab1-87b3f455fc01.png)

Superwhammybar:
I've had a stab at this.
Here's 2 rolls of 300 ore (ignore the XP, adjusted so I could do 300 rolls at once):
![image](https://user-images.githubusercontent.com/32879863/142779415-1c3c04b6-0dad-4511-bae6-b9bf12e1f818.png)
![image](https://user-images.githubusercontent.com/32879863/142779425-42c20018-cc49-4d44-9ff9-3cdf0d4d8669.png)

Here's a normal trip with correct XP:
[...]

PR #3181, opened 2021-11-21
Bso mysterious ore

Superwhammybar:
Response to https://github.com/oldschoolgg/oldschoolbot/issues/2572

Requires changes found in https://github.com/oldschoolgg/oldschoolbot/pull/3179

### Description:

Added new 120 mining requirement ore which act like geodes, where the ore itself contains many of another mineral or reward.

### Changes:

mining.ts:
Added level 120 Mysterious Ore
Better pet chance than Runite Ore, no clue scroll chance as it has them in rewards.
XP rates are ~550k per hour. Ore per hour is ~30.

miningActivity.ts
Added condition for if the user has 500m+ mining xp and a master mining cape equipped then there is a 1/300 chance to get a Dwarven Ore per roll.
As part of the above, added a return statement to let the user know if they missed out on this chance.

### Other checks:

### DIFF
--- a/src/lib/skilling/skills/mining.ts
+++ b/src/lib/skilling/skills/mining.ts
@@ -166,7 +166,7 @@ const ores: Ore[] = [
 	},
 	{
 		level: 120,
-		xp: 3_000,
+		xp: 3000,
 		id: 508,
 		name: 'Mysterious ore',
 		respawnTime: 90,
--- a/src/lib/skilling/skills/mining.ts
+++ b/src/lib/skilling/skills/mining.ts
@@ -19,7 +19,8 @@ const ores: Ore[] = [
 		xp: 5,
 		id: 1436,
 		name: 'Rune essence',
-		respawnTime: 0.5
+		respawnTime: 0.5,
+		loot: new LootTable().add('Rune essence', 1, 100)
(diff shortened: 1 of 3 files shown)


gc:
Things like this (ideas with big implications) should be polled/discussed before being PR'ed, as its not straightforward quick decision to merge or not merge this. Personally, I don't like the "ability to mine mysterious ore" idea.

Superwhammybar:
I assumed the poll on the issue (https://github.com/oldschoolgg/oldschoolbot/issues/2572) was good enough to warrant the work. Sorry if that wasn't the case.

gc:
I'm not totally blocking the idea, but I have no interest in merging or doing this idea, mostly because I learnt my lesson from sand geckos that its not a good idea to add boxes (or similar) as drops to a gathering skill, so you just infinitely profit with no costs, and its better off dropping boxes (or clues, etc) from skills that have an item sink. People will bring in countless items to the economy farming the ores, and it will be sand geckos 2.0.

Superwhammybar:
Gotcha. I don't play BSO so I don't know the impact, I just coded the solution to the suggestion.

Feel free to close the PR.

PR #3183, opened 2021-11-22
BSO - Fixed output crop plants yield fix

### LINKED ISSUE #2967: Farming Mushrooms
## 🐛 Bug Report
Farming mushroom patch, I receive 24 mushrooms no matter how many patches I plant. I have scroll, and master farmer + plopper. Planting 6 patches. Using ultracompost. 
Also running into and issue with =af where I can plant Limpwurt, but it says wait x time - usually reflects my allotment time remiang on this bug

Superwhammybar:
closes #2967 

### Description:

If a plant has a fixed crop yield (and doesn't require chopping) it will only return the yield of one plant, not the total for each alive plant.

This bug is also present in OSB but the only affected plants (Mushroom and Calquat Tree) only have 1 patch to farm anyway so multiplication isn't an issue. Should probably be fixed too though.

### Changes:

Multiply the fixedOutputAmount by the number of alivePlants

### Other checks:

### DIFF
--- a/src/tasks/minions/farmingActivity.ts
+++ b/src/tasks/minions/farmingActivity.ts
@@ -318,13 +318,9 @@ export default class extends Task {
 					harvestXp = 0;
 				} else if (plantToHarvest.givesCrops && chopped) {
 					if (!plantToHarvest.outputCrop) return;
-					// cropYield already counts each patch if variableYield is true:
-					loot[plantToHarvest.outputCrop] =
-						plantToHarvest.fixedOutput && plantToHarvest.fixedOutputAmount
-							? plantToHarvest.fixedOutputAmount * alivePlants
-							: cropYield;
-
-					harvestXp = cropYield * alivePlants * plantToHarvest.harvestXp;
+					
+					loot[plantToHarvest.outputCrop] = cropYield;
+					harvestXp = cropYield * plantToHarvest.harvestXp;
 				}
 			}
 
--- a/src/tasks/minions/farmingActivity.ts
+++ b/src/tasks/minions/farmingActivity.ts
@@ -318,7 +318,7 @@ export default class extends Task {
 					harvestXp = 0;
 				} else if (plantToHarvest.givesCrops && chopped) {
 					if (!plantToHarvest.outputCrop) return;
-					
+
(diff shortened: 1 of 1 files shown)


themrrobert (changes requested):
This won't work. Watch what happens if you try to harvest fruit trees with this patch,  you will end up getting way too many.  (It will multiply by  the number of patches twice)

The farming code needs a serious rework, it is the most spaghetti code in the entire bot, I believe.

Superwhammybar:
@themrrobert You're right, it's a bit spaghetti here.

Looking at the issue you've raised, it seems that the issue is related to harvesting xp? The crop yield was correct when I tested with apple trees but I did end up with more harvesting xp than expected.

This seems to stem from the if statement at 319 where it assumes cropYield hasn't been calculated correctly and multiplies it again by number of alive plants.

I'll see if I can figure, and test, a solution.

Superwhammybar:
I've amended the code to separate the calculation of crop yield from the calculation of xp.

This change has been tested against every patch type (although it only affects plants which need to be chopped, and give fixed yields). The harvesting xp is now as expected, and the yield is correctly calculated.

themrrobert:
This definitely looks better; although I'm not able to give it a proper review currently. 

However, this really should be PR'd to master instead of BSO to minimize merge conflicts in the future, and to support possible future expansion in OSRS.

One more thing to test would be fruit (and regular) trees without the WC level to chop them. There was a bug previously with this.

Lastly, the settings are per-plant, not per patch, so maybe a little more testing, both with + without WC level.

Make sure you test the new fruits (lychee is a good one;  they all have the same settings so no need to test avocado + mango also). Since you're already on BSO, best to test this first before migrating the patch to OSB (master)

Superwhammybar:
Thanks for the feedback.

Avocado was one of the random ones I chose when testing. Since then I've tested it with 3 random fruit trees and 3 random trees all with and without the WC level. Although fruit trees have a required level of 1 so that wasn't possible to have a without.

My problem now is that apparently there are 5 issues according to DeepScan, and the unit tests fail. But the unit tests work when I run them locally and the 5 issues aren't any files I've touched...

PR #3192, opened 2021-11-24
Fix +checkmasses with new Prisma update

themrrobert:
### Description:
checkmasses was failing because the dates in the returned activities were in raw string form and not Date() objects, this corrects that problem.

### Changes:

Changed query from raw into a filtered query, converted it into a flat activity in order to convert the data into a usable form, and updated the variable references in the string generation.

### Other checks:


src/commands/Utility/checkmasses.ts
```diff
@@ -18,29 +18,33 @@ export default class extends BotCommand {
...
+					channel_id: { in: channelIDs }
+				},
+				orderBy: {
+					finish_date: 'asc'
+				}
+			})
+		).map(convertStoredActivityToFlatActivity);
 
-		masses = masses.filter(m => m.data.users.length > 1);
+		masses = masses.filter(m => m.users.length > 1);
```
gc, line 34: just chain this after the map instead of re-assigning it.

src/commands/Utility/checkmasses.ts
```diff
@@ -18,29 +18,33 @@ export default class extends BotCommand {
...
 		const massStr = masses
-			.map(
-				m =>
-					`${m.type}${m.data.challengeMode ? ' CM' : ''}: ${m.data.users.length} users returning to <#${
-						m.channel_id
-					}> in ${formatDuration(m.finish_date.getTime() - now)}`
-			)
+			.map(m => {
+				console.log(m);
+				console.log(m.finish_date);
```
gc, line 43: logs

gc (changes requested):

PR #3204, opened 2021-11-25
Fishing alias

### LINKED ISSUE #1716: +m fish barb
+m fish barb 
should work for
+m fish barbarian

TastyPumPum:
Added Aliases for fishing

Closes #1716


src/lib/skilling/skills/fishing.ts
```diff
@@ -133,6 +133,7 @@ const fishes: Fish[] = [
 		xp: 100,
 		id: itemID('Raw swordfish'),
 		name: 'Swordfish',
+		alias: ['Sword'],
```
gc, line 136: our convention is to always have aliases lowercased.

gc (changes requested):

PR #3251, opened 2021-12-07
Skilling Outfit Boost Standardisation

### LINKED ISSUE #3249: Lumberjack doesn't give its exp boost.
## 🐛 Bug Report

Equipped lumberjack doesn't appear to give its boosts in any setup.

Superwhammybar:
Prompted by, and closes #3249 

### Description:

Make it clearer to a user why they are or are not benefiting from a given skilling outfit.
As part of this, a rework was done to standardise the way in which the boost percentage was calculated.
This also fixes a bug where the full set was searched for within the Skilling outfit, but the subsequent searches for individual parts of the outfit were looking anywhere.

### Changes:

util.ts:
Added constants for definitions of each outfit
Added a generic function that takes both a user and an outfit type and returns a calculated percentage boost based on the equipment in the Skilling gear slot.

util.test.ts:
Added test for new function, adapted from calcConBonusXP.test.ts.

calcConBonusXP.test.ts & calcConBonusXP.ts:
Removed, the generic util function covers these now.

src/commands/... :
Added new generic function
Added message to user about % XP gain from outfit, shows even at 0% to indicate that the outfit should be in the Skilling gear tab. No ambiguity.
Fixed a bug where you gained no extra XP in Volcanic Mine unless you had the full set.

src/tasks/... :
Added new generic function
Replaced old method of calculating gear boost which, as mentioned up top, was unintentionally (I assume) allowing outfit pieces to be equipped anywhere.
Fixed a bug in aerialFishingActivity.ts and fishingActivity.ts where bonus xp was double-counted.
Fixed a bug where you gained no extra XP in volcanicMineActivity.ts unless you had the full set.

### Other checks:

I tested the calculation before and after each change (commenting out any variant factors like randomFloat multiplication) as I went through and then tested them all at the end to make sure the bonus XP listed was correct when compared to the total XP expected.


LuxrayElite:
Checked using a local host bot.

Raised and solved issue: Pyromancer uses in wintertodt.

No other issues were found.

~Lux

Issue #3248, opened 2021-12-07
Hellfire arrows disappearing


Qwertyweezel:
## 🐛 Bug Report

I was trying to figure out how to get my borrowed hellfire bow to work for malygos, so I equipped the bow and 333 arrows to my wilderness setup. Couldn't get it working, so I =unequip wildy hellfire arrow, which deleted all but 1 hellfire arrow, which ended up in my bank. I believe the problem occured when I didn't type how many arrows to remove.

mylife212:
could you post pictures of the bug? IE: Equipping the arrows, unequipping them and showing they're deleted

themrrobert:
I'm unable to replicate this.

I suspect the arrows were used when attempting to kill something in the wildy with =k (possibly not wearing the ranged master cape or avas)

You cannot use a qty in the unequip command, it removes all of them.

If you can post screenshots of everything, from equipping the arrows to attempting to kill maly/etc, then I can look farther into it, it will help me narrow down what to look at. (I need to see exactly how 'it didn't work' when you tried to kill maly, etc)

Qwertyweezel:
![Uploading Screenshot_20211207-104036.jpg…]()

Qwertyweezel:
![Uploading Screenshot_20211207-104250.jpg…]()

Qwertyweezel:
![Uploading Screenshot_20211207-104303.jpg…]()

Qwertyweezel:
![Screenshot_20211207-104321](https://user-images.githubusercontent.com/90746611/145061643-c9e6ad0f-3769-4fa1-89c8-7a6b5571ecc1.jpg)
![Uploading Screenshot_20211207-104326.jpg…]()
![Uploading Screenshot_20211207-104500.jpg…]()

Qwertyweezel:
![Screenshot_20211207-104333](https://user-images.githubusercontent.com/90746611/145061667-72bf3326-d31c-4815-8173-7a832fbc000b.jpg)
![Screenshot_20211207-104509](https://user-images.githubusercontent.com/90746611/145061711-33a3df3e-b9c1-4665-9713-a7cfe856ff3e.jpg)
![Screenshot_20211207-104353](https://user-images.githubusercontent.com/90746611/145061712-60d2e7e4-defe-4269-a080-657a1dfbd3ad.jpg)
[...]

PR #3253, opened 2021-12-08
Fix bug where autoequip deletes stackable items

themrrobert:
### Description:

If you have a stack of items equipped to an equipment slot, then run autoequip on the same slot, autoequip deletes the entire stack (except 1) if you have the same item stack type equipped as autoequip would also equip.

### Changes:
If toRemoveFromGear exists in toRemoveFromBank, then instead of deleting the entire stack, only delete 1 if toRemoveFromGear(itemID) > 1

### Other checks:

Fixes [3248](https://github.com/oldschoolgg/oldschoolbot/issues/3248)


gc:
we really need a test for this!

Issue #3222, opened 2021-12-01
[OSB] Inferno % Chance


FatOrangeKitty:
## 🚀 Feature Proposal
A command that can show your Death Chance % of both waves and Zuk, without having to start a new run.

Something like +i chance

PR #3257, opened 2021-12-09
Add --chance flag for inferno start command

Superwhammybar:
Closes https://github.com/oldschoolgg/oldschoolbot/issues/3222

### Description:

Added a --chance flag to +inferno start command. This just shows the user their Boosts, pre-Zuk death chance and Zuk death chance.

### Changes:

Added a check for --chance flag and return message early (before supply usage).
Moved the construction of parts of the return message to earlier in the file to accommodate.

### Other checks:

Checked that the Inferno command with --chance flag doesn't take any items.
Checked that the information returned is the same between --chance flag and not.
Checked that inferno runs start as expected without the flag, removing and telling the user of consumables taken.


src/commands/Minion/inferno.ts
```diff
@@ -597,6 +597,27 @@ AND (data->>'diedPreZuk')::boolean = false;`)
...
+				: `*(You didn't get these: ||${preZukDeathChance.missed.join(', ')}||)*`
+		}
+**Zuk Death Chance:** ${zukDeathChance.value.toFixed(1)}% ${zukDeathChance.messages.join(', ')} ${
+			zukDeathChance.missed.length === 0
+				? ''
+				: `*(You didn't get these: ||${zukDeathChance.missed.join(', ')}||)*`
+		}`;
+
+		if (msg.flagArgs.chance) {
+			return msg.channel.send(`${boosts}${deathChances}`);
```
gc, line 618: This should probably be in the base command (`+inferno`) instead of being a flag, in the command to start one?

gc (changes requested):

src/commands/Minion/inferno.ts
Superwhammybar: I can move it into the base command but it would mean also copying up the code to both do an inferno run and parse/format the boosts and chances returned.

Since the flag is only intended to show the death chances, not the boosts (it was just there and nice to have), I will move the flag to the base command and call the death chance calculations from there.

Superwhammybar: Nevermind, looking at the code I'd be skipping all of the gear calculations that affect death chance by doing this.

I'll move it as you suggested and we can re-review.

gc:
I think we should also mention these in the base command so that people know it exists.

Superwhammybar:
> I think we should also mention these in the base command so that people know it exists.

Added a message into the base command to alert the user of the --chance flag.

TastyPumPum:
Bump

PR #3259, opened 2021-12-09
Make Trailblazer Graceful part of the Outfit

TastyPumPum:
Added the trailblazer graceful items to the relevant collections export so they provide the relevant boosts for having graceful equipped. Merged to BSO as trailblazer is currently unobtainable on OSB and doubt it ever will be. 


themrrobert:
Maybe retarget this patch to `master`? 
OSB doesn't have trailblazer yet, but it is something we'd like to implement as rewards from playing BSO in a leagues-like format.

Also, if it's in master, then it won't accidentally be overwritten (again) as this is not the first time this has been fixed. XD

Probably ask magna what he thinks about putting it in master before closing this

Issue #1332, opened 2021-02-09
Alchemy Runes


osrsolympus:
## 🐛 Bug Report

Casting high alchemy doesn't say it uses fire runes without a source of infinite fire runes equipped. I own staves which provide the affect, as well as a time of fire, but none of them are equipped in any gear setup, only a mud battlestaff. The message only says you'll consume nature runes.

mylife212:
fixed in next update

PR #3270, opened 2021-12-11
State fire rune cost (Alching)

LuxrayElite:
### Description:
Solves issue #1332
Mentions if you are using fire runes or not, 
no fire rune cost is mentioned when using unlimited sources

### Changes:

An added str variable which gets populated with the fire rune cost if over 0

State fire rune cost before nature runes if cost over 0 on confirmation message/cost

### Other checks:


src/commands/Minion/alch.ts
```diff
@@ -82,6 +82,10 @@ export default class extends BotCommand {
 				break;
 			}
 		}
+		let str = [];
+		if (fireRuneCost > 0) {
+			str.push(`${fireRuneCost}x Fire rune, `);
+		}
```
gc, line 88: This is weird making an array just to put 1 thing into it, you could use string concatenation instead below:

`${fireRuneCost > 0 ? `${fireRuneCost}x Fire rune` : ''}`

gc (changes requested):

Issue #2784, opened 2021-09-06
Agility master cape perk doesn’t work


sn0w1753:
Bug 🐛 BSO

Not sure if the perk of the agility master cape is supposed to be full graceful or not. But was under the impression of that. Cape doesn’t work in most activities that would show a boost for having full graceful no matter which equipment slot it’s in. Final decider of it definitely not working as intended was getting a -7.5% for not wearing full graceful while doing blast furnace no matter which slot the cape is in-or bank.

sn0w1753:
![FA4EFA53-A3A9-4688-BAAB-BE405227CB61](https://user-images.githubusercontent.com/78341255/132155729-bcd07ba1-9ffa-4f34-a5cf-eeb590165f9a.jpeg)

sn0w1753:
![C7861708-487A-4011-8C03-F3647CC9E28F](https://user-images.githubusercontent.com/78341255/132178269-dee68294-a116-40bb-84f4-698b53fb9702.jpeg)

mylife212:
Fixed in next update

PR #3272, opened 2021-12-11
Agility master cape acts as graceful cape

LuxrayElite:
### Description:

Solves #2784
Assumes master cape acts similar to max and skill cape

Maybe worth adding relevant expert cape if it is to provide the same effect

### Changes:

added 'agility master cape' to gracefulcapes

### Other checks:
Untested but assumed working


gc (changes requested):
Failing ESLint

PR #3276, opened 2021-12-11
Tears of Guthix

Superwhammybar:
### Description:

Add in the Tears of Guthix minigame.
This requires both a minion and the quest requirements for the quest Tears of Guthix. Ironmen have additional skill requirements for the quest, so that is also reflected here. This is something that could be removed as there is a way to buy a bullseye lantern (which is where their requirements stem from) if an ironman does Death to the Dorgeshuun, which could be assumed completed within the 43 QP required for Tears of Guthix.
The minigame can only be done once per week and a timestamp of the last go is stored in the database.
The minigame has scaling XP rewards, as per the wiki, from 10XP per tear (when their lowest skill is level 1) to 60XP per tear (when their lowest skill is level 30).

Tear collection calculation:
The number of QP a minion has is the amount of game ticks the minigame can be played for.
Streams of blue tears are static for 9 seconds now (15 ticks).
QP / 15 gives us how many full streams a minion could try take from.
The minion can collect between 40% and 95% of each stream.
For max QP (284) the maximum tears collected is 257, the minimum is 108.
For minimum QP (43) the maximum tears collected is 29, the minimum is 12.

### Changes:

Add lastTearsOfGuthixTimestamp to prisma schema
Add tears of guthix command
Add tears of guthix task
Add snake emoji for Luna
Add tearsOfGuthixActivity to task list and activity definition
Add Tears of Guthix to minigame list

### Other checks:


src/commands/Minion/tearsofguthix.ts
```diff
@@ -0,0 +1,109 @@
...
+			return msg.channel.send(
+				`You are not skilled enough to participate in Tears of Guthix. You need the following requirements: ${objectEntries(
+					skillReqs
+				)
+					.map(s => {
+						return msg.author.skillLevel(s[0]) < s[1]
+							? formatSkillRequirements({ [s[0]]: s[1] }, true)
+							: undefined;
+					})
+					.filter(f => f)
```
gc, line 70: do `.filter(notEmpty)`.

src/commands/Minion/tearsofguthix.ts
```diff
@@ -0,0 +1,109 @@
...
+		const skillReqs = {
+			[SkillsEnum.Firemaking]: 49,
+			[SkillsEnum.Crafting]: 20,
+			[SkillsEnum.Mining]: 20
+		};
+		const ironmanExtraReqs = {
+			[SkillsEnum.Smithing]: 49,
+			[SkillsEnum.Thieving]: 36,
+			[SkillsEnum.Slayer]: 35
+		};
```
gc, line 37: Move these outside of the command, and maybe (not sure if it will work) `as const` them to prevent the need for any casting.

src/tasks/minions/minigames/tearsOfGuthixActivity.ts
```diff
@@ -0,0 +1,63 @@
...
+		for (let stream = 0; stream < streams; stream++) {
+			const percentCollected = randInt(40, 95); // Collect 40 - 95% of each stream
+			tears += Math.ceil(15 * (percentCollected / 100));
+		}
+
+		// Calculate tear value
+		const baseXPperTear = 10;
+		const xpPerTearScaling = 50 / 29;
+		const xpScalingLevelCap = 30;
+		const skillLevel = user.skillLevel(lowestSkill as SkillsEnum);
```
gc, line 43: cast it at the source instead of doing it here and below too

gc (changes requested):

src/commands/Minion/tearsofguthix.ts
Superwhammybar: Moved them but `as const` didn't help with having to use cast later.

gc (changes requested):

src/commands/Minion/tearsofguthix.ts
```diff
@@ -0,0 +1,109 @@
...
+const skillReqs = {
+	[SkillsEnum.Firemaking]: 49,
+	[SkillsEnum.Crafting]: 20,
+	[SkillsEnum.Mining]: 20
+};
+const ironmanExtraReqs = {
+	[SkillsEnum.Smithing]: 49,
+	[SkillsEnum.Thieving]: 36,
+	[SkillsEnum.Slayer]: 35
+};
```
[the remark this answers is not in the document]
gc: nevermind.

PR #3285, opened 2021-12-13
Agility Arena XP purchase clarification

### LINKED ISSUE #3283: Brimhaven Agility Area Ticket Exchange
## 🐛 Bug Report

When doing =aa buy or +aa buy the bots suggest you should do =agilityarena buy xp 5 or +agilityarena buy xp 5. When attempting to do this command the bots then tell you, "You can only redeem tickets for XP at the following quantities: 1, 10, 25, 100, 1000.". Requesting to have the amount the bot suggest to either 10 or make 5 a possible quantity you could do.

Superwhammybar:
Closes #3283 

### Description:

Example used to be -agilityarena buy xp 5, but 5 is not a valid choice.

### Changes:

Replaced the fixed value of 5 with a pipe-delimited list of available options.

### Other checks:

### DIFF
--- a/src/commands/Minion/agilityarena.ts
+++ b/src/commands/Minion/agilityarena.ts
@@ -134,9 +134,9 @@ export default class extends BotCommand {
 
 Alternatively, you can convert tickets to XP (+10% XP for Karamja Medium Diary) using \`${
 					msg.cmdPrefix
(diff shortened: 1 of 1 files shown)


Superwhammybar:
Here's how it used to look:
image

Here is how it looks now:
image

PR #3288, opened 2021-12-13
Christmas event

gc:
### Other checks:


src/tasks/minions/christmasActivity.ts
```diff
@@ -0,0 +1,55 @@
...
+		const table = action === 'steal' ? stealItems : deliverItems;
+
+		const cl = user.cl();
+		for (const item of table) {
+			if (cl.has(item)) break;
+			loot.add(item);
+		}
+
+		if (!cl.has('Antisanta mask') && action === 'steal') {
+			loot.add(antiSantaOutfit);
```
themrrobert, line 35: Same as above with regards to giving out the outfit fairly for BSO players
(See code snippet near the top)

src/tasks/minions/christmasActivity.ts
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		if (roll(120)) {
```
themrrobert, line 38: This rate might want to be looked at. 

Rule of thumb is, whatever the chance is, that many attempts is only a 63% chance to get the item.

Here, 200 attempts (100 hours) @ 1/120 is only an 82% chance to get 1 item.

Obviously some people will be much luckier, and the distribution overall will be 1 per 120, but thats because 'lucky' players get multiple,  while unlucky players are **very** unlucky. I would make this just a **little** more generous, considering how close X-mass is  and how busy people are.

src/commands/christmas.ts
```diff
@@ -0,0 +1,143 @@
...
+				]
+			});
+		}
+		const delivered = msg.author.settings.get(UserSettings.PresentsDelivered);
+		const stolen = msg.author.settings.get(UserSettings.PresentsStolen);
+		return msg.channel.send({
+			content: `This year, minions are delivering presents to all the Gielnor children! ${
+				style === 'good'
+					? 'But, there are some evil minions out there stealing presents, we need to make sure everyone gets a present!'
+					: "And, we're gonna stop them and get all those presents for ourself!"
```
themrrobert, line 77: This should be `ourselves`, since it's including the Antisanta and the Player

src/commands/christmas.ts
```diff
@@ -0,0 +1,143 @@
...
+						content: 'Take this outfit, it will give you special powers for stealing presents!',
+						head: 'antiSanta'
+					})
+				]
+			});
+		}
+		const delivered = msg.author.settings.get(UserSettings.PresentsDelivered);
+		const stolen = msg.author.settings.get(UserSettings.PresentsStolen);
+		return msg.channel.send({
+			content: `This year, minions are delivering presents to all the Gielnor children! ${
```
themrrobert, line 74: It's spelled, `Gielinor,` missing the `i` after the `l`

themrrobert (changes requested):

src/commands/christmas.ts
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	getStyle(id: string) {
+		return murMurHashChance(id, 44) ? 'good' : 'evil';
+	}
+
+	@requiresMinion
+	async run(msg: KlasaMessage) {
+		const style = this.getStyle(msg.author.id);
+		if (style === 'evil' && !msg.author.cl().has('Antisanta mask')) {
```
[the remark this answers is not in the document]
gc: will worry about bso later. ill be modifying it hugely for bso anyway.

PR #3310, opened 2021-12-19
Add Black swan to DCP CL to make it equippable

themrrobert:
### Description:

Black swan pet isn't equippable as it's not on any pet CL

### Changes:

Added 'Black swan' to the Discontinued pets CL to make it equippable.

### DIFF
--- a/src/lib/data/CollectionsExport.ts
+++ b/src/lib/data/CollectionsExport.ts
@@ -1965,7 +1965,8 @@ export const discontinuedCustomPetsCL = resolveItems([
 	'Cob',
 	'Gregoyle',
 	'Mini Pumpkinhead',
-	'Seer'
+	'Seer',
+	'Black swan'
 ]);
 
 export const kingGoldemarCL = resolveItems([


TastyPumPum:
Shouldn't this go with the little parasite and dark squirrel as it's just a variation basically?

themrrobert:
> Shouldn't this go with the little parasite and dark squirrel as it's just a variation basically?

Tough call, but it's a custom pet for sure, which means it has to go in either custom pets or disc custom pets, and since the base pet won't be obtainable after the event, seems appropriate to put it in DCP.

I'm not opposed to putting in just 'custom pets' though

TastyPumPum:
> > Shouldn't this go with the little parasite and dark squirrel as it's just a variation basically?
> 
> Tough call, but it's a custom pet for sure, which means it has to go in either custom pets or disc custom pets, and since the base pet won't be obtainable after the event, seems appropriate to put it in DCP.
> 
> I'm not opposed to putting in just 'custom pets' though

I was just going off the basis, its just a metamorph basically. The dye is on the relevant CL so I feel the pet doesn't need to be on any CL (like the dark squirrel).

If it does need to be on a cl, I agree discontinued custom pets is the place for it then.

Issue #2743, opened 2021-09-01
Add a way to open all implings in your bank at once


mylife212:
## 🚀 Feature Proposal

It can get tedious typing the individual command for multiple implings. A way to open all at once would be easier for the user. Not sure how it would be implemented tho

eggaroonie:
throwback to lily prs
https://github.com/oldschoolgg/oldschoolbot/pull/2235

Issue #2185, opened 2021-06-24
+open all 


Kaaakaa:
Id suggest a +open all command, maybe some extra verifikation required

PR #3317, opened 2021-12-20
Add open all, any and until master

LuxrayElite:
### Description:

Original PR: #2235 
Author @Lilylicious 

Repull was with permission from Lily
Nerfed --all and updated conflicts
Further nerfing could easily be applied.

Partially solves the issues: #2743 #2185

### Changes:

Open --all, opening an owned number of a single openable (requires name variable)

Open --any, opens an owned number of first openable in the +open list (doesn't require name)

Open --master, open until a master clue is found. (requires name variable)

### Other checks:
image
image


Lilylicious (approved):
Looks good to me.

LuxrayElite:
Consolidated duplicate code into a single function
Fixed issues related to merging into a function

orsou (approved):
This would save us so much typing!

LuxrayElite:
Changes since draft:
Experimented with limit restrictions; chose not to impose based on other commands response time
Updated to match changes implemented by loot tracking commit
Played around with code structure so only 1 additional function is now implemented. The rest is added onto the default functions

Its been requested to implement true openAll... (rather than limited to an item which I nerfed it too)
OpenAll implings
Open implings until clue.

Would also make a BSO PR for this to remove --master restrictions along with an update to =create for being able to create

LuxrayElite:
Closed due to rewrite by Magna

PR #3318, opened 2021-12-21
Adds Snake weed and Nail beast nails to Treebeard

mylife212:
### Description:
Added the ingredients for Sanfew serums to Treebeard drop table, in quantities of 10-50 each. This can be adjusted, but should be noted that Sanfews are only used for Phosani's nightmare, so these quantities aren't game breaking or anything. 

Also should be noted - getting Sanfew ingredients yourself in BSO is ungodly slow, adding them to Treebeard table fits the theme of it being a herblore boss. 

### Changes:

- Added Nail beast nails to secondariesTable, qty of 10-50
- Added Snake weed to secondariesTable, qty of 10-50

### Other checks:
Was polled in BSO-Vote, and had a very good turnout, showing the community wants this. 
image


mylife212:
HamSalute

PR #3335, opened 2021-12-24
BSO Update

gc:
### Other checks:

### DIFF
--- a/src/commands/Minion/use.ts
+++ b/src/commands/Minion/use.ts
@@ -3,6 +3,7 @@ import { Bank } from 'oldschooljs';
 
 import { bossActiveIsActiveOrSoonActive, bossEvents, startBossEvent } from '../../lib/bossEvents';
 import { BitField } from '../../lib/constants';
+import { dyedItems } from '../../lib/dyedItems';
 import { UserSettings } from '../../lib/settings/types/UserSettings';
 import { BotCommand } from '../../lib/structures/BotCommand';
 import getOSItem from '../../lib/util/getOSItem';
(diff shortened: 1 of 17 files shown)


src/commands/Minion/use.ts
```diff
@@ -171,6 +172,30 @@ export default class extends BotCommand {
...
+		if (items.length === 2) {
+			// e.g. [Blood Dye, Dwarven warhammer (blood)]
+			const baseItem = dyedItems.find(i => i.dyedVersions.some(o => items.includes(o.item)));
+			const dyeToApply = baseItem?.dyedVersions.find(i => items.includes(i.dye));
+			const dyedVariantTheyHave = baseItem?.dyedVersions.find(i => items.includes(i.item));
+			if (baseItem && dyeToApply && dyedVariantTheyHave) {
+				await msg.confirm(
+					`Are you sure you want to use a ${dyeToApply.dye.name} on your ${dyedVariantTheyHave.item.name}?`
+				);
+				if (!msg.author.owns(dyedVariantTheyHave.item.id)) {
```
themrrobert, line 186: Need to check user's bank for both the target item AND the dye itself.

Attempting to do: `-use blood dye, dwarven warnana` without a blood dye yields an unhelpful error, `An unexpected error occurred :RSSad:` and an exception is thrown.

src/commands/bso/tames.ts
```diff
@@ -87,6 +87,12 @@ const feedableItems: FeedableItem[] = [
 		tameSpeciesCanBeFedThis: [TameType.Gatherer],
 		announcementString:
 			'Your tame can now collect items 10% faster thanks to the Ring of endurance helping them run for longer!'
+	},
+	{
+		item: getOSItem('Dwarven warhammer'),
+		description: '20% faster PvM',
```
themrrobert, line 93: The Tame kills so slow already that I feel like 20% isn't really worth it for a dwarven warhammer.
Example: It takes 73 minutes for the tame to kill 2x QBD. 

I personally think DWWH should be at least 40% for the tame, also.

(the cost for DWWH is currently ~25b )

src/lib/data/CollectionsExport.ts
```diff
@@ -2185,7 +2185,7 @@ export const allPetIDs = [
 	...discontinuedCustomPetsCL,
 	...chambersOfXericMetamorphPets,
 	...growablePets.map(petSeries => petSeries.stages).flat(1),
-	...resolveItems(['Little parasite', 'Dark squirrel'])
+	...resolveItems(['Little parasite', 'Dark squirrel', 'Dark swan'])
```
themrrobert: Should be  `Black swan`

src/lib/skilling/skills/hunter/creatures/bso.ts
```diff
@@ -34,6 +34,25 @@ const customBSOCreatures: Creature[] = [
...
+		table: new LootTable().every('Eastern ferret'),
+		qpRequired: 600,
+		huntTechnique: HunterTechniqueEnum.Tracking,
+		catchTime: 350,
+		slope: 0,
+		intercept: 99,
+		bait: qty => {
+			let req = new Bank();
+			const kibbleRequired = Math.ceil(qty / 2);
+			req.add('Extraordinary kibble', kibbleRequired);
```
themrrobert, line 53: There's no current use for 'Delicious kibble' yet which is the middle tier, perhaps this should use 'Delicious kibble', and save 'Extraordinary kibble' for a higher level thing?

src/tasks/minions/HunterActivity/hunterActivity.ts
```diff
@@ -143,6 +143,19 @@ export default class extends Task {
 				loot.add(itemID('Baby chinchompa'));
 			}
 		}
+		if (creature.name === 'Eastern ferret') {
+			const cl = user.cl();
+			const amountOfZippysGotten = cl.amount('Zippy');
```
themrrobert, line 148: I really don't like this CL mechanic for non-limited items. 
It makes competing for mostdrops impossible for newcomers, and more importantly, significantly demotivates players from continuing the content after 1 or two drops. 

If you think the rate is too common, then make it rarer, but I disagree with doing this for non-limited items.

themrrobert (changes requested):

Issue #2972, opened 2021-10-05
Add a way to revert zulrah uniques into 20k scales like in-game


mylife212:
## 🚀 Feature Proposal

What title says. The 3 uniques being Tanzanite fang, serpentine visage and magic fang.

mylife212:
merged, ty Lux!

PR #3349, opened 2021-12-27
Add zulrah uniques to revertables

LuxrayElite:
### Description:

Allows zulrah uniques to be exchanged for scales

raised by #2972 

### Changes:

Added:
Tanzanite fang, Magic fang, Serp visage, BP (uncharged), Serp helm (uncharged) for 20k scales

### Other checks:


src/lib/data/createables.ts
```diff
@@ -366,6 +366,56 @@ const hunterClothing: Createable[] = [
...
+		},
+		outputItems: {
+			[itemID("Zulrah's scales")]: 20_000
+		},
+		noCl: true
+	},
+	{
+		name: 'Revert serpentine helm (uncharged)',
+		inputItems: {
+			[itemID('serpentine helm (uncharged)')]: 1
```
gc, line 412: these need to be cased properly (also elsewhere)

gc (changes requested):

src/lib/data/createables.ts
LuxrayElite: Correctly capitalized all the zulrah reverables now.

PR #2447, opened 2021-07-25
2h fix

Fexbot:
### Description:

I enabled automatically equipping and unequipping two-handed equipment.

### Changes:

- When equipping a 2h weapon, any items in the weapon and shield slots will be unequipped automatically.
- When equipping a weapon or shield slot item, any 2h weapon equipped will be unequipped automatically.

### Other checks:

-   [x] I have tested all my changes thoroughly.

PR #3367, opened 2021-12-30
enable 2h equipping automatically

LuxrayElite:
### Description:
Original PR: #2447
Author @Fexbot 

I believed 6 months was long enough to recode this and fix any errors (mainly duping items)

The review was asked for in August and not seen movement 

Allows the equipping of 2H items without having to unequip both weapon and shield

Still can only equip a weapon or a shield if wielding a 2h but a lot less tedious until we equip multiple items at once

### Changes:
Changed the order of checking if an item can be equipped (Stackable, Stats and Wildy check) prior code had these after allowing item duping

Added associated logging messages already in the code to the new section

Check whether the item to be equipped is 2h, check if there's a weapon and shield then return them to the bank

Check whether the item to be equipped is a shield, weapon or 2h and if currently wielding 2h, return equipped 2h and equip a new item

### Other checks:
Manually tested with godswords,inq mace,ely, arma cbow and tbow (high-value items most likely to be raged about if an error in code)

example: 
image


themrrobert:
Yes, you are right this feature was removed because of a dupe.
I don't remember exactly how it was performed, do you?

We need to be absolutely sure here, if you remember how the dupe was done, please let me know (Cyr) via pm on discord and we can look into this thoroughly.

LuxrayElite:
Updated to meet new structure requirements. 
Had a convo with Cyr regarding duping. (All scenarios were already fixed in the first commit)
Personally haven't found any duping errors since the first commit but would like additional testing by others to look for errors

PR #3372, opened 2022-01-01
Fix CoX Gear score / boost conflict

themrrobert:
### Description:

As we discussed, this is necessary because you can't reach 100% gear score while having most of these items equipped.

The way it is now, it's impossible to reach 100% gear score AND have all boots. And if you bring down the BiS gear to match the boost gear would make it too easy to hit 100% gear score, so I think this is the best solution, though I'm open minded.

### Changes:

- Set mustBeEquipped to `false`
   (Only bso drygores should be `true`, but since they're BIS anyway, it can be removed)
- Changed tentacle charges used to 200 in CoX because you only really use whip for Olm, and very little for vasa/vanguards (but you shouldn't use it for vasa anyway)
- Now the sang/tentacle boosts work from bank, but only under 2 conditions:
- 1) You don't have a higher boost available
- 2) You have enough charges
- If you have tent/sang equipped, it won't let you join the raid if you don't have enough charges
- If they're not equipped, and you don't have enough charges, you just don't get the boost.
- Changed 'check can degrade' for Sang to match the actual charges degraded, 150
- It's the perfect solution :D

### Other checks:

### DIFF
--- a/src/commands/Minion/raid.ts
+++ b/src/commands/Minion/raid.ts
@@ -14,7 +14,6 @@ import {
 	hasMinRaidsRequirements,
 	minimumCoxSuppliesNeeded
 } from '../../lib/data/cox';
-import { TENTACLE_CHARGES_PER_RAID } from '../../lib/data/tob';
 import { degradeItem } from '../../lib/degradeableItems';
 import { ClientSettings } from '../../lib/settings/types/ClientSettings';
 import { UserSettings } from '../../lib/settings/types/UserSettings';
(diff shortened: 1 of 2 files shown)


src/commands/Minion/raid.ts
```diff
@@ -213,39 +211,28 @@ export default class extends BotCommand {
...
 		const isSolo = users.length === 1;
 
 		const totalCost = new Bank();
 
-		await Promise.all(
-			users.map(async u => {
+		await Promise.all([
+			degradeables.map(async d => {
+				await degradeItem(d);
+			}),
```
gc, line 224: `degradeables.map(degradeItem)` ?

gc (changes requested):

PR #3402, opened 2022-01-08
setup sotw

gc:
### Other checks:

### DIFF
--- a/src/commands/bso/sotw.ts
+++ b/src/commands/bso/sotw.ts
@@ -30,12 +30,15 @@ export default class extends BotCommand {
 			);
 		}
 
+		const startSQL = `to_timestamp(${start.getTime()}/1000) AT TIME ZONE 'UTC'`;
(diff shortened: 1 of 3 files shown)


src/commands/bso/sotw.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const sotwConfig = {
+	start: new Date('2022/1/10 15:00'),
+	finish: new Date('2022/1/17 15:00'),
+	skill: SkillsEnum.Runecraft,
+	notes: ['1st age items do not contribute', 'Master capes do not work']
+} as const;
+const { start, finish, skill } = sotwConfig;
+
+export const sotwIsActive = start.getTime() <= Date.now();
```
themrrobert, line 19: I would probably do
```ts
export const sotwIsActive = start.getTime() <= Date.now() <= finish.getTime();
```

But this is fine

PR #3421, opened 2022-01-15
Added support for normal and corrupted gauntlet in kill simulator

FannyPackFan:
### Testing

- All aliases added to the command were verified


LuxrayElite:
This is in part covered by an OSJS PR. (I think that only focuses on CG not gauntlet tho)
https://github.com/oldschoolgg/oldschooljs/pull/274

GJ on your first PR :)  (don't pre-emptively close based on my feedback)

FannyPackFan:
Ah didn't see the other variation in OSJS. Based on the general structure of the other monsters for the kill simulator, I think that one is the more "correct" solution. I just went with the simplest route to reuse the existing loot table.

PR #3424, opened 2022-01-15
Magic Overhaul

Superwhammybar:
This aims to consolidate magic into just the cast command and expand the available spells. I was unable to merge alch in as it has bespoke logic for favourites and I was unable to create a way to redirect a command over there from cast, using subcommands, as it doesn't use a string for item names but rather a list of Item objects. If anyone can figure it out, then please do let me know.

These changes add in reviving ensouled heads and offering ashes/bones via the Arceuus spellbook. 
Ensouled heads do not provide combat XP, but it assumes ~300 heads per hour based on EHB and Discord feedback.

### Description:

Adds in every spell in the game to be casted, with some that have no current purpose commented out (like Home Teleports or Telekinetic Grab). 

Since all spells are added into the cast command, it eliminated the need for the enchant command. This has been removed.

### Changes:

Removed Enchant command and references to Enchanting within DB schema and tasks.

Created files for each spellbook with all spells coded, but some commented out that have no apparent use. This replaces the old castables.ts and enchantables.ts files.

Created a new index file for Magic which defines the Castable spell interface. Level requirements for spells and XP per cast are now within one object each rather than several properties as before (e.g. xp, craftXp, constructionXp properties for XP). Input and outputs are now Banks. Each spell is given a MagicType for potential future use.

Casting Activity now uses spell name rather than ID as the ID was not scalable uniquely (previously used an itemID to represent each spell, but this falls apart when you have multiple instances of a spell like with enchanting and stringing jewelry).

Cast command returns a structured table which lists all currently available spells to cast to reduce ambiguity. This is ~300kB file so I'm not sure if this needs changing or not if it's too big.

### Other checks:

I tested the XP rates for spells between my PR and the master branch and found XP rates for existing spells is the same for the majority since they're basically copied over. I did find that the standard spellbook teleports (like Varrock teleport) becomes better XP rate on my branch as I have the tick rate as 4 (wiki rate) rather than 5. Looking at information on XP rates for spamming teleports, mainly found via YouTube videos, the new rate is closer to expected hourly rates.


LuxrayElite:
Currently having a test with this will edit as needed:

Can you change bolts to be "enchant <type> bolts" instead of "Enchant 10x jade bolts"   similar to how it previously was

LuxrayElite:
Bake pies currently no use except BSO (for now)
You left Arceuus home teleport in

Superwhammybar:
> Bake pies currently no use except BSO (for now) 

I will leave the spells in for when they are obtainable. When I commented out other spells it was mainly because I can't see a use-case at all for them, like home teleports and other non-combat or limited use spells.

PR #3425, opened 2022-01-15
Rada's Blessing Fishing

TastyPumPum:
Adding the function of Rada's blessing to give a chance of doubling a catch. Chance of it depending on the tier of the rada's blessing. 


src/tasks/minions/fishingActivity.ts
```diff
@@ -11,11 +11,26 @@ import { anglerBoostPercent, rand, roll } from '../../lib/util';
...
+		return { blessingEquipped: true, blessingChance: 8 };
+	} else if (skillingGear.hasEquipped(itemID("Rada's Blessing 3"))) {
+		return { blessingEquipped: true, blessingChance: 6 };
+	} else if (skillingGear.hasEquipped(itemID("Rada's Blessing 2"))) {
+		return { blessingEquipped: true, blessingChance: 4 };
+	} else if (skillingGear.hasEquipped(itemID("Rada's Blessing 1"))) {
+		return { blessingEquipped: true, blessingChance: 2 };
+	}
+	return { blessingEquipped: false, blessingChance: 0 };
+}
```
gc, line 26: do something like
```ts
const blessingBoosts = [["Rada's blessing 4", 8]];

for (const [itemName, boostPercent] of blessingBoosts) {
  if (hasitemequippedanywhere(itemname)) {
    return {whatever}
  }
}
```
instead of chaining if/else like this

gc (changes requested):

src/tasks/minions/fishingActivity.ts
TastyPumPum: Updated thanks :)

PR #3437, opened 2022-01-17
Add more Aliases to various agility courses

mylife212:
### Description:

Updated a few courses with new Aliases, which include possible typos and the surrounding area as an alias, since players refer to the courses this way sometimes. 

### Changes:

- Adds more Aliases to: Ape atoll, Pollniveach, Seers village, Fremennik and Falador courses.  

### Other checks:


mylife212:
cba with fixing lint issues

PR #3481, opened 2022-01-28
Aliases for pickpocket npcs

TastyPumPum:
Added aliases for various pickpockets, unable to add any to Pollnivnian Bandit until the oldschooljs PR is merged.  As it is using the wrong ID currently. 

https://github.com/oldschoolgg/oldschooljs/pull/275

### DIFF
--- a/src/commands/Minion/pickpocket.ts
+++ b/src/commands/Minion/pickpocket.ts
@@ -65,7 +65,10 @@ export default class extends BotCommand {
 			quantity = null;
 		}
 
-		const pickpocketable = Pickpocketables.find(npc => stringMatches(npc.name, name));
+		await msg.author.settings.sync(true);
+		const pickpocketable = Pickpocketables.find(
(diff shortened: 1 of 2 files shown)


src/commands/Minion/pickpocket.ts
```diff
@@ -65,7 +65,10 @@ export default class extends BotCommand {
 			quantity = null;
 		}
 
-		const pickpocketable = Pickpocketables.find(npc => stringMatches(npc.name, name));
+		await msg.author.settings.sync(true);
```
gc, line 68: why did you add this?

gc (changes requested):

src/commands/Minion/pickpocket.ts
TastyPumPum: Apologies, copied it over accidentally and everything worked fine in testing so didn't pick up on it

PR #3489, opened 2022-01-30
Buylimits

gc:
### Other checks:


src/lib/util/buyLimit.ts
```diff
@@ -0,0 +1,68 @@
...
+	buyLimitBank,
+	increaseFactor,
+	itemBeingBought,
+	quantityBeingBought,
+	baseCost,
+	absoluteLimit
+}: {
+	buyLimitBank: Bank;
+	/**
+	 * The price multiplier increases by 1 for each multiple of this amount they have bought. (1x, 2x, 3x, etc)
```
themrrobert, line 14: This comment i made is not correct with the current algorithm,, it  should be something like,
'The price multiplier doubles for every multiple of this amount they have bought'

src/lib/data/buyables/tokkulBuyables.ts
```diff
@@ -38,48 +41,55 @@ const TokkulShopItems: TokkulShopItem[] = [
...
+		aliases: ['silver'],
+		buyLimitFactor: 500
 	},
 	{
 		name: 'Coal',
 		inputItem: itemID('Coal'),
 		tokkulReturn: 4,
-		diaryTokkulReturn: 9
+		diaryTokkulReturn: 9,
+		buyLimitFactor: 800
```
themrrobert, line 60: I think the limit may be a bit low for coal; coal is not readily provided in the correct ratio to feed addy [let alone rune] smithing, so it would be nice if TKS was a viable method to get some extra coal.

src/lib/util/buyLimit.ts
```diff
@@ -0,0 +1,68 @@
...
+	let finalCost = 0;
+
+	if (quantityBeingBought <= 0) {
+		return { amountToBuy, finalCost };
+	}
+
+	const maxCanBuy = absoluteLimit - amountBought;
+	const allowedBuyAmount = quantityBeingBought > maxCanBuy ? maxCanBuy : quantityBeingBought;
+	// CHANGE THIS TO MATCH currentMultiplier
+	// let currentMultiplier = Math.floor(amountBought / increaseFactor) + 1;
```
themrrobert, line 38: this code line can be removed, but the comment above it might be useful to keep for  any possible algorithm changes in the future

tests/buyLimit.test.ts
```diff
@@ -0,0 +1,154 @@
...
+				itemBeingBought: getOSItem('Air rune'),
+				quantityBeingBought: 2500,
+				absoluteLimit: 10_000,
+				baseCost: 25
+			})
+		).toEqual({ amountToBuy: 1778, finalCost: 17_779_200 });
+		/*
+
+
+		// Test absolute limit and multiple increase tiers:
```
themrrobert, line 68: These are old tests for 1x,2x,3x,4x scaling, so they can also be removed

themrrobert (changes requested):

PR #3494, opened 2022-01-31
Add a collection export of gwd

### LINKED ISSUE #3490: cl gwd (osb)
Used to be able to do +cl gwd and see all the gwd bosses, doesn't seem to work anymore, was a handy command 

Edit - works on bso just not osb

TastyPumPum:
Closes #3490

Allows you to view all of the godwars cl at once. 


gc (changes requested):
This is an unofficial/made-up CL group, it needs to go in the 'custom' CL, and have "counts: false"

gc:
If any current CLs dont match ingame exactly, its 'not intended' and basically a bug, not by decision they should be there.

PR #3495, opened 2022-01-31
Bowfa cox

TastyPumPum:
As bowfa boosts cox more than dhcb, makes sense to check for that too for cm raids, not just dhcb or tbow. 


Redquaker:
Should this not check for full crystal armor in bank/equipped?

TastyPumPum:
Not until full crystal armour is a thing :eyes:

osDreamscape:
if/when crystal armour creation merges bowfa can be adjusted to require the armor for cms/boost

PR #3497, opened 2022-01-31
Changed buy confirmation on the sellto command to a button rather than a message

ianfights:
### Description:

Changed the buy confirmation on the sellto command to a button rather than a message

I made this to address 2 issues.
1. Message content becoming a privileged intent on April 30th
2. Consistency as a button is already used for the seller confirmation

I agree it is more difficult to catch in some busy channels but based on what I've seen they hope to avoid using the privileged intents if possible 


POKEBLOX6:
I made this to address 2 issues.
1. Message content becoming a privileged intent on April 30th
2. Consistency as a button is already used for the seller confirmation

I agree it is more difficult to catch in some busy channels but based on what I've seen they hope to avoid using the privileged intents if possible 

> Just want to say in some busy bso channels for example it can be hard to press buttons like this. Don't see this needed imo

PR #3498, opened 2022-02-01
BSO-Sepulchre patch

### LINKED ISSUE #3496: BSO Sepulchre Trips Not Returning
## 🐛 Bug Report
I send a bso sepulchre trip and while it is going I can check it with =m, but it doesn't return. My minion isn't stuck, it is free after the trip "ends", it just doesn't show up. I believe I receive xp when the trip ends, but will have to test to be sure.
Screenshot_20220131-163121_Discord

LuxrayElite:
### Description:

Added a check so it only tries to roll pages for the base game floors 

Closes #3496

### Changes:

wrapped the rolling of pages in an if

### Other checks:
Tested all pages still obtainable


LuxrayElite:
closed in favour of main branch

PR #3511, opened 2022-02-03
Add frozen key

gc:
### Other checks:


themrrobert (approved):
Just be careful with the merge to BSO since there already exist a 'frozen key' and key pieces (pieces have different names though)

PR #3523, opened 2022-02-06
Nex migration

gc:
### Other checks:

### DIFF
--- a/src/commands/rp.ts
+++ b/src/commands/rp.ts
@@ -1007,7 +1007,7 @@ WHERE "collectionLogBank"::jsonb ?| array['601', '605', '4272', '758', '759', '4
 						if (cl.has(normalItem)) {
 							const quantityToAdd = cl.amount(normalItem) - cl.amount(brokenItem);
 							if (quantityToAdd > 0) {
-								itemsToAddToTheirCL.add(brokenItem);
+								itemsToAddToTheirCL.add(brokenItem, quantityToAdd);
 							}
 						}
 					}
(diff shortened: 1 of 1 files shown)


src/commands/rp.ts
```diff
@@ -972,6 +972,49 @@ WHERE bank->>'${item.id}' IS NOT NULL;`);
...
+FROM users
+WHERE "collectionLogBank"::jsonb ?| array['601', '605', '4272', '758', '759', '432', '709', '2404', '2838', '4273', '788', '983', '2409', '9654', '2423'];`);
+				let IDs = res.map(i => i.id);
+
+				for (const id of IDs) {
+					const user = await client.fetchUser(id).catch(noOp);
+					if (!user) continue;
+					const itemsToAddToTheirCL = new Bank();
+					const cl = user.cl();
+					if (!cl.has('Virtus crystal') && (cl.has('Virtus wand') || cl.has('Virtus book'))) {
```
themrrobert, line 1003: If you're giving people full credit for both books + wands, why not give them credit if they happen to roll a fresh Virtus crystal since the update? I would probably remove the `!cl.has('Virtus crystal)` but you might have a good reason it's there ?  
Maybe just subtract the existing number of crystals from the total of old ones. (To account for potentially created orbs/books after the update) (and checking it's > 0)

src/commands/rp.ts
```diff
@@ -972,6 +972,49 @@ WHERE bank->>'${item.id}' IS NOT NULL;`);
...
+						itemsToAddToTheirCL.add('Virtus crystal', cl.amount('Virtus wand') + cl.amount('Virtus book'));
+					}
+					for (const [brokenItem, normalItem] of nexItems) {
+						if (cl.has(normalItem)) {
+							const quantityToAdd = cl.amount(normalItem) - cl.amount(brokenItem);
+							if (quantityToAdd > 0) {
+								itemsToAddToTheirCL.add(brokenItem);
```
themrrobert, line 1010: Add the quantityToAdd qty?
`itemsToAddToTheirCL.add(brokenItem, quantityToAdd);` ?

themrrobert (changes requested):

Issue #3545, opened 2022-02-11
Dark Altar


6enny6:
A lot of people are unaware of dark altar runecrafting because it has a seperate command from +rc

if you try +rc blood/soul could the bot return a message saying to use +darkaltar ?

PR #3548, opened 2022-02-12
RC Dark Altar Message Redirect

Superwhammybar:
Closes https://github.com/oldschoolgg/oldschoolbot/issues/3545

### Description:

Send a message back to the user to direct them to the darkaltar command if they try to RC blood or soul using the rc command.

### Changes:

Added message that returns if "blood" or "soul" is in the rune name being runecrafted. Directs them to the darkaltar command instead.
Darkaltar skill requirements apply to both blood and soul so removed reference to just "blood" runes in the "you don't have the skills" message.

### Other checks:

No runes have "blood" or "soul" in them so doing .includes() should be fine and doesn't break anything.


gc:
I think it would be better to remove the darkaltar command and just integrate it into the RC command, you can PR this if you wish, as im not sure when Ill get to doing it.

Issue #3554, opened 2022-02-13
Obtaining frozen key after losing it


SkumRobin:
We can't re-obtain a frozen key after losing it. Would like it to check for bank instead of cl whilst dropping it. Like the Cox or Tob pets.

PR #3560, opened 2022-02-14
Allow extra frozen key key pieces if dropped/sac'd/lost

themrrobert:
### Description:

Right now a lot of people are asking for extra pieces. There are several reasons:
OSB: People was those Sac CL slots
BSO: Key pieces are required for item contracts.

And then of course, Nex requires that you OWN the key, not just have it on your CL, so it has to be reclaimable.

### Changes:

Gives you a new key piece as long as you don't have either: 1. that key piece already or 2. a completed frozen key.

### Other checks:

closes https://github.com/oldschoolgg/oldschoolbot/issues/3554


6enny6:
closes #3554

PR #3563, opened 2022-02-14
Add rc alias

TastyPumPum:
Adds an alias of "rc" so people can do "+m lvl rc" etc


6enny6:
can the alias 'worst skill' also be added to this please

PR #3578, opened 2022-02-17
Mahoji- Laps

LuxrayElite:
### Description:

Converted laps to a / command
Made alching true by default (no XP waste) if you have the levels, runes, and favalch set up ofc
Made amount of laps default to max if an erroneous (larger than max) quantity is input instead of throwing a message
Ensured Ape atoll doesn't benefit from alching, also added penguin course to blocked alching list (may be wrong but you cant hold a staff)

### Changes:

Added laps.ts to mahoji commands
Added laps.ts to depreciated
Deleted laps.ts from minion

### Other checks:

Again further testing is appreciated


LuxrayElite:
now updated to include PR #3550

gc:
>Made alching true by default (no XP waste) if you have the levels, runes, and favalch set up ofc

Please avoid making unrelated changes like this when rewriting slash commands (or just in general). Even if its a good idea, it has to be discussed/talked about, and done separately from this rewrite, the goal is for the slash commands to be bug-free, and identical mostly to the normal commands.

LuxrayElite:
Rebased and squashed

Removed anything outside of the scope of the base command:
Bryo staff working for agility
PR #3550 
Alching by default

PR #3583, opened 2022-02-17
Mahoji-mage arena

LuxrayElite:
### Description:

Compressed Magearena and Magearena2 to a single slash command

Can be re-pulled as two commands if needed
### Changes:

added magearena.ts to mahoji
depreciated magearena.ts and magearena2.ts from minion

### Other checks:


LuxrayElite:
Converted by magna

PR #3590, opened 2022-02-18
Mahoji- Trek

LuxrayElite:
### Description:

Combine Trek and Trek shop to a single command
There was an error if you allowed quantity to be undefined when buying so it is now forced instead of max
Currently allowing aliases for name input due to autocomplete only working for the first variable (Perhaps command structure order change would be favourable)

### Changes:
Add trek.ts to mahoji
Depreciate trek.ts and trekshop.ts
Changed activity to handle mahoji commands for repeat trip
### Other checks:
As with all mahoji PR's additional testing is appreciated


LuxrayElite:
Amended so that item list is the first variable due to higher number of options compared to the difficulty
Autocomplete has to be the first item of the command

PR #3596, opened 2022-02-18
Fix max slash + crush offence stats (BSO)

### LINKED ISSUE #3585: Green stats in bso
Currently bis slash offensive gear is higher than green text. Green text is 288 and max is 295. I believe the discrepancy comes from ignis ring (i) which has +7 slash. I havent checked other melee green stats.

TastyPumPum:
osbot
osbot-1

Closes #3585 

The stab can be updated too:
offhand drygore/chaotic rapier gives 358 stab, 
363 is obtainable with offhand dragon claw, 
361 is not obtainable


themrrobert:
I like the idea of nerfing offhand dragon claw so that dual drygores are BIS and make 358 stab the max stab, (or whatever bis stab is)

PR #3604, opened 2022-02-19
Testing commands rewrite

gc:
### Other checks:


src/mahoji/commands/testpotato.ts
```diff
@@ -0,0 +1,342 @@
...
+			],
+			run: async ({
+				options,
+				userID
+			}: CommandRunOptions<{
+				max?: {};
+				patron?: { tier: string };
+				gear?: {};
+				reset?: {};
+				setminigamekc?: { minigame: string; quantity: number };
```
LuxrayElite, line 293: Quantity doesn't exist, the field is kc so it ends up passing an undefined value

PR #3625, opened 2022-02-25
Fix `/open` with Seed pack's.

themrrobert:
### Description:

Fixes Seed packs, because the osjs openable can't possibly work because there are different tiers based on the farming contract, and it just has to be handled in code.

**Note:** Currently it's so broken, you only seed 1 seed roll per open, so this is a medium priority.

### Changes:

- Exports the openSeedPack function
- Adds a lambda function to the 'Seed pack' openable to handle the opening of the seed packs.

### Other checks:

### DIFF
--- a/src/lib/openables.ts
+++ b/src/lib/openables.ts
@@ -255,9 +255,8 @@ const osjsOpenables: UnifiedOpenable[] = [
 			message?: string;
 		}> => {
 			const { plantTier } = args.user.settings.get(UserSettings.Minion.FarmingContract) ?? defaultFarmingContract;
-			const realQty = args.quantity;
 			const openLoot = new Bank();
-			for (let i = 0; i < realQty; i++) {
+			for (let i = 0; i < args.quantity; i++) {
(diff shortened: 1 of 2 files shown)


src/lib/openables.ts
```diff
@@ -246,7 +248,20 @@ const osjsOpenables: UnifiedOpenable[] = [
...
 		aliases: ['seed pack'],
-		output: Openables.SeedPack.table,
+		output: async (
+			args: OpenArgs
+		): Promise<{
+			bank: Bank;
+			message?: string;
+		}> => {
+			const { plantTier } = args.user.settings.get(UserSettings.Minion.FarmingContract) ?? defaultFarmingContract;
+			const realQty = args.quantity ?? 1;
```
gc, line 258: quantity cant be undefined, its always provided. its not a user-provided variable, its validated/checked.

gc (changes requested):

src/lib/openables.ts
themrrobert: ah thanks for updating that for me :)

PR #3633, opened 2022-02-27
Remove Tob items from mystery boxes

themrrobert:
### Description:

Stops Tob uniques from appearing in mystery boxes.

The recolored variants etc will still drop, this just prevents the CL from being obtainable by MBs.

### Changes:

Adds tob cl to `cantBeDropped` array.

### Other checks:


6enny6:
Could also remove all 2k tob capes from the economy, no one is even close to the kc requirement

PR #3640, opened 2022-02-28
Fix merging tames so it keeps consumed Shiny

### LINKED ISSUE #3639: Lost shiny tame when merging
Robson lost his shiny tame when merging two, correct me if im wrong but isnt shiny one of the attributes it ‘keeps’ when merging? 

if not, kinda annoying that if you get a shiny its locked to the stats it currently has

See photo below

2685A68B-A392-44BC-880F-9BC62F184E61

themrrobert:
### Description:

Fixed merging so that it also preserves Shiny of the consumed tame.

### Changes:

Added some code to check if either tame is shiny, and update the merged tame accordingly.

### Other checks:

Closes #3639

### DIFF
--- a/src/commands/bso/tames.ts
+++ b/src/commands/bso/tames.ts
@@ -886,7 +886,7 @@ export default class extends BotCommand {
 			return msg.channel.send("You can't merge two tames from two different species!");
 		}
 
-		const { mergingCost } = getTameSpecies(currentTame);
+		const { mergingCost, shinyVariant } = getTameSpecies(currentTame);
 
 		if (!msg.author.owns(mergingCost)) {
 			return msg.channel.send(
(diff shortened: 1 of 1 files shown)


imTURBObruh:
@themrrobert why not just follow the merge formula and not do what Robson did :p

> `=t select x` to select the tame you wish to **_keep_**, then `=t merge y` to **_merge y into x_**, ~~`deleting y`~~ in the process.

so make sure shiny tame is always one selected -- `x`, and then merge it with `y` and then your shiny still lives..

there is no need for a extra code because people fail to read the formula.

themrrobert:
I considered this, however this prevents you from choosing the desired tame ID number.

You still have to select the right one to keep any color other than shiny, but shiny is a special value that should be kept when merging a tame imo

imTURBObruh:
> I considered this, however this prevents you from choosing the desired tame ID number.
> 
> You still have to select the right one to keep any color other than shiny, but shiny is a special value that should be kept when merging a tame imo

what do you mean prevents you from choosing desired ID number? 

you select your shiny. because it's shiny and merge it with the extra adult dragon of your choice.

SpaceDevz:
> @themrrobert why not just follow the merge formula and not do what Robson did :p
> 
> > `=t select x` to select the tame you wish to **_keep_**, then `=t merge y` to **_merge y into x_**, ~~`deleting y`~~ in the process.
> 
> so make sure shiny tame is always one selected -- `x`, and then merge it with `y` and then your shiny still lives..
> 
> there is no need for a extra code because people fail to read the formula.

Yes you could do this, but having this other check would prevent a-lot of frustration.

PR #3649, opened 2022-03-03
Barb Assault Buy Fix

TastyPumPum:
Enables you to buy items again using your honour points, if no quantity is provided quantity is set as 1


TastyPumPum:
Reopened as I think it should be merged and can always be moved to a buy command at a later point.

PR #3675, opened 2022-03-13
Move several minigames to slash commands

gc:
### Other checks:


LuxrayElite:
This PR doesn't include the relevant changes to the associated activities to make it repeatable on trip end

LuxrayElite:
Tears of guthix, rogues den and plunder don't have minion is busy checks

PR #3677, opened 2022-03-14
Add Nex bank filter (BSO)

### LINKED ISSUE #3672: Add Fixed Nex Items to --nex flag
## 🚀 Feature Proposal

Currently =b --nex only shows items on the nex collection log. I propose this be expanded to include the fixed items as well because they are just as relevant as the broken ones. This change would be similar to how --corp shows assembled spirit shields and --gwd shows assembled godswords

TastyPumPum:
Adds a filterable of --nex to allow fixed items to show in the bank. Similar how corp shows the assembled spirit shields etc.

Closes #3672

### DIFF
--- a/src/lib/data/filterables.ts
+++ b/src/lib/data/filterables.ts
@@ -31,7 +31,11 @@ import {
 	cluesMasterRareCL,
 	cluesMediumCL,
 	cluesSharedCL,
+	nexCL,
+	pernixOutfit,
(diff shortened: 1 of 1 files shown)


gc:
shouldnt this include the broken items?

TastyPumPum:
> shouldnt this include the broken items?

It includes everything on the nex cl,  so includes broken items that way

PR #3689, opened 2022-03-20
Fix spawnlamp so it requires T3

themrrobert:
### Description:

Fixes =spawnlamp so it requires Tier 3 as advertised.

### Changes:

- Does what it says on the tin.
- Removes some logic that was causing all Tiers to be allowed to spawnlamp.

### Other checks:


gc:
Don't think this is needed anymore.

PR #3705, opened 2022-03-25
Changed combat level emoji

6enny6:
Changed the emoji used to show combat level from the one sword thing to the two crossed swords


themrrobert:
Where is this emote from? It looks like a broken image to me

PR #3709, opened 2022-03-27
[Nex] Fix exponential decay function

themrrobert:
### Description:

I provided some code that was an incorrect conversion of the math. This corrects that.
The red line is the function as implemented, and the blue line is the function in the PR. As you can see they overlap nearly perfectly.

### Changes:

- Fixes the equation in exponential decay function

### Other checks:

image

Also tested the code itself to make sure the numbers were the same. Less than 1% difference until 60% gear score, and then max of 2% difference at 20% (22.52 vs 22.99) as you can see from the image, they're extremely close.


src/lib/util.ts
```diff
@@ -720,7 +720,6 @@ export function removeFromArr<T>(arr: T[], item: T) {
  * @param percent The percent to scale
  * @returns percent
  */
-export function exponentialPercentScale(percent: number, decay = 0.0038) {
-	const decayedPercent = Math.pow(100 * Math.E, -decay * (100 - percent));
-	return decayedPercent * 100;
+export function exponentialPercentScale(percent: number, decay = 0.021) {
+	return 100 * Math.pow(Math.E, -decay * (100 - percent));
```
gc, line 724: this gonna break/change the current scaling though? :S

themrrobert: No it's a drop in replacement.

The numbers are off by a very small margin, but nothing noticeable. (1% average)

I did a lot of testing to make sure it wouldn't break anything, you put `90` into each function, one gives 80.8 one gives 81.05

PR #3734, opened 2022-03-31
Nex Loot Bug Fixes

Superwhammybar:
### Description:

The handleNexKills function currently does not deal with player deaths for all loot rolls. This means that a player that dies during a kill still has a pet chance. It is also possible for a player to be chosen to receive a unique who had died during the kill. Meaning a unique should have been given out but wasn't because the player was dead.
The kill also gave all members an elite if the 1/20 chance was hit, this changes it so each survivor rolls to receive one.

### Changes:

handleNexKills function now creates a list of surviving team members for each kill. 
This list is then used to:
* Choose a random unique recipient if the 1/43 chance is hit
* Iterate over to distribute loot, including a 1/20 chance for an elite clue
* Choose a random team member to receive the pet, if the 1/500 chance is hit

### Other checks:

Pum helped me test in my private server
* Same non-unique loot is still being allocated to survivors
* Loot is only given to living players (uniques, pets and clues included) - To test this, I gave one person a death for each kill to ensure and no loot was given
* Uniques seemed evenly distributed

### DIFF
--- a/src/lib/simulation/misc.ts
+++ b/src/lib/simulation/misc.ts
@@ -211,5 +211,4 @@ const NexNonUniqueBaseTable = new LootTable()
 export const NexNonUniqueTable = new LootTable()
 	.every(NexNonUniqueBaseTable, 2)
 	.oneIn(25, 'Nihil shard', [1, 20])
-	.oneIn(100, 'Rune sword')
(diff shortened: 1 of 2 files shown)


gc:
Good work, thank you.

PR #3740, opened 2022-04-03
Fixes `=m ironman` bug that wasn't properly clearing all relevant data.

themrrobert:
### Description:

When using the ironman command, under certain circumstances, some data fails to be reset.
This fixes that, and also automatically applies the AgeBypass to irons

### Changes:

- Always processes all Database commands, instead of aborting on fail.
- Adds BypassAge bitfield on the fresh iron

### Other checks:


gc:
Closing this for reasons discussed.

PR #3744, opened 2022-04-05
Prevent negative sell prices

themrrobert:
### Description:

Because tax is rounded up, if you sell 1x Mind rune, for example (worth 0.24 bot price), then 0.24 - 1 = -.76, which is rounded down to -1.

### Changes:

Math.max(0, sellPrice); to make 0 the absolute floor sell price.

### Other checks:


gc:
Fixed from osb

PR #3761, opened 2022-04-08
Rc stam fix

TastyPumPum:
Previously you could click the usestaminas no flag on normal runes and it would slow the trip down by 3x. This PR fixes that.

Added an optional nostams on the runes to allow you to specify which can't use staminas. 
Made it so it overrides what the user selects on the usestams optional boolean, pushes a message advising Stamina's can't be used with this rune.


src/lib/skilling/skills/runecraft.ts
```diff
@@ -14,6 +14,7 @@ export interface Rune {
 	tripLength: number;
 	inputRune?: Bank;
 	inputTalisman?: Bank;
+	nostams?: boolean;
```
gc, line 17: needs to be camelcased, and it might be better to not do a double-negative `noStams: false`, if you do `stams: true` it would be only a few that need to have it on them too?

gc (changes requested):

src/lib/skilling/skills/runecraft.ts
TastyPumPum: Good idea, resolved the changes requested

PR #3771, opened 2022-04-10
Fix more nex bugs (Nex wipes totally broken)

themrrobert:
### Description:

Fixed a bunch of Nex bugs related to Nex wipes

### Changes:

- Fixes `+rp checkmasses` not showing the 'fake duration' when a wipe happens
- Now will only give loot for successful kills, instead of all kills in a fraction of the time.
- Now trip will end at a random point in the middle of the Wiped kill, instead of at the very end. (Between 5-95%)
- Fixes the initial mass-send messaging so it uses 'fake duration' for both Kills/hr and duration text.

### Other checks:

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -33,7 +33,12 @@ import { POHBoosts } from './poh';
 import { Rune } from './skilling/skills/runecraft';
 import { SkillsEnum } from './skilling/types';
 import { ArrayItemsResolved, Skills } from './types';
-import { GroupMonsterActivityTaskOptions, RaidsOptions, TheatreOfBloodTaskOptions } from './types/minions';
+import {
+	GroupMonsterActivityTaskOptions,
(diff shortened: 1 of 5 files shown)


src/lib/util.ts
```diff
@@ -265,6 +270,10 @@ export function isTobActivity(data: any): data is TheatreOfBloodTaskOptions {
 	return 'wipedRoom' in data;
 }
 
+export function isNexActivity(data: any): data is NexTaskOptions {
+	return 'wipedKill' in data;
```
gc, line 274: This should definitely be a *bit* more robust, highly likely something else could have 'wipedKill'. e.g. check 2-3 properties? I will just commit this myself to save time.

themrrobert: Yea I checked to make sure nothing else was already using it, but you're right in the future could definitely happen

PR #3774, opened 2022-04-11
Mining Rework

Lajnux:
### Description:

Remade how mining is calculated in the bot and added buyable pickaxes and new mining objects like granite/sandstone

### Changes:

- Changed Minable objects

- Added buyable pickaxes

- Added more Minable objects

- Balanced xp/rates, mining rates to more accurate reflect in game.

### Other checks:


src/lib/util.ts
```diff
@@ -144,10 +144,51 @@ export function convertXPtoLVL(xp: number, cap = 99) {
...
 
-export function determineScaledOreTime(xp: number, respawnTime: number, lvl: number) {
-	const t = xp / (lvl / 4 + 0.5) + ((100 - lvl) / 100 + 0.75);
-	return Math.floor((t + respawnTime) * 1000) * 1.2;
+	const bankTime = ore.bankingTime;
+	const chanceOfSuccess = ore.slope * lvl + intercept;
+	const respawnTimeOrPick = ticksBetweenRolls > ore.respawnTime ? ticksBetweenRolls : ore.respawnTime;
+
+	let newQuantity = 0;
+	while (timeElapsed < user.maxTripLength('Mining') / (Time.Second * 0.6)) {
```
themrrobert, line 169: Okay so I have a lot of problems with this.

0. How can you mine just a few rocks with this? It forces a full trip down your throat every time.
1. This is extremely inefficient. While it does simulate well, this is going to be calculating 2000 iterations for a 40 minute trip. (3000/hr)
2. This returns the number of successful ore mined which is then displayed to the user, so the user will likely cancel until they get a high number of rocks, especially with things like Runite/Amethyst. 
3. These factors compound when a user cancels and retries multiple times, running thousands of iterations per attempt.

This needs to be thought out more. Probably the best way to handle 0+2 are:
- Do the 'fakeDuration' thing (don't forget the +m status message too)
- Accept quantities and let the loop go until maxTripLenth OR quantity specified is reached.

As for issues 1+3, I think we need to think of some optimizations, potentially at the cost of some accuracy, although might be possible to optimize without losing accuracy.

**Ooh here's a good idea** Maybe we can make a few fakeDuration mode where it says, "Trip will take between XX and YY minutes"  [obviously storing both for consistency with +m status]
And  the trip will finish sometime in the middle. 

Best way to achieve this? 
```ts
const [, realTime] = determineMiningTime(qty, etc);
const minTime = reduceNumByPercent(realTime, 25);
const maxTime = increaseNumByPercent(realTime, 25);
const fakeMin = addVariation(minTime, 20);
const fakeMax = addVariation(minTime, 20);
console.log(`Trip will take between ${formatDuration(fakeMin)} and ${formatDuration(fakeMax)}`);
```

This guarantees that the real time is between the min+max, and scales proportionally by trip length

src/tasks/minions/miningActivity.ts
```diff
@@ -119,7 +132,7 @@ export default class extends Task {
 			user,
 			channelID,
 			str,
-			['mine', [quantity, ore.name], true],
+			['mine', [duration > 0.9 * user.maxTripLength('Mining') ? null : quantity, ore.name], true],
```
themrrobert, line 135: This isn't enough to prevent quantity from 'overflowing' with some bad RNG. (after quantity is implemented obviously)
The +mine QTY ore command needs to handle excess gracefully, because we don't want ppl 'cheating' the system retrying. Maybe it hides the amount as well as gives a fake time. Just say, 'mining for x minutes'

themrrobert (changes requested):
Okay we need some major changes. See my giant comment in src/lib/util.ts


Read it carefully there's a lot of stuff that applies in many places

src/lib/util.ts
Lajnux: Found a solution based on information from this feedback.

src/tasks/minions/miningActivity.ts
Lajnux: this only makes it so that it always tries to do max trip if duration is above 90%, this because otherwise some new trips will be stuck with lower quantities even tho the person pressing the ''Y'' want a full 30-40 min trip. But with the other masking that suggested earlier I think this is okay.

src/lib/skilling/skills/mining.ts
```diff
@@ -61,61 +96,104 @@ const ores: Ore[] = [
...
 		petChance: 211_886,
 		clueScrollChance: 211_886
 	},
+	{
+		level: 40,
+		xp: 65,
+		id: 444,
+		name: 'Gold ore',
+		respawnTime: 5,
+		bankingTime: 20,
```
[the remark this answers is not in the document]
Lajnux: Great idea! Implemented for banking

src/mahoji/commands/mine.ts
```diff
@@ -81,68 +227,141 @@ export const mineCommand: OSBMahojiCommand = {
...
 			}
 		}
-		// Give gem rocks a speed increase for wearing a glory
-		if (ore.id === 1625 && userHasItemsEquippedAnywhere(user, 'Amulet of glory')) {
-			timeToMine = Math.floor(timeToMine / 2);
-			boosts.push('50% for having an Amulet of glory equipped');
+
+		let armourEffect = 0;
+		for (const armour of varrockArmours) {
+			if (hasItemsEquippedOrInBank(user, [armour.id])) {
```
gc, line 285: dont nest this, invert it

src/mahoji/commands/mine.ts
```diff
@@ -81,68 +227,141 @@ export const mineCommand: OSBMahojiCommand = {
...
+		if (skills.crafting >= 99 && (ore.name === 'Gold ore' || ore.name === 'Silver ore')) {
+			goldSilverBoost = true;
+			boosts.push(`**70%** faster ${ore.name} banking for 99 Crafting`);
 		}
 
-		const maxTripLength = calcMaxTripLength(user, 'Mining');
+		let miningCapeEffect = 0;
+		if (hasItemsEquippedOrInBank(user, [miningCape.id])) {
+			for (const [name, value] of Object.entries(miningCape)) {
+				if (name === ore.name) {
```
gc, line 309: invert this

src/mahoji/commands/mine.ts
```diff
@@ -15,36 +14,173 @@ import { getSkillsOfMahojiUser, mahojiUsersSettingsFetch } from '../mahojiSettin
...
+		'Gold ore': 0,
+		Granite: 0,
+		'Mithril ore': 0,
+		'Adamantite ore': 0,
+		'Runite ore': 0,
+		Amethyst: 0
+	}
+];
+
+const miningCape = {
```
gc, line 167: this variable needs to be renamed to a descriptive name which describes what it actually is.

src/lib/util.ts
```diff
@@ -152,10 +154,53 @@ export function convertXPtoLVL(xp: number, cap = 99) {
...
+	const bankTime = goldSilverBoost ? ore.bankingTime / 3.3 : ore.bankingTime;
+	const chanceOfSuccess = ore.slope * lvl + intercept;
+	const respawnTimeOrPick = ticksBetweenRolls > ore.respawnTime ? ticksBetweenRolls : ore.respawnTime;
+
+	let newQuantity = 0;
+
+	while (timeElapsed < calcMaxTripLength(user, 'Mining') / (Time.Second * 0.6)) {
+		while (!percentChance(chanceOfSuccess)) {
+			timeElapsed += ticksBetweenRolls;
+		}
```
gc, line 184: put tick-duration part outside as its own variable.

src/mahoji/commands/mine.ts
```diff
@@ -81,68 +227,141 @@ export const mineCommand: OSBMahojiCommand = {
...
-				boosts.push(`${pickaxe.reductionPercent}% for ${itemNameFromID(pickaxe.id)}`);
+				currentPickaxe = pickaxe;
+				boosts.push(`**${pickaxe.ticksBetweenRolls}** ticks between rolls for ${itemNameFromID(pickaxe.id)}`);
 				break;
 			}
 		}
+
+		if (currentPickaxe === null) {
+			return 'You need to own a pickaxe suitable for your mining level.';
+		}
```
gc, line 263: We don't require this. just default to a bronze pickaxe if they have none.

src/mahoji/commands/mine.ts
```diff
@@ -81,68 +227,141 @@ export const mineCommand: OSBMahojiCommand = {
...
+
+		let glovesRate = 0;
 		if (skills.mining >= 60) {
 			for (const glove of gloves) {
 				if (userHasItemsEquippedAnywhere(user, glove.id)) {
-					timeToMine = reduceNumByPercent(timeToMine, glove.reductionPercent);
-					boosts.push(`${glove.reductionPercent}% for ${itemNameFromID(glove.id)}`);
-					break;
+					for (const [name, value] of Object.entries(glove)) {
+						if (name === ore.name) {
```
gc, line 270: invert this

gc (changes requested):

src/mahoji/commands/mine.ts
```diff
@@ -15,36 +15,173 @@ import { mahojiUsersSettingsFetch } from '../mahojiSettings';
...
 	}
 ];
 
+const miningCapeOreEffect = {
+	id: itemID('Mining cape'),
+	Clay: 5,
+	'Copper ore': 5,
+	'Tin ore': 5,
+	'Iron ore': 5,
+	'Silver ore': 5,
```
gc, line 174: this object makes no sense, or atleast the prescence of `id` - shouldnt mix unrelated/random things together.  You can use a `Bank` for this.

gc (changes requested):

PR #3777, opened 2022-04-12
Capture Users Last Loot

Superwhammybar:
### Description:

When getting loot via bank images, it can be hard to discern what it is that has been received if you don't recognise the item. To combat this I've added a entry in the DB which tracks what the last loot that was added to the users bank was, excluding impling captures. This can be returned by a new `/minion` command `last_loot` in 3 different formats:
* Text - Text string of loot, strings over 500 characters are sent as .txt instead
* Names - Bank image with names
* Image - Bank image as normal

### Changes:

Add lastLoot entry to Prisma Schema and UserSchema
Add LastLoot option for UserSettings
Updated `addItemsToBank` function to track the last received loot in the aforementioned field, only if it's not a impling capture.
Add a command to `/minion` called `last_loot` which returns the loot in 3 different formats mentioned above.

### Other checks:
Checked it with a large amount of commands and all loot seems to be appropriately tracked.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -247,6 +247,7 @@ model User {
   lastTearsOfGuthixTimestamp BigInt   @default(1)
   sacrificedValue            BigInt   @default(0)
   bank                       Json     @default("{}") @db.Json
+  lastLoot                   Json     @default("{}") @db.Json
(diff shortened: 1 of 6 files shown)


TastyPumPum:
Bump pls 🙏

gc:
Don't plan on adding this currently, sorry.

PR #3778, opened 2022-04-12
Obis return message

DarkWorldsArtist:
### Description:

adds a message to make trips where you receive an obis stand out

### Other checks:


gc:
This includes the ignecarus scale change

PR #3791, opened 2022-04-14
Fix 3x +m data commands

themrrobert:
### Description:

3 of the `+m data` commands are broken because of a Prisma bug. I have got to the bottom of it, and there's nothing we can do on our end except to rewrite our calls using `$queryRawUnsafe()`

### Changes:

Fixes the follow +m data commands:
- Personal Monster KC
- Personal Inferno Death Times
- Personal Farmed Crops

### Other checks:


gc:
We could just do `user_id::text`? But yea not troubled with the unsafe here.

themrrobert:
> yes that should work, too hehe. I really tried to see if there was a way prisma would work properly, maybe i'll submit a PR over there

PR #3793, opened 2022-04-15
Updated GWD/Bandos Bank BG Requirements

6enny6:
Added all nex collection log to the requirements for the gwd/bandos bank bg
Added ancient godsword to the item cost requirement


gc:
This is a pretty random change. Not sure how I feel about it, would also possibly want to remove it from all the current owners?

6enny6:
its not that random, the bg is for completing gwd collection log, which now includes Nex.

I think it makes sense to remove from current owners too.

6enny6:
Actually i am mistaken, our GWD collection log does not have Nex on yet and im not sure that log is even on OSRS, although i still think it would be s fitting change

PR #3804, opened 2022-04-18
Fix Revenants on BSO - Hellfire bow multiplier, cost, etc

themrrobert:
### Description:

Fixes Hellfire bow on Revs. The boost was missing, the cost was missing

### Changes:

- Replaces Hellfire bow boost
- Adds the Hellfire arrow cost back

### Other checks:


gc:
Opinionated changes shouldn't be included like this in a PR thats just to fix a bug.

themrrobert:
> Opinionated changes shouldn't be included like this in a PR thats just to fix a bug.

Fair and fixed

PR #3810, opened 2022-04-19
Achievement Diary Rewards - Buyable/CL

### LINKED ISSUE #2135: Achievement diary gear
## 🚀 Feature Proposal

A way to get multiple items from the achievement diary gear plz

### LINKED ISSUE #3781: Move diary rewards to buy cmd
Move the diary rewards (radas blessing etc) to `+buy`, this will resolve the issue of not being able to get more than 1. 

On OSRS you can buy as many as you want from Perdu. 
https://oldschool.runescape.wiki/w/Perdu

TastyPumPum:
Adds the achievement diary rewards to the buy command provided they have done the original claim.
Adds a cl to track the achievement diary rewards (in custom and uncounted)

Closes #3781 (This might not be the ideal way but I'm open to feedback)
Closes #2135


src/lib/data/buyables/perdu.ts
```diff
@@ -0,0 +1,293 @@
+import resolveItems from '../../util/resolveItems';
+import { Buyable } from './buyables';
+
+export const perduBuyables: Buyable[] = [
+	{
+		name: 'Karamja gloves 1',
+		gpCost: 1000,
+		ironmanPrice: 200,
+		collectionLogReqs: resolveItems(['Karamja gloves 1'])
+	},
```
gc, line 10: I think this entire file could be replaced with just:
```ts
diariesCL.map(itemName => ({ name: itemName, gpCost: 1000, ... })
```

gc (changes requested):

src/lib/data/buyables/perdu.ts
TastyPumPum: Resolved the above comment

src/lib/data/buyables/perdu.ts
```diff
@@ -1,293 +1,9 @@
...
-		gpCost: 1000,
-		ironmanPrice: 200,
-		collectionLogReqs: resolveItems(["Rada's blessing 4"])
-	}
-];
+export const perduBuyables: Buyable[] = diariesCL.map(itemName => ({
+	name: itemName,
+	gpCost: 1000,
+	ironmanPrice: 200,
+	collectionLogReqs: [itemName]
```
gc, line 8: you cant change this to be just strings/item names. (1) any incorrect item names wont cause any error until someone actually tries to buy one (2) its a standard that we always use item IDs, using `resolveItems()`, for things like this.

gc (changes requested):

src/lib/data/buyables/perdu.ts
TastyPumPum: It worked in testing so I assumed it was good, my apologies. I think I have made the relevant changes, and it still works as it should in testing

PR #3813, opened 2022-04-20
Fix bug while converting to BSO ironman

themrrobert:
### Description:

When using the ironman command, under certain circumstances, some data fails to be reset.
This fixes that, and also automatically applies the AgeBypass to irons

### Changes:

- Always processes all Database commands, instead of aborting on fail.
- Adds BypassAge bitfield on the fresh iron
- Logs to Sentry if a DB command actually fails

### Other checks:

### DIFF
--- a/src/lib/minions/functions/becomeIronman.ts
+++ b/src/lib/minions/functions/becomeIronman.ts
@@ -51,21 +51,15 @@ Type \`confirm permanent ironman\` if you understand the above information, and 
 
 		await msg.author.settings.reset();
 
-		try {
-			await Promise.all([
-				prisma.slayerTask.deleteMany({ where: { user_id: msg.author.id } }),
-				prisma.playerOwnedHouse.delete({ where: { user_id: msg.author.id } }),
-				prisma.minigame.delete({ where: { user_id: msg.author.id } }),
-				prisma.xPGain.deleteMany({ where: { user_id: BigInt(msg.author.id) } }),
-				prisma.newUser.delete({ where: { id: msg.author.id } }),
-				prisma.activity.deleteMany({ where: { user_id: BigInt(msg.author.id) } }),
-				prisma.tameActivity.deleteMany({ where: { user_id: msg.author.id } }),
-				prisma.tame.deleteMany({ where: { user_id: msg.author.id } }),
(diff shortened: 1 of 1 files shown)


src/lib/minions/functions/becomeIronman.ts
```diff
@@ -52,23 +55,26 @@ Type \`confirm permanent ironman\` if you understand the above information, and
 
 		try {
 			await prisma.slayerTask.deleteMany({ where: { user_id: msg.author.id } });
-			await prisma.playerOwnedHouse.delete({ where: { user_id: msg.author.id } });
-			await prisma.minigame.delete({ where: { user_id: msg.author.id } });
+			prisma.playerOwnedHouse.delete({ where: { user_id: msg.author.id } }).catch(noOp);
+			prisma.minigame.delete({ where: { user_id: msg.author.id } }).catch(noOp);
 			await prisma.xPGain.deleteMany({ where: { user_id: BigInt(msg.author.id) } });
-			await prisma.newUser.delete({ where: { id: msg.author.id } });
+			prisma.newUser.delete({ where: { id: msg.author.id } }).catch(noOp);
```
gc, line 61: Don't think the `await's` need to be removed on these

src/lib/minions/functions/becomeIronman.ts
```diff
@@ -52,23 +55,26 @@ Type \`confirm permanent ironman\` if you understand the above information, and
...
+			await msg.author.settings.update([
+				[UserSettings.Minion.Ironman, true],
+				[UserSettings.Minion.HasBought, true],
+				[UserSettings.BitField, BitField.BypassAgeRestriction]
+			]);
+			return msg.channel.send('You are now an ironman.');
+		} catch (e) {
+			// Actual database error
+			logError(e, { command: 'minion ironman', user_id: msg.author.id });
+			return msg.channel.send('We are in an unknown state. Likely inconsistent.');
```
gc, line 76: This isn't really an appropriate user error message. Something like: `An error occurred trying to make you an Ironman, please contact us for help.` would be better

gc (changes requested):

src/lib/minions/functions/becomeIronman.ts
themrrobert: Oops that was supposed to be internal message only, alright fixing these up.

PR #3820, opened 2022-04-21
Collecting - White Berries

TastyPumPum:
Adds the white berries to the collectables, with the required qp and relevant skill levels to do the quests.
Used rates from the wiki - https://oldschool.runescape.wiki/w/Money_making_guide/Collecting_white_berries


src/commands/Minion/collect.ts
```diff
@@ -64,6 +64,20 @@ export const collectables: Collectable[] = [
...
+		skillReqs: {
+			ranged: 60,
+			thieving: 50,
+			agility: 56,
+			crafting: 10,
+			fletching: 5,
+			cooking: 30
+		},
+		duration: Time.Minute * 4.05
+	},
```
gc, line 80: Should require some runes or dueling rings or something for banking, no?

gc (changes requested):

PR #3826, opened 2022-04-23
BSO - Revert Nex Items for components

TastyPumPum:
Following the positive feedback from the #bso-vote poll.
This now allows you to breakdown the Nex drops for 1 of the relevant component, this makes it so the original way is still the most beneficial and preferred, but gives irons a more realistic way to get components and gives dupes a use. (Especially ancestral components)

### DIFF
--- a/src/lib/data/creatables/bsoItems.ts
+++ b/src/lib/data/creatables/bsoItems.ts
@@ -53,16 +53,16 @@ const nexCreatables: Createable[] = [
 			[itemID('Armadylean components')]: 1
 		}
 	})),
-	...brokenTorvaOutfit.map(part => ({
-		name: `Revert ${getOSItem(part).name}`,
-		inputItems: new Bank().add(part),
+	...brokenTorvaOutfit.map(piece => ({
+		name: `Revert ${getOSItem(piece).name}`,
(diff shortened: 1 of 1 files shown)


src/lib/data/creatables/bsoItems.ts
```diff
@@ -135,6 +135,141 @@ const componentRevertables: Createable[] = [
...
+	},
+	{
+		name: 'Revert torva full helm (broken)',
+		inputItems: {
+			[itemID('Torva full helm (broken)')]: 1
+		},
+		outputItems: {
+			[itemID('Bandosian components')]: 1
+		}
+	},
```
gc, line 147:
```ts
brokenTorvaOutfit.map(piece => ({ name: `Revert ${itemName(piece)`, input: new Bank().add(piece), ... }))
```

Would be significantly shorter and I think easier to maintain.

gc (changes requested):

src/lib/data/creatables/bsoItems.ts
TastyPumPum: Have done it in a similar way to line 50 to keep it more consistent. Have testing it in the testing server let me know if you want me to change it like above though.

Issue #3618, opened 2022-02-22
Emerged inferno and range master cape 


22rifleman:
## 🐛 Bug Report

Ranged master cape equipped in range slot vs avas assembler in range slot. Why should ranged master cape consume more supplies?
![B6077A66-72A0-4948-BFFC-E250D2B52085](https://user-images.githubusercontent.com/91299009/155203060-29a9d9d4-6c5e-4a5b-9b82-48673fe79049.jpeg)
![AE0DD64F-931F-412B-9C4E-D7A5098E8E1A](https://user-images.githubusercontent.com/91299009/155203068-95d33a76-003a-409d-b89e-c68f7d3d0083.jpeg)

PR #3843, opened 2022-04-29
BSO - added range cape check to reduce arrow & dart cost

### LINKED ISSUE #3827: Ranged Master Cape at Inferno
## 🐛 Bug Report

It does not appear as if the ranged master cape ammo saving perk works on normal inferno. 

With Ranged Master Cape: 
image

With Assembler:
image

lordzuku:
### Description:
added master range cape effect to work on zuk

### Changes:

passed a boolean for if a master range cape is worn to the cost object, checked if true & then modified the cost per hour for darts & arrows
### Other checks:

-   [Y ] I have tested all my changes thoroughly.
fixes 
#3827 
#3618


6enny6:
should this be added to lib/data/similaritems.ts ?

lordzuku:
> should this be added to src/lib/data/similaritems.ts ?
> 
> make ranger master act as an assembler

iirc the range max cape shouldnt act as an assembler, otherwise the ammo saving bonuses might stack, would probably need it coded into future commands as a separate ammo saving systems

anthonyburrow:
> would probably need it coded into future commands as a separate ammo saving systems

I believe this already exists in the "specialRemoveItems" function which is called in inferno.ts, however it only accounts for the Assembler:

https://github.com/oldschoolgg/oldschoolbot/blob/10a15038ed377c1a6f5f0a43c466973b030c7e96/src/extendables/User/Bank.ts#L142

I would probably recommend making this function more generalized for BSO additions by adding the Ranged Master Cape percentage in this cost function instead of making a new condition for just the inferno.

lordzuku:
> > would probably need it coded into future commands as a separate ammo saving systems
> 
> I believe this already exists in the "specialRemoveItems" function which is called in inferno.ts, however it only accounts for the Assembler:
> 
> https://github.com/oldschoolgg/oldschoolbot/blob/10a15038ed377c1a6f5f0a43c466973b030c7e96/src/extendables/User/Bank.ts#L142
> 
> I would probably recommend making this function more generalized for BSO additions by adding the Ranged Master Cape percentage in this cost function instead of making a new condition for just the inferno.

unless im mistaken (which i might be, just recently started looking over the project)
that would only apply for regular m k trips & would still need a manual patch for fixing the inferno.

currently there is an already built system for m k trips for range cape saving

PR #3876, opened 2022-05-07
Open all/until

gc:
### Other checks:


src/mahoji/lib/abstracted_commands/openCommand.ts
```diff
@@ -0,0 +1,184 @@
...
+		loot,
+		messages: [
+			`You opened ${amountOpened}x ${openable.openedItem.name}, ${
+				loot.has(openUntil.id)
+					? `until you got a ${openUntil.name}!`
+					: `but you didn't get a ${openUntil.name}!`
+			}`
+		],
+		openables: [openable],
+		kcBank: new Bank().add(openable.openedItem.id)
```
themrrobert, line 86: Shouldn't this be:
```ts
kcBank: new Bank().add(openable.openedItem.id, amountOpened)
```

themrrobert (changes requested):
1 bug and 1 suggestion.

PR #3889, opened 2022-05-10
BSO - Doug 4x ore when mining

TastyPumPum:
Following the positive #bso-votes poll
Adds a perk of doug multiplying ores by 4x.


TastyPumPum:
Magna isn't interested in this.

Issue #3909, opened 2022-05-15
Giveaway timer


6enny6:
Add the timer system thing from `/farming check_patches` where it shows the realtime to the giveaway command, so you can see exactly when it will finish.

This is it currently

![image](https://user-images.githubusercontent.com/76587427/168464139-06f1bfef-7f06-4b50-a409-9a22125136d6.png)

the timers im referencing vvv

![image](https://user-images.githubusercontent.com/76587427/168464188-1a66e4f2-03ad-4d05-93ab-a866dbe26bef.png)

6enny6:
this could be added to all minion trip sending messages too

![image](https://user-images.githubusercontent.com/76587427/168479132-3f5bc78c-e080-4b3d-8078-ddaea6a47ef8.png)

PR #3911, opened 2022-05-15
BSO Giveaway Timer Format Change

Superwhammybar:
https://github.com/oldschoolgg/oldschoolbot/issues/3909

### Description:

The current format involves people having to do maths, and people are lazy.

### Changes:

Use the time builder (similar to check_patches command) to construct both the finishing date/time and the relative finish time when the message is viewed.

Now looks like this:
image

### Other checks:

### DIFF
--- a/src/commands/Minion/giveaway.ts
+++ b/src/commands/Minion/giveaway.ts
@@ -71,10 +71,10 @@ export default class extends BotCommand {
 		}
 
 		const message = await msg.channel.sendBankImage({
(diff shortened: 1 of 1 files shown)


gc:
Why is this a BSO pr? it doesnt look bso-specific

Superwhammybar:
You're right, in my head I assumed it was BSO for some reason. I will fix.

Superwhammybar:
Remade here https://github.com/oldschoolgg/oldschoolbot/pull/3917

PR #3925, opened 2022-05-17
BSO - Fix DG XP and Rates Displayed on Return

anthonyburrow:
### Description:

Currently the XP, XP rate, and token rate shown on the Dungeoneering trip return message are not correct, as 1. the XP does not account for master cape multiplier and 2. the rates shown were only the XP or tokens from the trip itself instead of an actual rate.

This fixes most said problems, and the end trip message is much more understandable, however the XP *rate* still does not account for the master cape multiplier. To keep the same format as before, you may need some way of returning the raw XP value calculated from addXP(). A change like this would also help fix the incorrect Wintertodt and Farming XP shown too.

### Changes:

- Make use of the addXP return for the XP shown.
- Use the correct variable in rawXPHr calculation so that an actual rate is given.
- Calculate the tokens per hour directly instead of letting Gorajan XP boost influence the result beforehand.

### Other checks:

### DIFF
--- a/src/tasks/minions/dungeoneeringActivity.ts
+++ b/src/tasks/minions/dungeoneeringActivity.ts
@@ -44,17 +44,20 @@ export default class extends Task {
 				bonusXP += Math.floor(xp * (gorajanEquipped / 2));
 				xp += bonusXP;
 			}
-			await u.addXP({
+			const xpStr = await u.addXP({
 				skillName: SkillsEnum.Dungeoneering,
 				amount: xp / 5,
-				duration
+				duration,
+				minimal: true
 			});
 			await u.settings.update(
 				UserSettings.DungeoneeringTokens,
 				u.settings.get(UserSettings.DungeoneeringTokens) + tokens
 			);
-			let rawXPHr = (xp / (duration / Time.Minute)) * 60;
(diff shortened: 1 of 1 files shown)


src/tasks/minions/dungeoneeringActivity.ts
```diff
@@ -44,17 +44,20 @@ export default class extends Task {
...
+				minimal: true
 			});
 			await u.settings.update(
 				UserSettings.DungeoneeringTokens,
 				u.settings.get(UserSettings.DungeoneeringTokens) + tokens
 			);
-			let rawXPHr = (xp / (duration / Time.Minute)) * 60;
-			rawXPHr = Math.floor(xp / 1000) * 1000;
+			let rawXPHr = (xp / minutes) * 60;
+			rawXPHr = Math.floor(rawXPHr / 1000) * 1000;
```
gc, line 58: Is this going to result in the xp/hr being shown twice in the message? I tried looking for your testing of this PR in the testing server, but couldnt find any dung related stuff.

gc (changes requested):

gc:
Closing due to no response. Feel free to re-open if you wish.

PR #3944, opened 2022-05-24
Fix T6 on patreon

themrrobert:
### Description:

Fixes T6 so it works properly on Patreon

### DIFF
--- a/src/tasks/patreon.ts
+++ b/src/tasks/patreon.ts
@@ -105,7 +105,7 @@ export default class PatreonTask extends Task {
 
 		const bitFieldToRemove = bitFieldFromPerkTier(from);
 		const bitFieldToAdd = bitFieldFromPerkTier(to);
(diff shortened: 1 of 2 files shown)


src/tasks/patreon.ts
```diff
@@ -31,6 +31,7 @@ patreonApiURL.search = new URLSearchParams([
 ]).toString();
 
 export const tiers: [PatronTierID, BitField][] = [
+	[PatronTierID.Six, BitField.IsPatronTier5],
```
gc, line 34: is this meant to be t6 = t5?

gc (changes requested):

src/tasks/patreon.ts
themrrobert: No, because there is no PatreonTier6 as far as the bot is concerned, it maps to Tier5, just like the github t6 does.

Issue #3953, opened 2022-05-26
+m clue buttons


6enny6:
The three clue buttons on `+m` should prioritise by highest tier you own. Currently it only shows Hard/Elite/Master. 

For example if you owned every tier of clue it would still show Hard/Elite/Master, but if you owned an Easy, Medium and an Elite, it would show those three as buttons instead

PR #3956, opened 2022-05-26
adjust clueFilter minion buttons

Lajnux:
### Description:

Changed so that it creates up to 3 buttons for the highest tiered clue scrolls that are owned.

### Changes:

Adjusted the filter for ClueTiers on the minion command.

### Other checks:


Lajnux:
resolves #3953

PR #3968, opened 2022-05-30
Fixes gold deletion bug when users table row missing.

themrrobert:
### Description:

Currently, if you sent money to a user/id that doesn't have a row in the users table, it will delete the money from the sender, and return an error.

### Changes:

- Runs mahojiUsersSettingsFetch() command on the recipient 
-    (This initializes the users' row the same way a user running any command does)

### Other checks:


gc:
Hacky fix but will have to work for now.

PR #3976, opened 2022-06-04
Allow Slepey tablet drop if you don't have the CL

### LINKED ISSUE #3319: Slepey Tablet drop on BSO phosani
Currently, if you own a slepey tablet or have used one, you can't get it as a drop from phosani in BSO.
since we can trade untradeables in BSO this can be a issue for CL hunters that go for all slots, me and a friend both had bought a slepey tablet to speed up phosani grind to get it. right now the only way of obtaining a new unique tablet is through umb's, which goes against the idea of the CL bosses. please make it so that we can still obtain slepey tablets even after having obtained or used 1.

themrrobert:
### Description:

This doesn't affect OSB, which is why it's only being merged to BSO.
Currently, if you traded for a slepey tablet to activate it for the boost, you can not obtain it for the CL.

This corrects that problem.

### Changes:

- Allows the Slepey tablet drop if you don't have it on the CL, (along with the exist restrictions)

### Other checks:

Closes #3319

### DIFF
--- a/src/tasks/minions/minigames/nightmareActivity.ts
+++ b/src/tasks/minions/minigames/nightmareActivity.ts
@@ -70,7 +70,10 @@ export default class extends Task {
 					deaths[user.id] = deaths[user.id] ? ++deaths[user.id] : 1;
 					kcAmounts[user.id]--;
 				} else {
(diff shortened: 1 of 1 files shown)


TURB042O:
oh I love you

PR #3977, opened 2022-06-04
200M xp pet rates

Lajnux:
### Description:

Players with [200 million experience](https://oldschool.runescape.wiki/w/200_million_experience) in the corresponding skill will be fifteen times more likely in receiving the skilling pet.

Added function that calculates drop rate for skilling pets both as number and loot table.

### Changes:

Players with [200 million experience](https://oldschool.runescape.wiki/w/200_million_experience) in the corresponding skill will be fifteen times more likely in receiving the skilling pet.

Added function that calculates drop rate for skilling pets both as number and loot table.

### Other checks:


gc (changes requested):
This is a DRY disaster. Make a function which does this, like 
```ts
function skillingPetChance(user: User | KlasaUser, skill: SkillsEnum, baseDropRate: number) {
   const xp = ...;
   if (xp < MAX_XP) return baseDropRate;
   return baseDropRate / 15;
}
```

gc (changes requested):
1. dont typecast so many things
2. you cannot do this to loottables in this way. just leave loottables out, or do what you did within this PR itself, remove it from the table and roll it manually.

gc (changes requested):
should add a test for this. make sure it is giving the right droprates

Issue #3026, opened 2021-10-16
No Food Boost


osrsolympus:
## 🚀 Feature Proposal

Some mobs don’t use food (for example: goblins, imps, shades). They should be updated to get the 4% bonus for “no food” currently only given to mobs which have a base food cost which is somehow mitigated.

PR #3978, opened 2022-06-04
boost monster that require no food

Lajnux:
### Description:

boost monster that require no food

### Changes:

boost monster that require no food

### Other checks:


Lajnux:
resolve #3026

src/mahoji/lib/abstracted_commands/minionKill.ts
```diff
@@ -463,6 +463,9 @@ export async function minionKillCommand(
 				return e.message;
 			}
 		}
+	} else {
+		boosts.push('4% for no food');
+		duration = reduceNumByPercent(duration, 4);
 	}
```
gc, line 469: pointless, duplicated `else`. use the above one thats already there.

gc (changes requested):

src/mahoji/lib/abstracted_commands/minionKill.ts
Lajnux: can't use the above one since nested.

PR #3987, opened 2022-06-05
Improve Grand exchange filter response message

themrrobert:
### Description:

Currently, when a message is removed from GE, you get a generic message that isn't always helpful, and doesn't cover all possible reasons the message could be removed.

I initially planned to expand this with mod-enabled commands to blacklist emojis on the fly, but since we're going to lose the readmessage intent in the near future anyway, I figured this was a good solution to hold us over until then.

**Response to criticism:** 450 characters is already generous for GE messages. This limit is intentionally there to make you choose carefully which ads and emojis you want to use in your messaging. There is also a discord character limit that could easily be overrun if emojis didn't count toward the character limit.

### Changes:

 - Restricts usage of the red/blue alarm, and **all** animated emojis.
 - The DM you get from the bot now tells you exactly why your ad was removed.
 - An explanation is now given for long messages that could be from external emojis using a lot of characters.
 - Your message length is now displayed if that was the reason your message was rejected.

### Other checks:

Examples: (Users message only shown for example, the DM does not contain this obviously)
image
image

### DIFF
--- a/src/monitors/grandExchange.ts
+++ b/src/monitors/grandExchange.ts
@@ -2,6 +2,9 @@ import { KlasaMessage, Monitor, MonitorStore } from 'klasa';
 
 import { Channel, SupportServer } from '../lib/constants';
 
+const filteredCustomEmojis = ['redAlarm', 'blueAlarm'];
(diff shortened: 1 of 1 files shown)


gc:
Closing because I'm putting the filter in robochimp imminently

PR #4015, opened 2022-06-12
ironman sell

Lajnux:
### Description:

Made it possible for ironmen to sell items for the lowest general store stock prices. 

### Changes:

Made it possible for ironmen to sell items for the lowest general store stock prices. 

### Other checks:


gc (changes requested):
Your tests are failing, after i updated the PR, because they seem to expect a non-integer value for the price. which is faulty/bad anyway, GP should always be an int.

gc (changes requested):
since you're now doing the flooring in the sell cmd, looks like this will cause (bad) errors in the sacrifice command, and now needs to be floored there too.

PR #4020, opened 2022-06-13
Only show available bankbgs.

themrrobert:
### Description:

Removes Bank BG's the user can't use from the menu (not including Patron tiers, I left those visible as an up-sell)

### Changes:

- Removes disabled backgrounds from the /bankbg option.

### Important:
 When merging to BSO, the first filter should be changed to:
`.filter(bg => isMod || bg.available || bg.owners?.includes(user.id))`

This will allow the filter to show the user's custom bg.

### Other checks:

### DIFF
--- a/src/mahoji/commands/minion.ts
+++ b/src/mahoji/commands/minion.ts
@@ -1,7 +1,7 @@
 import { FormattedCustomEmoji } from '@sapphire/discord.js-utilities';
 import { ApplicationCommandOptionType, CommandRunOptions } from 'mahoji';
 
-import { MAX_LEVEL, PerkTier } from '../../lib/constants';
+import { BitField, MAX_LEVEL, PerkTier } from '../../lib/constants';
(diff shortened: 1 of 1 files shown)


src/mahoji/commands/minion.ts
```diff
@@ -80,10 +80,12 @@ export const minionCommand: OSBMahojiCommand = {
 					type: ApplicationCommandOptionType.String,
 					name: 'name',
 					description: 'The name of the bank background you want.',
-					autocomplete: async value => {
+					autocomplete: async (value, user) => {
+						const mahojiUser = await mahojiUsersSettingsFetch(user.id);
```
gc, line 84: only select the bitfield column

gc (changes requested):

src/mahoji/commands/minion.ts
themrrobert: done

PR #4026, opened 2022-06-15
Fixes PoH item mount

themrrobert:
### Description:

I have identified and fixed all the issues with this command.

### Changes:

- Identified (reproduced) and fixed dupe exploit
- Fixed logic with adding + removing the right items.
- Switched to the queue'd bank alteration functions, instead of settings.update('bank');
- Adds `getBankOfMissingItems` function to show exactly what you're missing, instead of the full cost.
- **Important:** This is much more efficient than doing `missing = costBank.remove(userBank);` because my function only iterates over the [few] costBank items, while the other way iterates thru the entire bank.

### Other checks:


src/mahoji/commands/poh.ts
```diff
@@ -86,6 +89,25 @@ export const pohCommand: OSBMahojiCommand = {
 					}
 				}
 			]
+		},
+		{
+			type: ApplicationCommandOptionType.Subcommand,
+			name: 'mountitem',
```
gc, line 95: `mount_item`, we use snake_case for these.

src/mahoji/commands/poh.ts
```diff
@@ -86,6 +89,25 @@ export const pohCommand: OSBMahojiCommand = {
...
+				{
+					type: ApplicationCommandOptionType.String,
+					name: 'name',
+					description: 'The item you want to mount.',
+					required: true,
+					autocomplete: async (value, user) => {
+						const mUser = await mahojiUsersSettingsFetch(user.id, { bank: true });
+						const bank = new Bank(mUser.bank as ItemBank);
+						let res = bank.items().filter(i => i[0].name.toLowerCase().includes(value.toLowerCase()));
+						return res.map(i => ({ name: `${i[0].name}`, value: i[0].name.toString() }));
```
gc, line 107: `ownedItemOption` exists for this already

gc (changes requested):

PR #4036, opened 2022-06-18
Leaderboard slash cmd

gc:
### Other checks:


src/mahoji/lib/abstracted_commands/ironmanCommand.ts
```diff
@@ -0,0 +1,120 @@
...
+
+	try {
+		await prisma.user.delete({
+			where: { id: user.id }
+		});
+		await prisma.user.create({
+			data: createOptions
+		});
+		await prisma.slayerTask.deleteMany({ where: { user_id: user.id } });
+		await prisma.playerOwnedHouse.delete({ where: { user_id: user.id } });
```
themrrobert, line 104: Isn't there a bug where this will fail if the user doesn't have a PoH? I thought I changed this to add a `.catch(noOp)` afterwards? 

Same with minigame + newUser (Anything using delete and not deleteMany).

Without this, after the first failure, it will refuse to continue to delete the remaining rows, inhibiting a  full wipe.

themrrobert (changes requested):

PR #4038, opened 2022-06-20
Fix /bsominigames fishing_contest fish when no location specified

themrrobert:
### Description:

Currently, location is optional, and keeping it that way, and merging this PR best preserves the existing behavior to make the switch to slash commands as non-invasive as possible.

With this change, not specifying a location on the `fish` subcommand, is the same as choosing `stats_info`

**Currently, choosing `fish` without a location causing an Exception and error.**

### Changes:
- Handle missing location as a stats_info request.

### Other checks:


gc:
This logic should be in the command itself. Random boolean logic on options like this is confusing and less maintainable.

PR #4083, opened 2022-06-28
Distribute Unique Loot Correctly (Hydra/Totem/Sire)

Superwhammybar:
### Description:

At present the code correctly identifies which of the 3 possible items (for each scenario) the user has the least of and adds the loot to that one. The code fails to update the data it is using to identify the least had item which means it ends up just adding the same one over and over (only if multiple are obtained in one go).

### Changes:

Update `combinedLoot` with the newly added loot to assist in determining the new lowest value.

### Other checks:


themrrobert (approved):
Looks good to me

PR #4109, opened 2022-07-05
[BSO] Add more invention toggles from Player Feedback

themrrobert:
### Description:

While a case can be made to make all of the non-equip-required inventions toggleable, these 3 are the most important, in my opinion. 

**I used the following criteria when deciding which Inventions should be the first to receive toggles:**
1. Cost/rarity of invention materials
2. Frequency with which you may want to disable said invention
3. Ongoing / usage cost of invention materials based on use-cases.

**Dwarven toolkit**: You want to disassemble many different items, but you are being taxed 2 Dwarven parts for each single disassemble, the same as for an entire trip. This can quickly add up and cost several hundred mil to disassemble all items.

**Clue upgrader**: While the materials aren't nearly as rare/expensive, there are oftentimes when you want to exclusively farm a specific type of clue, and not have a % 'upgraded' **especially when medium=>hard is actually a significant downgrade.**

**Master Hammer and chisel:** It is common when doing skilling activities, especially crafting, that you want to control your trip time for various bonuses, like holiday events, doubles, and pet loot (ie. doug, peky, etc). Since a lot of crafting materials are in short supply, you often don't want to speed up your trip.

### Changes:

- Added toggles for Dwarven toolkit, Clue upgrader, and Master hammer and chisel
- Added the bitfield constants of course
- Added the descriptions for the +rp c / minion info screen
- Added the hooks to disable said inventions when toggles are enabled.

### Other checks:

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -348,7 +348,10 @@ export const enum BitField {
 	HasScrollOfTheHunt = 204,
 	HasBananaEnchantmentScroll = 205,
 	HasDaemonheimAgilityPass = 206,
-	DisabledGorajanBoneCrusher = 207
+	DisabledGorajanBoneCrusher = 207,
+	DisabledClueUpgrader = 208,
+	DisabledDwarvenToolkit = 209,
+	DisabledMasterHammerAndChisel = 210
 }
 
 interface BitFieldData {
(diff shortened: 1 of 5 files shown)


gc:
As the above comment says, this already exists, unless im missing something.

themrrobert:
Yes it does, I looked for this command and in the code, of course i see it both places now, I must have had a blind spot XD

PR #4138, opened 2022-07-08
Invention - Flexible materials rebalancing

themrrobert:
### Description:

As discussed, Flexible materials are by far the most hard to get. 
1. They only come from Bows
2. All the bows are woefully underleveled
3. Flax is 20k/each
4. Elder logs are 40k/each (More than 10x more than Magic logs).

**I still believe after this update Flexible will be the rarest and hardest to obtain Material, but at least now it will be possible to stock up enough for a few days of boosted Fishing.**

### Changes:

- Bows / crossbows now have the same level tier as their Fletching requirement, as that's the only way they come into the game.
- Unstrung bows now have the Tier level of the fletching requirement minus 5 levels.
- Reduce usage cost slightly of the MechaRod

### Other checks:


src/lib/invention/groups/Bows.ts
```diff
@@ -8,20 +8,20 @@ export const Bows: DisassemblySourceGroup = {
...
+		{ item: i('Maple shortbow'), lvl: 55 },
+		{ item: i('Maple longbow'), lvl: 60 },
+		{ item: i('Yew shortbow'), lvl: 70 },
+		{ item: i('Yew longbow'), lvl: 75 },
+		{ item: i('Yew comp bow'), lvl: 75 },
+		{ item: i('Magic shortbow'), lvl: 80 },
+		{ item: i('Magic longbow'), lvl: 85 },
+		{ item: i('Elder bow'), lvl: 99 },
 
 		{ item: i('Dark bow'), lvl: 90 },
```
gc, line 26: dark bow should probably be 99 too, if elder bow is 99

themrrobert: Done

PR #4141, opened 2022-07-09
OSB Leagues

gc:
### Other checks:

### DIFF
--- a/prisma/robochimp.prisma
+++ b/prisma/robochimp.prisma
@@ -29,8 +29,9 @@ model User {
     bits Int[]
 
     leagues_completed_tasks_ids Int[]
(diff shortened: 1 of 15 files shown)


prisma/robochimp.prisma
```diff
@@ -28,5 +28,10 @@ model User {
     id   BigInt @id @unique
     bits Int[]
 
+    leagues_completed_tasks_ids Int[]
```
themrrobert, line 31: This will likely need something similar to the startupScripts.ts to set a default value and not null.

I know its robochimp but still probably a good idea.

themrrobert (changes requested):

PR #4155, opened 2022-07-10
Bandosian components

TastyPumPum:
As bandosian components are on the creatables cl, this PR adds an optional boolean that can be added onto reverts to add them on to the cl.


themrrobert:
why 54 commits for the like 3 line changes lol. you gotta reset your root BSO branch and stop using the github button to merge upstream changes.

TastyPumPum:
@gc I note you labelled this as a suggestion but at present the creatables cl is uncompletable unless this gets merged. I guess the alternative would be to exclude bandosian components from the cl

themrrobert:
I personally wish everything respected the noCL flag, like just add `noCl: true` to all the reverts that shouldn't add to CL. 

But...., not sure if magna would like that. Definitely better, however, is changing the name to `forceAddToCl?: boolean`
So it's not like implied to be used only with Revert, it can work anywhere. (Makes more legible / maintainable code)

> @gc I note you labelled this as a suggestion but at present the creatables cl is uncompletable unless this gets merged. I guess the alternative would be to exclude bandosian components from the cl

PR #4168, opened 2022-07-12
[Leagues-BSO] Fixes "Unlock all Slayer rewards" bug for leagues

themrrobert:
### Description:

The enum has a dummy entry to take the place of the '0th' element, as well as includes non-unlockables, such as the herb sack and slayer ring. I probably should have named it `SlayerRewardsEnum` but oh well. 

### Changes:

- Pre-filters a list of just the unlockable Slayer rewards so this doesn't have to be computed every time.
- Updates the task to use the new list for the length check

### Other checks:

### DIFF
--- a/src/lib/slayer/slayerUnlocks.ts
+++ b/src/lib/slayer/slayerUnlocks.ts
@@ -16,15 +16,26 @@ export interface SlayerTaskUnlocks {
 }
 
 export enum SlayerTaskUnlocksEnum {
-	// Unlockables
-	MalevolentMasquerade = 2,
+	Dummy = 0,
+	// Not in use, but in theory gives 10% boost
+	GargoyleSmasher,
+	// Slayer helm unlock
+	MalevolentMasquerade,
+	// Create slayer rings
 	RingBling,
+	// Unlock Red Dragons (not in use)
 	SeeingRed,
+	// Unlock mith Dragons (not in use)
 	IHopeYouMithMe,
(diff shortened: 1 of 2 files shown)


gc:
```ts
export enum SlayerTaskUnlocksEnum {
	Dummy = 0,
```
disaster

themrrobert:
> ```ts
> export enum SlayerTaskUnlocksEnum {
> 	Dummy = 0,
> ```
> 
> disaster


Yea definitely wouldn't do that today, in fact, I can just pull that out and set GargoyleSmasher to 1, but hat's not even enabled either. And even if neither of those were there, the code still wouldn't work, because not all of those are unlocks, the Enum isn't named well either

src/lib/slayer/slayerUnlocks.ts
```diff
@@ -16,26 +16,15 @@ export interface SlayerTaskUnlocks {
...
 	IHopeYouMithMe,
-	// Unlock aviansies (not in use)
 	WatchTheBirdie,
-	// TzHaar unlock (not in use)
 	HotStuff,
-	// Lizardman unlock (not in use)
 	ReptileGotRipped,
-	// Unlock boss tasks. Definitely will use this one for the preroll.
 	LikeABoss,
-	// Unlock superiors
```
gc, line 38: these should be osb changes

gc (changes requested):

PR #4179, opened 2022-07-14
Update barbarianAssaultActivity.ts

Mraedis:
Fix missing space at end of Honour Level

### Other checks:


themrrobert (approved):
Typically I like to use pre-pended spaces on the later messages, however since there is always additional text added to this string, AND it follows the same behavior as the optional text in the following if() block, I think this is good.

PR #4191, opened 2022-07-16
Fixes problem where old item names map to custom items

themrrobert:
### Description:

Currently, with custom items, if they took the ID of a pre-existing item, the old item name still maps to the custom item.

**Examples:**
An indigo square => lil lamb
An indigo pentagon => doug
Abyssal demon => Ori
?? => Smokey

So when people are sacrificing osrs items that start with 'A' they inadvertantly sacrifice lil lamb + doug (actually happened)
Or someone evil could try to buy '???' for  100m [aka smokey] or trick someone into dropping/ sac'ing it since  they think it's a worthless UMB item.

### Changes:

- Ensures that the old name isn't being used by looking up the custom item name, and ensuring the new name is being used only.

### Other checks:

Tested: =b, =sellto, /sacrifice, /sell, /trade
(PS. There is a difference with the parseBankString between `0 Abyssal demon`, and `Abyssal Demon` that's why there's 2 checks.

### DIFF
--- a/src/lib/customItems/util.ts
+++ b/src/lib/customItems/util.ts
@@ -9,6 +9,7 @@ import getOSItem from '../util/getOSItem';
 export const customPrices: Record<number, number> = [];
 
 export const customItems: number[] = [];
+export const overwrittenItemNames: Map<string, Item> = new Map();
 
 export function isCustomItem(itemID: number) {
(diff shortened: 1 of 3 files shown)


gc (changes requested):
needs tests

Issue #4061, opened 2022-06-24
Add alert/message for clue scrolls back


6enny6:
Non patrons are no longer alerted they got a clue scroll in their loot, it used to say ‘You got a hard clue scroll etc’.

Assuming this was removed with the addition of patron only clue button

Some people use discord with images disabled and so are unaware they got a clue anymore.

Could message saying you received a clue be added back please

![2D2B8C12-F07A-44EF-A2F9-63CA0C1E3CD7](https://user-images.githubusercontent.com/76587427/175617109-87c1f630-a01e-412f-8c66-37b835691c40.jpeg)

PR #4203, opened 2022-07-19
Add clue message to trip return

TastyPumPum:
Adds a clue message to the loot return trip, this is useful when messages don't always load pictures at first. This got removed when the patron clue button got added.


6enny6:
closes #4061

gc (commented):
Should we only show this if they DONT have the button?

TastyPumPum:
> Should we only show this if they DONT have the button?

Have adjusted it for this, good idea.

PR #4225, opened 2022-07-21
Revenants death reduction - maxed combats

### LINKED ISSUE #3867: Have max combat stats give a small death reduction to Revenants
## 🚀 Feature Proposal

Basically what title says. In the 'bis' gear (in terms of Welfare, including Craws), you can only get your death chance up to 20%. Having a 1/5 chance to die every trip is a bit much, and **I think its reasonable to reduce this to 15% (aka flat 5% decrease) if your minion has max combat stats, including Prayer.** 

In-game, you do get pk'd quite often at Revenants but you don't die nearly as much as you would in the bot!

TastyPumPum:
If the user has maxed combats removes the 5% default chance from the calculations.

Closes #3867


gc:
What does this make the death chance now, if maxed?

PR #4231, opened 2022-07-21
Add Lil' Creator to server notifications

TastyPumPum:
I haven't been able to test this, as I'm not aware of a way to test it. But hopefully it looks good, would be a neat feature.


gc (changes requested):
Its always going to show them as receiving it on the last crate.

PR #4239, opened 2022-07-23
birdhouseButton

Lajnux:
### Description:

Added so that a birdhouse Button is shown on return trip if ready and the player is T1+

### Changes:

Added so that a birdhouse Button is shown on return trip if ready and the player is T1+

### Other checks:


gc:
FYI: waiting for this PR first https://github.com/oldschoolgg/oldschoolbot/pull/4269, unless you want to change this PR to point to that branch which would be faster/easier

PR #4261, opened 2022-07-27
DKS - disallowed attack styles

### LINKED ISSUE #2795: DK Attack Styles
Users should be forced to use/switch to correct attack style when killing DKs

698FFE5A-1127-422D-962D-DF064FFEE3CD

TastyPumPum:
Adds the disallowed attack styles to the dagannoth kings.

Closes #2795

### DIFF
--- a/src/lib/minions/data/killableMonsters/chaeldarMonsters.ts
+++ b/src/lib/minions/data/killableMonsters/chaeldarMonsters.ts
@@ -363,7 +363,7 @@ export const chaeldarMonsters: KillableMonster[] = [
 			prayer: 43
 		},
 		healAmountNeeded: 20 * 2,
-		disallowedAttackStyles: [SkillsEnum.Attack, SkillsEnum.Strength, SkillsEnum.Defence, SkillsEnum.Magic],
+		disallowedAttackStyles: [SkillsEnum.Attack, SkillsEnum.Strength, SkillsEnum.Magic],
 		attackStyleToUse: GearStat.AttackRanged,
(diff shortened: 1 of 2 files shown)


themrrobert:
What is the purpose of this? I've used defensive style on Prime, especially. It's a little less efficient in game, but that doesn't mean we shouldn't let people train it, imo.

6enny6:
@themrrobert see the issue #2795 

yes of course its possible to use mage on prime for example, but the kill would take long on osrs and that would not be reflected the same on osb, you would just receive the xp without reduced kill speed on osb. 

should be forced to use the style theyre weak to (DKs) unless we want to add a penalty for using a style theyre not weak to

themrrobert:
> @themrrobert see the issue #2795
> 
> yes of course its possible to use mage on prime for example, but the kill would take long on osrs and that would not be reflected the same on osb, you would just receive the xp without reduced kill speed on osb.
> 
> should be forced to use the style theyre weak to (DKs) unless we want to add a penalty for using a style theyre not weak to

If my DK trio PR was a thing this wouldn't even be an issue. Idk about you guys, but I can easily clear Prime with longrange before the next one spawns. The respawn timer is a bigger bottleneck at max efficiency, so I disagree here.

On the shamans, it definitely slows you down, but i still don't think that disabling defence is the right way to go about it.

And a negative boost needs to be well thought out, because, like in the DKs instance, you have enough time to finish the kill before the next one spawns, and the way it's designed is essentially like you're killing a room full of 3x Prime / whatever you're killing.

PR #4270, opened 2022-07-29
CL Custom (creatables) unobtainable black mask fix

TastyPumPum:
Currently the black mask is on the cl custom but unobtainable as it's autoconverted from the (10) version on the drops.

I also tried checking in this command still works if an old user had a black mask (10) in their bank but it wouldn't work for me. So deleting this command seems the best route.

Note - This won't fix the issue with black mask always being purple or not appearing on sac [slayer] log


themrrobert:
How does this fix the issue you've named about not being able to obtain it on the CL? The item would still be on the CL after this update and still be unobtainable right?

I have another PR that's quite old now that actually fixes the problem with the black mask on the CL.

Maybe just rename this PR because I don't think the PR is related to the title in anyway, unless something's changed massively with how that CL is generated?

TastyPumPum:
> How does this fix the issue you've named about not being able to obtain it on the CL? The item would still be on the CL after this update and still be unobtainable right?
> 
> I have another PR that's quite old now that actually fixes the problem with the black mask on the CL.
> 
> Maybe just rename this PR because I don't think the PR is related to the title in anyway, unless something's changed massively with how that CL is generated?

Well cl custom has creatables, and as its a creatable here with cl it's pulled to there, you could also add no cl to get the same effect but as I tried testing if the command still worked and it didn't I felt it was best to delete it

themrrobert:
Okay, just make it clear in the description / ideally title, too, that  it won't fix the issue with black mask always being purple or not appearing on sac [slayer] log

themrrobert (approved):

PR #4280, opened 2022-08-01
Add demonic offering spells to castables

TastyPumPum:
Enables you to gain prayer xp from the demonic ashes that currently have no use in the bot.

XP/hr seems slightly higher than what the wiki suggests, but the ticks are accurate to what the wiki says.


Lajnux:
The rates here is accurate to in game following same ticks, it would be great if this was added along with #4283 and #4281 since that would make the whole prayer skill be in a good state.

PR #4298, opened 2022-08-05
KG repeat trip on solo death

### LINKED ISSUE #4297: [BSO] Goldemar has no repeat trip button when you die/wipe
## 🐛 Bug Report

title. when your minion dies at goldemar, there is no repeat trip button.

TastyPumPum:
Adds the repeat trip to when your minion dies at KG.

Closes #4297 


gc (changes requested):
`${'You were'}`

PR #4306, opened 2022-08-07
DuelCommand GP deletion => Add new error type and log duelCommand to trace

themrrobert:
### Description:

The `/gamble duel` command ate some players money, and there's no obvious reason.

The duel was accepted, but before completion, another error was thrown (or the awaiter ended early, causing the try/catch the fall-through). This caused the duel to 'cancel' before completion, and both players lost their money.

### Changes:

- Added some logging to the duel failure so it logs errors that are NOT from timeout, as my testing strongly implies it was not a timeout error. [Process of elimination would tell us it's a timeout anyway if it happens again with no log]

### Other checks:

### DIFF
--- a/src/mahoji/lib/abstracted_commands/duelCommand.ts
+++ b/src/mahoji/lib/abstracted_commands/duelCommand.ts
@@ -80,7 +80,8 @@ export async function duelCommand(
 			user_id: String(duelSourceUser.id),
 			target_user_id: String(duelTargetUser.id),
 			wager: wagerAmount.toString(),
-			error_code: typeof err === 'object' && 'code' in err ? String(err.code) : 'NONE'
+			error_code: typeof err === 'object' && 'code' in err ? String(err.code) : 'NONE',
+			command: 'duel'
 		};
 		logError(err, extras);
 
--- a/src/mahoji/lib/abstracted_commands/duelCommand.ts
+++ b/src/mahoji/lib/abstracted_commands/duelCommand.ts
@@ -7,7 +7,7 @@ import { Bank, Util } from 'oldschooljs';
 import { Emoji, Events } from '../../../lib/constants';
 import { UserSettings } from '../../../lib/settings/types/UserSettings';
 import { channelIsSendable, updateGPTrackSetting } from '../../../lib/util';
-import { BotError, logError } from '../../../lib/util/logError';
+import { logError } from '../../../lib/util/logError';
 import { mahojiParseNumber } from '../../mahojiSettings';
 
 async function checkBal(user: KlasaUser, amount: number) {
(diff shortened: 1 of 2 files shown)


src/mahoji/lib/abstracted_commands/duelCommand.ts
```diff
@@ -151,8 +167,19 @@ export async function duelCommand(
...
+		if (!(failError instanceof Error)) {
+			failError = new Error(`Duel failed for unknown reason: ${String(err)}`);
+			failError.name = 'duelCommand Error';
+		}
+		return duelFail(failError);
 	}
-	return cancel();
+	const failError = new Error("duelCommand fell through try/catch this shouldn't happen");
+	failError.name = 'duelCommand Error';
+	return duelFail(failError);
```
gc, line 184: this seems really unnecessary. worst case just put it in a finally{}

gc (changes requested):
I dont get all the rigmarole. if it fails, just log the error, why all this other stuff?

src/mahoji/lib/abstracted_commands/duelCommand.ts
themrrobert: We don't want a finally because that will run every time and won't tell us anything.

Because the original code has `return cancel()` i can't rule out that that's where the error came from. (There is no guaranteed return in the try{} block.

src/mahoji/lib/abstracted_commands/duelCommand.ts
```diff
@@ -72,6 +73,21 @@ export async function duelCommand(
 		return "Duel cancelled, user didn't accept in time.";
 	}
 
+	function duelFail(err: any) {
+		duelMessage.delete().catch(noOp);
+		const wagerAmount = amount ?? 0;
+		const extras = {
```
themrrobert, line 79: This extra data is required so we can properly  track/refund a duel that fails, as that has already happened.

The function is required to avoid duplicate code

src/mahoji/lib/abstracted_commands/duelCommand.ts
```diff
@@ -151,8 +167,19 @@ export async function duelCommand(
 		if (selection.customID === 'CONFIRM') {
 			return await confirm(amount);
 		}
-	} catch (err) {
-		return cancel();
+	} catch (err: any) {
+		const code = typeof err === 'object' && 'code' in err ? err.code : null;
+		if (code === 'INTERACTION_COLLECTOR_ERROR') {
+			return cancel();
```
themrrobert, line 173: obviously we don't want to log an error if 'timeout' is the reason, but if it's another error/reason/message then we do want to log that.

PR #4309, opened 2022-08-08
Finish migrating +b features to /bank

themrrobert:
### Description:

These are all old features from `+b` that are not yet implemented on /bank

### Changes:

1. Adds bank string search `items` parameter. (Like spawn items vs item) This allows common searches like, `0 2424, 1242, 4342` or `cannonballs, dragon med helm, karil's leathertop`
2. Adds 2 new output formats: `text_page` (paginated view) and `text_full` which saves a file containing your entire bank. (These also work in combination with the `show_id` flag)
3. There is a new flag 'wide' which gives you the full wide bank image.
4. There is a 2nd flag option now, so you can combine 2 flags together, like wide + show_id
5. Sync klasaUsers bank before running.

### Other checks:

-  And the testers tested it


gc (changes requested):
`show_wide` doesnt make sense to me, just `wide`?

PR #4044, opened 2022-06-20
Charges Abyssal tentacle

Lajnux:
### Description:

Fixes so Abyssal tentacle gives charges when being created, resolves bug where peoples abyssal tentacle disappeared.

### Changes:

Fixes so Abyssal tentacle gives charges when being created, resolves bug where peoples abyssal tentacle disappeared.

### Other checks:

-   [X] I have tested all my changes thoroughly.

PR #4344, opened 2022-08-16
Fix abyssal tentacle wasting whips on create with onCreate() charges

themrrobert:
### Description:

When Abyssal tentacle was launched as a degradeable, because the 10k charges were the default, and there was no way to add charges with the \/create method, we limited the user's ability to create multiple tentacles, and we DONT revert it to a Kraken tentacle. 

This was changed and has caused several people to lose their whips when attempting to re-create their abyssal tentacle.

### Changes:

- Adds onCreate() optional function to Creatable interface
- Inserts code into create.ts command to call and process the results of onCreate()
- Adds the onCreate() hook to the Abyssal tentacle creatable

### Other checks:

### DIFF
--- a/src/lib/data/creatables/slayer.ts
+++ b/src/lib/data/creatables/slayer.ts
@@ -1,5 +1,7 @@
 import { Bank } from 'oldschooljs';
+import { toKMB } from 'oldschooljs/dist/util/util';
 
+import { mahojiUserSettingsUpdate } from '../../../mahoji/mahojiSettings';
 import { SlayerTaskUnlocksEnum } from '../../slayer/slayerUnlocks';
 import { Createable } from '../createables';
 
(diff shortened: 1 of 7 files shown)


Lajnux:
I would say the best short term fix would be to add #4044 instead. If a more centralised function is wanted after that can be resolved.

src/lib/data/creatables/slayer.ts
```diff
@@ -130,7 +132,16 @@ export const slayerCreatables: Createable[] = [
 		}),
 		outputItems: new Bank({ 'Abyssal tentacle': 1 }),
 		GPCost: 0,
-		maxCanOwn: 1
+		maxCanOwn: 1,
```
Lajnux, line 135:
```suggestion
		maxCanOwn: 1,
		
```
Can be removed with this PR right?

themrrobert: It technically could be removed, however then you get into a situation where there's multiple Abyssal tentacles, all sharing the same charge pool, which I don't think is the best idea considering they degrade into just the tentacle.

Lajnux: Yea I see what you mean. Then it looks good. 🥰🐳

PR #4363, opened 2022-08-19
sarachnis rebalance

osDreamscape:
### Description:

Add up to 20% in boosts with used gear ingame

before killrates were roughly 30/hr, this should put them around 46-47, ingame killrates has ironman EHB at 56/hr, so this is still markedly lower than feasible ingame rates but is a needed rebalance for a boss that is often killed for clues, and adds additional usage for scythe (which currently isn't used many places)

### Changes:

Added more boosts to sarachnis, 15% for scythe, 2% karil's skirt, 3% kari'ls top

### Other checks:


Lajnux:
so with all boosts applied that this PR gives average kill time of 1.012 minutes without weekend boost (59.3 kills/hour).
and 0.91 minutes kill time with weekend boost (66 kills / hour). which isn't inline with the description of the PR.

osDreamscape:
kept boosts, reverted base killtime to original 2.35, using values Fishy provided this brings kph with all boosts to 47.44, well below 56 kph 6hr ironman record

Lajnux:
Looks good to be merged.

themrrobert:
Out of curiosity why not bring it inline with ironman kph in osrs?

osDreamscape:
Without doing a thorough test myself ingame, the 6hr record for sarachnis is an outlier to the other values, so while it seems feasible, meeting in between the record and second place seemed more appropriate. I'm hesitant to propose balancing bosses to maximum possible efficiency as well, as I think a slight reduction to those values feels more realistic to expect.

PR #4387, opened 2022-08-27
Tame armor

gc:
### Other checks:


src/mahoji/commands/tames.ts
```diff
@@ -835,6 +884,17 @@ async function killCommand(user: KlasaUser, channelID: bigint, str: string) {
 		suffix: 'tame'
 	});
 
+	const deathChance = monster.deathChance ? monster.deathChance({ tame }) : 0;
+	const died = percentChance(deathChance);
+	let fakeDuration: number | undefined = undefined;
+	if (died) {
+		const deathKC = randInt(1, quantity);
```
themrrobert, line 891: I would definitely change this so a death doesn't wipe your entire trip, only the kills you failed. Otherwise it would be better to do a lower quantity and you're punished for having Patreon / Zak.

```ts
const deathChance = monster.deathChance ? monster.deathChance({ tame }) : 0;
let fakeDuration: number | undefined = undefined;
let deaths = 0;
for (let x = 0; x < quantity; x++) {
	if (percentChance(deathChance)) deaths++;
}
if (deaths) {
	// Fake duration is actually the real duration. note to self
	// Reduce quantity by however many deaths there were.
	fakeDuration = ((quantity - deaths) / quantity) * speed;
}
```


Then obviously, just pass the `quantity - deaths` loot for the quantity of the task, and give partial loot

src/lib/tames.ts
```diff
@@ -1,25 +1,182 @@
...
+	},
+	{
+		id: customDemiBosses.QueenBlackDragon.id,
+		minArmorTier: getOSItem('Justiciar igne armor')
+	},
+	{
+		id: customDemiBosses.SeaKraken.id,
+		minArmorTier: getOSItem('Justiciar igne armor')
+	},
+	{
```
themrrobert, line 146: You have Sea kraken here twice

src/lib/tames.ts
```diff
@@ -1,25 +1,182 @@
...
+	loot: (opts: { quantity: number; tame: Tame }) => Bank;
+	deathChance?: (opts: { tame: Tame }) => number;
+	oriWorks?: false;
+	mustBeAdult?: true;
+	minArmorTier?: Item;
+} & Omit<KillableMonster, 'table'>;
+
+function calcPointsForTame(tame: Tame) {
+	const lvl = tame.max_combat_level;
+	if (lvl < 75) return 10_000;
```
themrrobert, line 78: **This is just a suggestion, not an issue.**

While i'm not saying cater to ironmen, I do think they might be getting overly hit with this, as you can go 3, 4 eggs without getting 80+ if ur just a little unlucky.

I might scale it 15k => 24k, this is just a suggestion

themrrobert (changes requested):

PR #4399, opened 2022-08-30
Bingo

gc:
### Other checks:


themrrobert (approved):

src/mahoji/lib/abstracted_commands/volcanicMineCommand.ts
```diff
@@ -208,7 +208,7 @@ export async function volcanicMineShopCommand(
 	} else {
 		await transactItems({
 			userID: user.id,
-			collectionLog: true,
```
LuxrayElite, line 211: I understand this is to combat obtain 12k silver or being obtained however, this globally disables the CL from completion for 2 weeks (affecting non-bingo participants).

Volcanic mine teleport isn't a CL-only slot perhaps make something to handle this one item

PR #4434, opened 2022-09-06
Motherlode Mine

TastyPumPum:
Adds the ability to mine at the motherlode mine and obtain nuggets through it, removes the old way of getting golden nuggets.
If the user has 100 golden nuggets and 72 mining it will mine at the upper level, otherwise it will use the lower level. 


TastyPumPum:
This has been tested by testers and seems in line with osrs.

TastyPumPum:
Tested by myself since the conflicts and back working again.

TastyPumPum:
Working again after the latest conflitcts.

gc:
I dont think its an acceptable way to code it. You're breaking the contract of this array/the type/the name/etc. This is a list of ores, and you put something that isnt an ore in it. That, ontop of the fact its just a huge `else` branch in the mining activity - It should be its own activity.

src/lib/skilling/skills/mining.ts
```diff
@@ -250,6 +259,20 @@ const ores: Ore[] = [
 	}
 ];
 
+const MotherlodeMines: MotherlodeMine[] = [
```
gc, line 262: This makes no sense to me. Why is it an array?

gc (changes requested):

src/lib/skilling/skills/mining.ts
Lajnux: I think it's just a result of that we had 2 motherlode mine options in past, upper and lower which then got dismissed. but yes the [ ] should be removed.

src/lib/skilling/skills/mining.ts
```diff
@@ -250,6 +246,20 @@ const ores: Ore[] = [
...
+		id: 12_011,
+		name: 'Motherlode mine',
+		respawnTime: 5.5,
+		bankingTime: 60,
+		slope: 0.181,
+		intercept: 23.48,
+		petChance: 247_200
+	}
+];
+
```
gc, line 262: Why is this an array? I'm pretty sure I already commented on this in my last review

src/lib/types/minions.ts
```diff
@@ -92,6 +92,13 @@ export interface MiningActivityTaskOptions extends ActivityTaskOptions {
 	powermine: boolean;
 }
 
+export interface MotherlodeMiningActivityTaskOptions extends ActivityTaskOptions {
```
gc, line 95: What "oreID" does MLM have? its not an ore.

gc (changes requested):

src/lib/types/minions.ts
TastyPumPum: The "oreID" of Gem Rock is mapped to "uncut opal" so i didnt think we was using oreID in a strict manner. I used the ID of pay-dirt though as I assumed it was the same concept.

src/lib/skilling/skills/mining.ts
TastyPumPum: Not sure on how you feel it would be best to have the details of this then, it made sense to me to follow a similar suit as other mine-ables.

src/lib/types/minions.ts
gc: There's no reason at all to do this. Even if there was, its not an acceptable solution. MLM is not an ore, make it its' own thing.

src/lib/skilling/skills/mining.ts
gc: Remove the array, and make it just an object. MLM is only one thing, theres no purpose in making it an array. You dont have to try make it the same as ores.

gc (changes requested):

src/mahoji/commands/mine.ts
```diff
@@ -248,19 +253,27 @@ export const mineCommand: OSBMahojiCommand = {
 		userID,
 		channelID
 	}: CommandRunOptions<{ name: string; quantity?: number; powermine?: boolean }>) => {
+		const user = await mUserFetch(userID);
+		let { quantity, powermine } = options;
+
+		const MotherlodeMine = Mining.MotherlodeMine.name === options.name;
+
+		if (MotherlodeMine) {
+			return motherlodeMineCommand({ user, channelID, quantity, powermine });
```
gc, line 262: Powermining at MLM?

gc (changes requested):

src/mahoji/commands/mine.ts
```diff
@@ -248,19 +253,27 @@ export const mineCommand: OSBMahojiCommand = {
 		userID,
 		channelID
 	}: CommandRunOptions<{ name: string; quantity?: number; powermine?: boolean }>) => {
+		const user = await mUserFetch(userID);
+		let { quantity, powermine } = options;
+
+		const MotherlodeMine = Mining.MotherlodeMine.name === options.name;
```
gc, line 259: camelcase.

gc (changes requested):

src/mahoji/lib/abstracted_commands/motherlodeMineCommand.ts
```diff
@@ -0,0 +1,104 @@
...
+		ticksBetweenRolls: currentPickaxe.ticksBetweenRolls,
+		glovesRate,
+		armourEffect,
+		miningCapeEffect,
+		powermining: powermine,
+		goldSilverBoost,
+		miningLvl: miningLevel
+	});
+
+	const duration = timeToMine;
```
gc, line 77: this is useless - just name it duration in the assignment.

src/mahoji/lib/abstracted_commands/motherlodeMineCommand.ts
```diff
@@ -0,0 +1,104 @@
...
+		boosts.push(
+			'\nYou are mining on the upper level of the motherlode mine, due to having 100 golden nuggets in your cl and 72 mining or higher'
+		);
+	}
+
+	const glovesRate = 0;
+	const armourEffect = 0;
+	const miningCapeEffect = 0;
+	const goldSilverBoost = false;
+	const powermine = false;
```
gc, line 61: cape/armor/gloves/etc doesnt effect it all?

gc (changes requested):

src/mahoji/lib/abstracted_commands/motherlodeMineCommand.ts
TastyPumPum: Correct (source - https://oldschool.runescape.wiki/w/Mining_enhancers#Exceptions)

src/lib/skilling/functions/miningBoosts.ts
TastyPumPum, line 1: Just want to check that this seems an appropriate location and name for the file, had to move them due to circular dependancies

PR #4493, opened 2022-09-24
Fix thieving pet calculation

themrrobert:
### Description:

This can lead to guaranteed pet drop [on bso]  or more likely than it should be on OSB

### Changes:

- Roll pet chance per success instead rolling (petChance/successCount)

### Other checks:


Lajnux:
this would result in stall no longer dropping pet

Lajnux:
> this would result in stall no longer dropping pet

looks good to be merged now

PR #4495, opened 2022-09-24
Change BSO skilling pet multiplier to 5b

themrrobert:
### Description:

As we discussed, this changes the 15x increase of pet rate requirement to 5b XP for BSO

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -771,7 +771,7 @@ export function skillingPetDropRate(
 	skill: SkillsEnum,
 	baseDropRate: number
 ): { petDropRate: number } {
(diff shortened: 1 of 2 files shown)


Lajnux:
this doesn't resolve the core issue which is the pets being posted in notifications for BSO. If someone at 5B wanna keep spamming the notifications they still can.

themrrobert:
> this doesn't resolve the core issue which is the pets being posted in notifications for BSO. If someone at 5B wanna keep spamming the notifications they still can.

That is definitely not the core issue.  As far as Im concerned that's a totally separate issue.

PR #4502, opened 2022-09-26
adjust expire time shootingstar

Lajnux:
### Description:

Adjusted the expire length slightly and made it also scale for patron tiers to be more reasonable. 

### Changes:

Adjusted the expire length slightly and made it also scale for patron tiers to be more reasonable. 

### Other checks:

### DIFF
--- a/src/lib/util/calcMaxTripLength.ts
+++ b/src/lib/util/calcMaxTripLength.ts
@@ -5,7 +5,7 @@ import { PerkTier } from '../constants';
 import { SkillsEnum } from '../skilling/types';
 import { skillLevel } from './minionUtils';
 
-function patronMaxTripBonus(user: MUser) {
+export function patronMaxTripBonus(user: MUser) {
 	const { perkTier } = user;
 	if (perkTier === PerkTier.Two) return Time.Minute * 3;
 	else if (perkTier === PerkTier.Three) return Time.Minute * 6;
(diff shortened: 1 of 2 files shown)


themrrobert:
Since shooting stars are AFK, why not make them 2x max trip length like other 'AFK' stuff such as alching + cannonballs?

Lajnux:
> Since shooting stars are AFK, why not make them 2x max trip length like other 'AFK' stuff such as alching + cannonballs?

I didn't think about that tbh, and it's doable but would be slight spaghetti code in the determineMineFunction

themrrobert:
> > Since shooting stars are AFK, why not make them 2x max trip length like other 'AFK' stuff such as alching + cannonballs?
> 
> I didn't think about that tbh, and it's doable but would be slight spaghetti code in the determineMineFunction

I would just take whatever `determineMineFunction` returns and multiply it by 2.

However idk if people would even want to spend that much time at a shooting star 🤷

Lajnux:
> > > Since shooting stars are AFK, why not make them 2x max trip length like other 'AFK' stuff such as alching + cannonballs?
> > 
> > 
> > I didn't think about that tbh, and it's doable but would be slight spaghetti code in the determineMineFunction
> 
> I would just take whatever `determineMineFunction` returns and multiply it by 2.
> 
> However idk if people would even want to spend that much time at a shooting star 🤷

Yea then we would have to double the output, but it wouldn't really be accurate doing that since shooting stars have layers and then the outcome would kinda screw with the whole layer simulation. I think it's fine as is but the expire length def needs this slight length increase.

themrrobert:
I approve these changes as-is

PR #4504, opened 2022-09-27
adds "mysterious" alias to mysterious tree farming

mylife212:
### Description:
What title says, it accepts aliases for "mysterious tree" and "mystery", but not mysterious.

### Other checks:


themrrobert:
Would you mind adding a few more aliases? specifically the first word of each multiple-word plant ? ❤️ 

(except potato cactus because potato)

PR #4506, opened 2022-09-27
Giant's Foundry

Lajnux:
### Description:

Giants' Foundry minigame along with smithing set and effects.

### Changes:

Giants' Foundry minigame along with smithing set and effects.

### Other checks:


src/lib/types/minions.ts
```diff
@@ -316,6 +316,12 @@ export interface PuroPuroActivityTaskOptions extends MinigameActivityTaskOptions
 	darkLure: boolean;
 }
 
+export interface GiantsFoundryActivityTaskOptions extends MinigameActivityTaskOptions {
+	alloyName: string;
```
gc, line 320: use IDs not names, always in task options.

src/mahoji/commands/minigames.ts
```diff
@@ -808,6 +810,63 @@ export const minigamesCommand: OSBMahojiCommand = {
...
+			options: [
+				{
+					type: ApplicationCommandOptionType.Subcommand,
+					name: 'start',
+					description: 'Start a trip.',
+					options: [
+						{
+							type: ApplicationCommandOptionType.String,
+							name: 'name',
+							description: 'The alloy/metal you wanna use.',
```
gc, line 827: use proper english.

src/mahoji/lib/abstracted_commands/giantsFoundryCommand.ts
```diff
@@ -0,0 +1,272 @@
...
+		},
+		metalScore: 20,
+		sections: 4
+	},
+	{
+		name: 'Steel',
+		level: 30,
+		cost: {
+			[itemID('Steel bar')]: 28
+		},
```
gc, line 39: never use the `{ [itemID]: }` format. always use `new Bank()`

src/mahoji/lib/abstracted_commands/giantsFoundryCommand.ts
```diff
@@ -0,0 +1,272 @@
...
+		metalScore: 130,
+		sections: 7
+	}
+];
+
+export const giantsFoundryBuyables: { name: string; output: ItemBank; cost: number; aliases: string[] }[] = [
+	{
+		name: 'Double ammo mould',
+		output: resolveNameBank({ 'Double ammo mould': 1 }),
```
gc, line 98: change these to `new Bank`.

gc (dismissed):
Seems mostly good, although i dont know anything about the actual minigame. and assuming its been tested.

src/tasks/minions/minigames/giantsFoundryActivity.ts
```diff
@@ -0,0 +1,141 @@
...
+		for (let i = 0; i < quantity; i++) {
+			let quality = Math.min(Math.floor(randomVariation(metalScore - 5 + avgMouldBonus, 10)), 199);
+			xpReceived += (Math.pow(quality, 2) / 73 + 1.5 * quality + 1) * 30;
+			reputationReceived += quality;
+
+			// Increse and save down Giant Weapons Made
+			const tipID = randInt(1, tipMoulds.length);
+			const bladeMouldID = randInt(1, bladeMoulds.length);
+			const forteMouldID = randInt(1, forteMoulds.length);
+			const weaponID = tipID.toString() + bladeMouldID.toString() + forteMouldID.toString();
```
themrrobert, line 85: This won't work.

How will you differentialte 1, 11, 3 from 11, 1, 3.

you should use the code i gave you to combine them into a single number

src/tasks/minions/minigames/giantsFoundryActivity.ts
```diff
@@ -0,0 +1,141 @@
...
+		let reputationReceived = 0;
+		let xpReceived = 0;
+		let weaponName = '';
+		for (let i = 0; i < quantity; i++) {
+			let quality = Math.min(Math.floor(randomVariation(metalScore - 5 + avgMouldBonus, 10)), 199);
+			xpReceived += (Math.pow(quality, 2) / 73 + 1.5 * quality + 1) * 30;
+			reputationReceived += quality;
+
+			// Increse and save down Giant Weapons Made
+			const tipID = randInt(1, tipMoulds.length);
```
themrrobert, line 82: should be doing from 0 - (length -1) since indexes start at 0.

src/tasks/minions/minigames/giantsFoundryActivity.ts
```diff
@@ -0,0 +1,141 @@
...
+			const bladeMouldID = randInt(1, bladeMoulds.length);
+			const forteMouldID = randInt(1, forteMoulds.length);
+			const weaponID = tipID.toString() + bladeMouldID.toString() + forteMouldID.toString();
+			weaponName = `${tipMoulds[tipID - 1]} ${bladeMoulds[bladeMouldID - 1]} ${forteMoulds[forteMouldID - 1]}`;
+			await user.incrementGiantsWeaponsMade(parseInt(weaponID));
```
themrrobert, line 87: This means you are updating the DB for every single sword, instead of once for the entire lot...

Do it like i pasted in #developers, where it grabs + then updates the ItemBank, and only after you finish the quantity loop, do you then update the weapons_made option.

src/tasks/minions/minigames/giantsFoundryActivity.ts
```diff
@@ -0,0 +1,141 @@
...
+		const xpRes = await user.addXP({
+			skillName: SkillsEnum.Smithing,
+			amount: xpReceived,
+			duration
+		});
+
+		const currentUserReputation = user.user.foundry_reputation;
+
+		await user.update({
+			foundry_reputation: currentUserReputation + reputationReceived
```
themrrobert, line 101: this is where you should also be updating the weapons_made value at the same time to avoid multiple db writes.

themrrobert (changes requested):

src/tasks/minions/minigames/giantsFoundryActivity.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export const giantsFoundryTask: MinionTask = {
+	type: 'GiantsFoundry',
+	async run(data: GiantsFoundryActivityTaskOptions) {
+		const { quantity, userID, channelID, duration, metalScore, alloyID } = data;
+		const user = await mUserFetch(userID);
+		const alloy = giantsFoundryAlloys.find(i => i.id === alloyID);
+		if (!alloy) {
+			return 'An issue occurred trying to find a alloy using the alloy ID.';
+		}
```
gc, line 27: you can null assert in cases like this. remove this `if`, and use `!` to say its always there. we want it to error if its not there

gc (changes requested):

src/lib/giantsFoundry.ts
```diff
@@ -0,0 +1,75 @@
...
+];
+
+export const TOTAL_GIANT_WEAPONS = tipMoulds.length * bladeMoulds.length * forteMoulds.length;
+
+// weaponID stored as hex 0a0403 => 10, 4, 3
+export function decodeGiantWeapons(weaponID: string) {
+	const decimal = parseInt(weaponID, 16);
+	const [tipMould, bladeMould, forteMould] = [decimal >> 16, (decimal >> 8) & 255, decimal & 255];
+	return [tipMould, bladeMould, forteMould];
+}
```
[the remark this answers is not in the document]
Lajnux: I mostly wanted to learn bitwise and hex conversion to improve my coding skill and with Cyrs help understanding it I really like the way it's stored, it makes the length of the keys same size always rather than being of different lengths with like ''11-0-2'' and ''10-10-10''.

I can still change back to that way if you want, I tested it out and it works great and is fast.

src/mahoji/lib/abstracted_commands/giantsFoundryCommand.ts
```diff
@@ -0,0 +1,293 @@
...
+		type: 'Minigame',
+		totalCost,
+		changeType: 'cost',
+		users: [
+			{
+				id: user.id,
+				cost: totalCost
+			}
+		]
+	});
```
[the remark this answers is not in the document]
Lajnux: tob_cost and tob_loot is currently in the user table and not userstats table, you want this moved aswell?

src/lib/giantsFoundry.ts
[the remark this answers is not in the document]
gc: I'd rather it changed back. It's good you learned and that it makes things the same length, but besides that, its little benefit for a much more complex way of doing it.

src/lib/giantsFoundry.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function generateRandomGiantWeapon(): [number, number, number] {
+	return [randInt(0, tipMoulds.length - 1), randInt(0, bladeMoulds.length - 1), randInt(0, forteMoulds.length - 1)];
+}
+
+export function giantWeaponName(weapon: string | number[]) {
+	if (typeof weapon !== 'string' && weapon.length < 3) {
+		logError(new Error('Invalid weapon data'));
+		return '';
+	}
```
gc, line 72: Are you sure you want to return an empty string here? It would probably be better to throw the error.

gc (changes requested):

src/lib/giantsFoundry.ts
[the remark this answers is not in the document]
Lajnux: changed back

src/lib/giantsFoundry.ts
Lajnux: Yes because it's in the activity, if we throw a error (which shouldn't happen) we would make that trip of the user wasted rather than just a string being incorrect.

PR #4517, opened 2022-09-29
Added POH Items

### LINKED ISSUE #4382: return the =poh items command into /.
## 🚀 Feature Proposal

there was this command which was =poh items which listed all items you could build and to which slot of the house,this was lost when =poh turned into /poh . the wiki for some reason states that there is /poh items despite its not a thing, only thing you can do currently is /poh build and then randomly search for thing. the link below takes to the discord oldschoolgg to the spot where picture below was taken at. https://discord.com/channels/342983479501389826/792691343284764693/979416160598634557
image

ghosts-are-spooky:
### Description:

Added command that was there previously to list the buildable items in PoH

### Changes:

Added subcommand on PoH, it lists out in the same format as before, but in text output format due to character limit in discord.

### Other checks:

Closes #4382

### DIFF
--- a/src/mahoji/commands/poh.ts
+++ b/src/mahoji/commands/poh.ts
@@ -103,6 +103,19 @@ export const pohCommand: OSBMahojiCommand = {
 					required: true
 				}
 			]
+		},
+		{
+			type: ApplicationCommandOptionType.Subcommand,
+			name: 'items',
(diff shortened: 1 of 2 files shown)


Lajnux:
make this a PR for master rather than BSO since it's a feature that both bots can utilize.

PR #4523, opened 2022-09-29
Fix Ironman self cracker

ghosts-are-spooky:
### Description:

Fixed iron accounts not being able to pull a cracker with themselves

### Changes:

Modified the self check on iron to use minion.id instead of minion. This is the same as for non-irons.

### Other checks:


themrrobert (approved):
Needed fix, looks good.

PR #4528, opened 2022-09-30
Disable repeating clue trips

TastyPumPum:


themrrobert:
This should only really be on OSB, so once this is merged, can you put it back on BSO?

Ty!

PR #4542, opened 2022-10-02
Improve efficiency of hanldeTripFinish

themrrobert:
### Description:

Improves efficiency by only making the PerkTier enabled database queries if the user actually has the required PerkTier.

A small efficiency, yes, but these add up.

### Changes:

- Places birdhouse and slayer task db queries inside a perkTier `if` wrapper.

### Other checks:


src/lib/util/handleTripFinish.ts
```diff
@@ -112,15 +112,13 @@ export async function handleTripFinish(
 	if (clueReceived && perkTier > PerkTier.One) components.addComponents(makeDoClueButton(clueReceived));
```
LuxrayElite, line 112: couldnt you also wrap this bit into it 👀 removes the additional perk check

keeping all patron features wrapped together

themrrobert: Moving those awaits/database calls into an `if` block save several milliseconds while the one you're referring to saves microseconds, roughly 1000x less.

I try to change the minimum amount required for several reasons, like getting it merged quicker, and avoiding code bloat in other cases.

PR #4549, opened 2022-10-04
Fix Woodcutting bugs related to Sulliusceps

themrrobert:
### Description:

Sulliusceps currently only drop mushrooms, instead of their usual drops.
Also, Sulliusceps / Redwoods are dropping regular bird nests when they should only be dropping clue nests.

### Changes:

- Adds LootTable support to Wc Logs
- Adds the right drops to the LootTable for Sullisceps.
- Removes regular bird nests from Redwoods / Sulliusceps, but keeps clue nests.
- Converts Log.id to use itemID('Log name'). -- **Verified that none of the IDs changed**

### Other checks:

-   [x] I have tested all my changes thoroughly. Also tested by testers.


src/lib/skilling/skills/woodcutting.ts
```diff
@@ -1,12 +1,24 @@
+import { LootTable } from 'oldschooljs';
+
 import { Emoji } from '../../constants';
 import itemID from '../../util/itemID';
 import { Log, SkillsEnum } from '../types';
 
+const sulliuscepTable = new LootTable()
+	.add('Numulite', [4, 8], 34)
```
Lajnux, line 8:
```suggestion
	.add('Numulite', [1, 60], 34)
```
The loottable doesn't seem to match the wiki loot table. 
Screenshot_20221005-043745_Chrome

themrrobert: I'm not where the wiki got those numbers, but it's 100% NOT a roll from 1-60.

If I put that into the LootTable, it would be like 5x too much. Maybe there is a 1 in 1000 chance it gives 60, but I personally spent several hours doing sulliuceps myself to get a good idea of how many numulite you actually get, as well as identifying whether or not you numulite/fossils are tertiary or standard drops (standard).

PR #4556, opened 2022-10-06
make repeat trip always possible (same as for farming trips)

Lajnux:
### Description:

make repeat trip always possible (same as for farming trips)

### Changes:

make repeat trip always possible (same as for farming trips)

### Other checks:


gc:
Why?

Issue #4553, opened 2022-10-05
Issue with getting bonus slayer points when doing Inferno 


gpRollMaster:
prior slayer task info
You've completed **879** tasks and received **15** points; giving you a total of **1740;** return to a Slayer master.

After completion of inferno info
You've completed **880** tasks and received **15** points; giving you a total of **1755;** return to a Slayer master.

Loot: 32,880x Tokkul, 1x Infernal cape
XP: +80,000 :ranged:+35,000 :hitpoints:+25,000 :magic:You received 125,000 :slayer: XP (76k/Hr)
You made it through 100.00% of the Inferno.

Didnt miss out on a lot of points but would be nice to get this addressed as if it was the 50 tasks in a row and above its a lot more that would be lost.

gpRollMaster:
![inferno slayer](https://user-images.githubusercontent.com/97349844/194128080-258abf70-a27e-4e1b-a273-02316918df79.png)

ard35:
Fixed on next update

PR #4557, opened 2022-10-06
Fixes Slayer Task Bugs for Inferno and Fight Caves

ard35:
### Description:

- Resolves Issue #4553.
- Resolves a bug where failing an Inferno run silently completes the slayer task and awards slayer points.

### Changes:

- Calculate current slayer task streak before calculating points to award (like monsterActivity does it).
- Only complete an Inferno task is the user does not die.
- Add a Slayer Task Cancelled message when the user dies in Inferno and is on task.
- Fix message in Fight Caves to be more specific (Task Cancelled -> Slayer Task Cancelled). 

### Other checks:

### DIFF
--- a/src/tasks/minions/minigames/infernoActivity.ts
+++ b/src/tasks/minions/minigames/infernoActivity.ts
@@ -93,6 +93,8 @@ export const infernoTask: MinionTask = {
 				}
 			}
 			if (isOnTask) {
+				text += '**Slayer task cancelled.**\n';
+
 				await prisma.slayerTask.update({
 					where: {
 						id: usersTask.currentTask!.id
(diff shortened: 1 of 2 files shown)


src/tasks/minions/minigames/infernoActivity.ts
```diff
@@ -105,28 +107,32 @@ export const infernoTask: MinionTask = {
...
 				},
 				data: {
 					quantity_remaining: 0,
-					skipped: deathTime ? true : false
+					skipped: false
 				}
 			});
 
-			text += `\n\n**You've completed ${newUser.slayer_task_streak} tasks and received ${points} points; giving you a total of ${newUser.slayer_points}; return to a Slayer master.**`;
+			text += `\n\n**You've completed ${currentStreak} tasks and received ${points} points; giving you a total of ${secondNewUser.newUser.slayer_points}; return to a Slayer master.**`;
```
Lajnux, line 135:
```suggestion
			text += `\n\n**You've completed ${currentStreak} tasks and received ${points} points; giving you a total of ${secondNewUser.slayer_points}; return to a Slayer master.**`;
```
This doesn't look right, same in fightCavesActivity

ard35: The const `newUser` is already deconstructed from the object returned from updating the slayer task streaks. `secondNewUser` is the actually object returned from updating the points which contains the newUser property which is why I have to call it here. This is the same way that `monsterActivity.tsx` does it.

Issue #4543, opened 2022-10-03
Hunter success chance bso 


DarkBugz:
The level scaling for successful hunting caps at 99 instead of 120 for bso.

gc:
This should go in #bso-vote, unless it was previously like this and therefore might be a bug.

DarkBugz:
Fishy said when he originally coded hunter he put no such level cap so it might be new with slash commands.

Lajnux:
> Fishy said when he originally coded hunter he put no such level cap so it might be new with slash commands.

I only said I don't remember putting such a change but then I checked back now on the old PR and I did actually add it so that hunter potion wouldn't break.

PR #4564, opened 2022-10-09
Change hunter scaling to max level for BSO

ghosts-are-spooky:
### Description:

Fixed 99-120 scaling for hunter. Fixes https://github.com/oldschoolgg/oldschoolbot/issues/4543

### Changes:

Changed the calculate hunter success from a max value of 99 to use the constant MAX_LEVEL, this way it will stay 99 for OSB, but increase to 120 for BSO.

This affects BSO only and primarily with black chinchompa.

BSO Catch rates here:
Screenshot 2022-10-09 at 12 24 30

### Other checks:


gc:
This isn't really a bug, labelling as a suggestion.

Lajnux (approved):
I simulated this change for BSO and it looks as a okay change. Checked all creatures for BSO nothing breaks and it dosen't affect the BSO creatures very much since most of them are already 100% catch rate.

PR #4565, opened 2022-10-09
Feature/docs command

ghosts-are-spooky:
Description:
Added /docs command in order to search for wiki articles.
Added /admin sync_docs subcommand.

Changes:
Prisma: 
Schema updated to create table WikiDocs, holds Id, name and path. This was to not need to query gitbook api every request, and to be able to validate input query so no broken pages would be served.

Admin Sync Docs:
Added this for any wiki contributor, mod, admin etc. Needs to be run only when Wiki is updated.

Docs Command:
SQL Select to find matching articles using query string and wikidoc name.
Run checks the query value against all docs to match a url path of a valid article, if found sends the related article.

Docs Activity:
Sync Docs - Uses gitbook api to pull pages and their sections (page anchors), loops through pages until all articles retrieved. Then does an upsert to db to update pages + add any new.

GetDocsResults - Case insensitive sql query to find matching articles using query value against page name, strip out " - " from both name/query as this was added to split sections from a page more clearly in search results.

GetAllDocsResults - Selects all articles to use for validation on command run. 

Todo:
Give config details / setup api key.
Test new admin subcommand doesn't interfere with any permissions on others.

Config requires additional values:
GITBOOK_TOKEN - API Key from: https://app.gitbook.com/account/developer
GITBOOK_API_URL - Static (can move directly into url calls if needed)
GITBOOK_SPACE_ID - Different for OSB/BSO
GITBOOK_URL - Different for OSB/BSO

Other checks:
[x] I have tested all my changes thoroughly.


ghosts-are-spooky:
failed test due to missing config const

ghosts-are-spooky:
Ready for code review

ghosts-are-spooky:
Ready for review - typescript test failing due to missing config files

PR #4578, opened 2022-10-11
Update max mage stats with void staff

### LINKED ISSUE #4577: Void Staff has higher stats but isn't green
## 🐛 Bug Report

doesn't make sense to me why equipping void staff which took virtus wand to make wouldn't be a green number like when virtus wand is equipped, especially when void has a higher number.

TastyPumPum:
Closes #4577 

Swaps the max mage stats to 509 which is achieved with the void staff rather than the virtus wand.

195012568-768aa1cf-23d8-4653-aafd-ca333d131e7d


ghosts-are-spooky:
Spellbound ring (I) is +13 over seers I, abyssal tome is +10 over virtus book, so max should be 532

Issue #3383, opened 2022-01-02
Finish Tob/Cox


6enny6:
Add
+finish tob
+finish cox

now they are both on the minions people want to do simulations of both, they used to be on here not sure what happened 👍🏻

TURB042O:
/simulate cox is a thing
not tob or anything else raid wise unfortunately.

LuxrayElite:
> /simulate cox is a thing not tob or anything else raid wise unfortunately.

I'm waiting for @ard35 to resolve the tob issues on bso otherwise this issue would be closed as both suggested features are implemented and working

Issue #3142, opened 2021-11-13
Phosani Simulation


6enny6:
add `+kill x phosani` (sim) and `+finish phosani`

Issue #2033, opened 2021-06-10
+finish cox cm


ghost:
I like to use '+finish zulrah' to see uniques and when I might expect them myself. It's really motivating. I and others I have spoken to would like to be able to '+finish cox cm'. I wrote it in powershell below. The logic is that you need 3 twisted kits for each unique ancestral piece and only one dust for the pet. Anything last 1 dust and 3 kits is excess duplicates.

```
$kc = 0
$kit = 0
$dust = 0

while (($kit -lt 3) -or ($dust -lt 1)) {
$kc += 1

if ((get-random 75) -eq (get-random 75) -and ($kit -lt 3)) {
$kit += 1
Write-Host $kc 'kit'
}

if ((get-random 400) -eq (get-random 400) -and ($dust -lt 1)) {
$dust += 1
Write-Host $kc 'dust'
}

}
```

At some point we will find out the tob hm drop rates and adding a '+finish tob hm' would be just as simple. Thank you and please add :)

PR #4603, opened 2022-10-17
Add New Monsters / Activities and Aliases to the Finish Command

ard35:
### Description:

Adds the ability to use aliases in the finish command. Also adds more options for simulation using this command (COX CM, TOB, TOB HM, Nightmare, Phosani, Gauntlet, Corrupted Gauntlet, Nex)

Closes issues: #3383, partially #3142, #2033

### Changes:

- Pass monster and activity aliases through to command options
- Allow the above following monsters / activities to be simulated

### Other checks:

### DIFF
--- a/src/mahoji/commands/finish.ts
+++ b/src/mahoji/commands/finish.ts
@@ -32,6 +32,7 @@ export const finishCommand: OSBMahojiCommand = {
 		const kcBank = new Bank();
 		let kc = 0;
 		const maxAttempts = val.maxAttempts ?? 100_000;
+		let unlockedSpiritOutfit = false;
 		for (let i = 0; i < maxAttempts; i++) {
 			if (val.cl.every(id => loot.has(id))) break;
 			kc++;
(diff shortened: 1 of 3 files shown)


src/mahoji/commands/finish.ts
```diff
@@ -36,6 +36,32 @@ export const finishCommand: OSBMahojiCommand = {
...
+			if (val.name.includes('Theatre of Blood')) {
+				if (kc === 100) thisLoot.add('Sinhaza shroud tier 1', 1);
+				if (kc === 500) thisLoot.add('Sinhaza shroud tier 2', 1);
+				if (kc === 1000) thisLoot.add('Sinhaza shroud tier 3', 1);
+				if (kc === 1500) thisLoot.add('Sinhaza shroud tier 4', 1);
+				if (kc === 2000) thisLoot.add('Sinhaza shroud tier 5', 1);
+			}
+			if (val.name === 'Corrupted Gauntlet') thisLoot.add('Gauntlet cape', 1);
+			if (val.name === 'Tempoross') {
+				if (loot.amount('Spirit flakes') >= 4800) {
```
ard35, line 57: Actually going to rework this slightly and pull these out of the loop so it's not being added on every kill after 4800 flakes has been reached.

gc:
Please make sure to test each one, check if any take too long to finish (it should be practically instant).

ard35:
> Please make sure to test each one, check if any take too long to finish (it should be practically instant).

Yep, ran tests on them all and they finish nearly instantly.

src/mahoji/commands/finish.ts
```diff
@@ -26,16 +26,42 @@ export const finishCommand: OSBMahojiCommand = {
...
+				if (kc === 1500) thisLoot.add("Xeric's general", 1);
+				if (kc === 2000) thisLoot.add("Xeric's champion", 1);
+			}
+			if (val.name.includes('Theatre of Blood')) {
+				if (kc === 100) thisLoot.add('Sinhaza shroud tier 1', 1);
+				if (kc === 500) thisLoot.add('Sinhaza shroud tier 2', 1);
+				if (kc === 1000) thisLoot.add('Sinhaza shroud tier 3', 1);
+				if (kc === 1500) thisLoot.add('Sinhaza shroud tier 4', 1);
+				if (kc === 2000) thisLoot.add('Sinhaza shroud tier 5', 1);
+			}
```
gc, line 55: these should be extracted into the `Finishable` objects.
e.g.

```
tertiaryDrops: ({kc}) => {
if (kc === 1) do something;
}
```

src/mahoji/commands/finish.ts
```diff
@@ -26,16 +26,42 @@ export const finishCommand: OSBMahojiCommand = {
 	],
 	run: async ({ interaction, options }: CommandRunOptions<{ input: string }>) => {
 		await deferInteraction(interaction);
-		const val = finishables.find(i => stringMatches(i.name, options.input));
+		const val = finishables.find(i => stringMatches(i.name, options.input) || i.aliases?.includes(options.input));
```
gc, line 29: we always check `stringMatches` on aliases, not direct string match. aliases arent included in the first place because theres autocomplete and so not a big need for them, so maybe dont even need to add them.

gc (changes requested):

src/mahoji/commands/finish.ts
ard35: ah right, I fixed the check. A lot of people were requesting aliases to work for this command, I don't think it's bad thing to have them either

PR #4604, opened 2022-10-17
Fix DoubleLoot trigger for premium time patrons.

themrrobert:
### Description:

Currently. `/tools patron doubleloot` doesn't work for premium time patrons.
Often these are fully paid patron memberships that are gifted from others.

Also we should be consistent in treating premium time patrons the same as bitfield patrons, and a perkTier check is more consistent now than using bitfields.

### Changes:

- Replaces the bitfield check with getUsersPerkTier
- Supporting code updated

### Other checks:

### DIFF
--- a/src/mahoji/commands/tools.ts
+++ b/src/mahoji/commands/tools.ts
@@ -566,11 +566,11 @@ async function checkMassesCommand(guildID: string | undefined) {
 ${massStr}`;
 }
 
-function calcTime(perkTier: number) {
+function calcTime(perkTier: PerkTier) {
 	for (const [bit, dur] of [
-		[7, Time.Minute * 90],
-		[6, Time.Minute * 40],
-		[5, Time.Minute * 20]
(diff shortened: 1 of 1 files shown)


src/mahoji/commands/tools.ts
```diff
@@ -566,21 +566,20 @@ async function checkMassesCommand(guildID: string | undefined) {
...
-	throw new Error('User had no 4/5/6 bit');
+	throw new Error('User is not a Tier 4+ Patron');
 }
 
 async function patronTriggerDoubleLoot(user: MUser) {
-	if (
-		![BitField.IsPatronTier4, BitField.IsPatronTier5, BitField.IsPatronTier6].some(i => user.bitfield.includes(i))
-	) {
+	const perkTier = getUsersPerkTier(user);
+	if (perkTier < 5) {
```
gc, line 582: Use `PerkTier` enum instead of the random numbers

gc (changes requested):

PR #4612, opened 2022-10-18
Allow Creation of Uncharged Trident From Trident of The Seas (Full)

dzials:
### Description:

Added a creatable to craft the uncharged trident of the seas.  This is to fix the scenario where a kraken boss drops the full trident, but the toxic trident requires and uncharged trident to be made.

### Changes:

Added a creatable item instance.

### Other checks:


src/lib/data/creatables/slayer.ts
```diff
@@ -105,6 +105,14 @@ export const slayerCreatables: Createable[] = [
 		outputItems: new Bank({ 'Uncharged toxic trident': 1 }),
 		GPCost: 0
 	},
+	{
+		name: 'Uncharged trident',
+		inputItems: new Bank({
+			'Trident of the seas (full)': 1,
+		}),
+		outputItems: new Bank({ 'Uncharged trident': 1 }),
```
Lajnux, line 113:
```suggestion
		outputItems: new Bank({
			'Uncharged trident': 1,
			'Death rune': 2500,
			'Chaos rune': 2500,
			'Fire rune': 12_500,
		}),
```
Missed a few output items
image

Lajnux: > @Lajnux should this be moved to the revertables then and treated like the blowpipe?

if you add a way to create Trident of the seas then yes you can just make the opposite called revert Trident of the seas

dzials:
Discord_t98JH6Mcm9
Discord_5JsYjghNGL

PR #4621, opened 2022-10-21
[BSO] Allows minion to do multiple CoX trips since they're fast

themrrobert:
### Description:

Finally, the hotly demanded feature is here! 

Multiple CoX trips! Now you can send your minion on for a trip as long as you've earned! You pay a lot for the privilege of extra time on your trips, now you can finally reap the rewards!

### Changes:

- Quantity value on the command (requires sync_command)
- Adds notification square 🟧 to Steve / Takon drops ( I can just make this purple or w/e if you want)
- Had to rearrange the activity quite a bit because it was **very** one-raid focused.
- Fixes Sang staff [smaller boost] taking priority over Void staff.

### Other checks:

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -118,6 +118,7 @@ export const enum Emoji {
 	Purple = '🟪',
 	Green = '🟩',
 	Blue = '🟦',
+	Orange = '🟧',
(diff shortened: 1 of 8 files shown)


themrrobert:
Closing to to apply to osb first => bso

PR #4631, opened 2022-10-24
fix xp/hr visibility and xp strings + trawler xp bug

Lajnux:
### Description:

fix xp/hr visibility and xp strings + trawler xp bug

### Changes:

fix xp/hr visibility and xp strings + trawler xp bug

### Other checks:


src/lib/addXP.ts
```diff
@@ -139,7 +139,7 @@ export async function addXP(userID: string, params: AddXpParams): Promise<string
 		str += params.minimal
 			? `+${Math.ceil(preMax).toLocaleString()} ${skillEmoji[params.skillName]}`
 			: `You received ${Math.ceil(preMax).toLocaleString()} ${skillEmoji[params.skillName]} XP`;
-		if (params.duration && !params.minimal) {
+		if (params.duration) {
```
themrrobert, line 142: This is actually intended not to give XP/hr if minimal is selected. 

Unless Magna said it's okay to change this, I would leave it alone, and just add `minimal = false` to the `addXp()` calls where you want to see the XP/hr.

If you already got the OK from Magna, then ignore this.

src/tasks/minions/HunterActivity/aerialFishingActivity.ts
```diff
@@ -100,41 +100,32 @@ export const aerialFishingTask: MinionTask = {
...
-		await user.addXP({ skillName: SkillsEnum.Hunter, amount: huntXpReceived });
+		const fishXP = await user.addXP({
+			skillName: SkillsEnum.Fishing,
+			amount: fishXpReceived,
+			duration: data.duration
+		});
+		const huntXP = await user.addXP({
+			skillName: SkillsEnum.Hunter,
+			amount: huntXpReceived,
+			duration: data.duration
```
themrrobert, line 111: see you could add `, minimal: false` to this calls to get the XP/hr.

themrrobert (changes requested):
Sorry, I don't think this is a high priority like some bug fixes that can crash the bot

src/lib/addXP.ts
Lajnux: I like the minimal messages at some places but excluding the xp/hr part  I don't see necessary anywhere. It's good if xp/hr is always visible on any trip because it makes it way easier for the users to spot methods and stuff that result in a too low/too high xp/hr. For OSB especially where the xp/hrs are known but also useful for BSO. And yes most users could convert but our brains are lazy.

src/tasks/minions/HunterActivity/aerialFishingActivity.ts
Lajnux: > see you could add `, minimal: false` to this calls to get the XP/hr.

if it's undefined it also gives xp/hr.

PR #4635, opened 2022-10-25
Don't check pure essence quantity if using daeyalt

ghosts-are-spooky:
### Description:

Fixed issue where pure essence quantity was checked when Daeyalt runecrafting

### Changes:

Added a check if Daeyalt = true, don't run pure essence quantity check

### Other checks:


ghosts-are-spooky:
Closing to put this on master instead

PR #4636, opened 2022-10-25
Remove pure essence check for daeyalt runecrafting

ghosts-are-spooky:
### Description:

Fixed an issue where pure essence amount is checked when Daeyalt runecrafting

### Changes:

Added Daeyalt_essence = false as a requirement to check for pure essence quantity

### Other checks:


ghosts-are-spooky:
BSO Specific, OSB is coded differently in RC due to all BSO boosts

PR #4668, opened 2022-11-03
Allow cooking karambwans to repeat

TastyPumPum:
Adds an alias of "cooked karambwan" to karambwans. As it gets the name from itemID and is currently unable to repeat the trips.


Lajnux:
The name should be cooked karamwan and alias karambwan for consistency

gc (changes requested):
aliases are always lowercase

PR #4674, opened 2022-11-05
Dont re-fetch the user each addXP call

themrrobert:
### Description:

This is a very nice change as it saves multiple full user (MUser) fetches per task.

PVM tasks can add XP up to 6 different skills at a time, and most tasks add XP in at least 1 or 2+ skills.

Every time user.addXP() is called, it fetches a brand new MUser object, even though you're literally calling addXP from an MUser object.

### Changes:

- user.addXP() passes `this` instead of `this.id` to addXP()
- addXP() expects to receive `user: MUser` now

### Other checks:

### DIFF
--- a/src/lib/MUser.ts
+++ b/src/lib/MUser.ts
@@ -201,7 +201,7 @@ export class MUserClass {
 	}
 
 	addXP(params: AddXpParams) {
-		return addXP(this.id, params);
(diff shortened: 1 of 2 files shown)


Lajnux (approved):
Great PR and nice spotted.

PR #4676, opened 2022-11-06
Fix all current OSB Circular imports

### LINKED ISSUE #4732: New CL items at Revenants don't appear as purple
## 🐛 Bug Report

When you get a new CL slot, the item is outlined in purple to indicate so. At Revenants, this appears to have bugged out and it shows up as any other item. It definitely worked before though. 

Image of bug below (Can verify its not a dupe) 
image

themrrobert:
### Description:

This fixes all current circulars in OSB. Checkout before and after

Before:
20221113-osb-6

After:
20221113-osb-5

### Changes:

- Packages filterLootReplace() into it's own file
- Removes the extraneous use of filterLootReplace() in the revenantsActivity.ts file
- Lots more circular cleanup

### Other checks:

This is also important because the circular issue is compounded in bso:

20221106-bso-1

Fixes #4732

### DIFF
--- a/src/lib/clues/stashUnits.ts
+++ b/src/lib/clues/stashUnits.ts
@@ -4,7 +4,6 @@ import { partition } from 'e';
 import { CommandResponse } from 'mahoji/dist/lib/structures/ICommand';
 import { Bank } from 'oldschooljs';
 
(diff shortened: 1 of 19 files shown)


themrrobert:
*sigh*

PR #4677, opened 2022-11-06
Fix Moktang Circulars

themrrobert:
### Description:

Fixes Moktang circulars.

Activities should probably be separated from the commands to avoid circulars with handleTripFinish(). (Since handleTripFinish includes repeat trip, which includes all the commands, so the activities should **not** include handleTripFinish() )

Assuming the other Circular PR is also merged, here is the before + after:

Before:
20221106-bso-2

After:
20221106-bso-3

### Changes:

- Separates moktangActivity into it's own file.

### Other checks:


gc:
Fixed here https://github.com/oldschoolgg/oldschoolbot/commit/f8f6b3871f56965e1364090fc19a009276f56fb4

PR #4690, opened 2022-11-09
Fix PVM Degradeable boosts

themrrobert:
### Description:

Currently, the degradeable item boosts only work if the preferred attack style of the monster matches the weapon you're using.
This lets it work where it shouldn't, and NOT where it should.

This changes it so that if you are able to use that attack style on the monster, then the boost will be applied.

### Changes:

- Uses the current training style to determine if a degradeable item will boost.
- Splits the boost application from the charge degradation so the boost can influence the kills/trip.

### Other checks:

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -468,6 +468,17 @@ export function convertAttackStyleToGearSetup(style: OffenceGearStat | DefenceGe
 	return setup;
 }
 
+export function convertPvmStylesToGearSetup(attackStyles: SkillsEnum[]) {
+	const usedSetups: GearSetupType[] = [];
+	if (attackStyles.includes(SkillsEnum.Ranged)) usedSetups.push('range');
+	if (attackStyles.includes(SkillsEnum.Magic)) usedSetups.push('mage');
+	if (![SkillsEnum.Magic, SkillsEnum.Ranged].some(s => attackStyles.includes(s))) {
+		usedSetups.push('melee');
+	}
+	if (usedSetups.length === 0) usedSetups.push('melee');
+	return usedSetups;
+}
+
 export function sanitizeBank(bank: Bank) {
 	for (const [key, value] of Object.entries(bank.bank)) {
 		if (value < 1) {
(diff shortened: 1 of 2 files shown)


src/mahoji/lib/abstracted_commands/minionKill.ts
```diff
@@ -210,9 +211,8 @@ export async function minionKillCommand(
 	const degItemBeingUsed = [];
 	for (const degItem of degradeableItemsCanUse) {
 		const isUsing =
-			monster.attackStyleToUse &&
-			convertAttackStyleToGearSetup(monster.attackStyleToUse) === degItem.attackStyle &&
-			user.gear[degItem.attackStyle].hasEquipped(degItem.item.id);
+			convertPvmStylesToGearSetup(attackStyles).includes(degItem.attackStyle as GearSetupType) &&
+			user.gear[degItem.attackStyle as GearSetupType].hasEquipped(degItem.item.id);
```
gc, line 215: Dont cast things unnecessarily. Just give the property a type.

src/lib/util.ts
```diff
@@ -468,6 +468,17 @@ export function convertAttackStyleToGearSetup(style: OffenceGearStat | DefenceGe
...
+	const usedSetups: GearSetupType[] = [];
+	if (attackStyles.includes(SkillsEnum.Ranged)) usedSetups.push('range');
+	if (attackStyles.includes(SkillsEnum.Magic)) usedSetups.push('mage');
+	if (![SkillsEnum.Magic, SkillsEnum.Ranged].some(s => attackStyles.includes(s))) {
+		usedSetups.push('melee');
+	}
+	if (usedSetups.length === 0) usedSetups.push('melee');
+	return usedSetups;
+}
+
```
gc, line 481: we miiiight have another function that already does this?

gc (changes requested):

src/lib/util.ts
themrrobert: There's a similar one, but nothing that solves this problem yet.

The similar one is `convertAttackStyleToGearSetup` but 'AttackStyle' does mean the minionKill() SkillsEnum attack style, it uses GearStat source

PR #4697, opened 2022-11-10
BSO Leagues task fix for heat res. pots

Jonesy4:
Heat res. potions need 110 herblore to make. The league task had them as 120.


TastyPumPum (approved):
Looks good

PR #4714, opened 2022-11-15
Free t1 perks for maxed

gc:
### Other checks:

### DIFF
--- a/src/mahoji/commands/claim.ts
+++ b/src/mahoji/commands/claim.ts
@@ -1,0 +1,65 @@
+import { ApplicationCommandOptionType, CommandRunOptions } from 'mahoji';
+
+import { BitField, Channel } from '../../lib/constants';
+import { roboChimpUserFetch } from '../../lib/roboChimp';
+import { stringMatches } from '../../lib/util';
+import { sendToChannelID } from '../../lib/util/webhook';
+import { OSBMahojiCommand } from '../lib/util';
+
+const claimables = [
+	{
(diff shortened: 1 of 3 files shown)


src/mahoji/commands/claim.ts
```diff
@@ -0,0 +1,65 @@
...
+	],
+	run: async ({ options, userID }: CommandRunOptions<{ name: string }>) => {
+		const user = await mUserFetch(userID);
+		const claimable = claimables.find(i => stringMatches(i.name, options.name));
+		if (!claimable) {
+			return 'Invalid thing to claim.';
+		}
+
+		if (!(await claimable.hasRequirement(user))) {
+			return 'You are not elligible for this reward.';
```
Superwhammybar, line 59: Typo, it's `eligible`

src/mahoji/commands/claim.ts
```diff
@@ -0,0 +1,67 @@
...
+			autocomplete: async () => {
+				return claimables.map(i => ({
+					name: i.name,
+					value: i.name
+				}));
+			}
+		}
+	],
+	run: async ({ options, userID }: CommandRunOptions<{ name: string }>) => {
+		const user = await mUserFetch(userID);
```
themrrobert, line 54: i would put this below the claimable lookup so we don't fetch the user needlessly if the claimable isn't found

themrrobert (changes requested):

PR #4726, opened 2022-11-17
Redwoods, longer trip duaration

TastyPumPum:
Following the postive reaction in osb-discussion, allows redwoods to get the same treatment as cannonballs with the double trip length. https://discord.com/channels/342983479501389826/1042430911582572554


gc:
Would be appreciated to add this to wiki, somewhere.

PR #4729, opened 2022-11-17
Remove duplicate entries in /finish

themrrobert:
### Description:

- Remove duplicate entries in /finish when a custom entry overrides a default one.

### Changes:

- Add check to see if monster exists already in finishables before adding the standard monster pair.

### Other checks:


src/lib/finishables.ts
```diff
@@ -245,6 +245,7 @@ const monsterPairedCLs = Monsters.map(mon => {
 	const cl = allCollectionLogsFlat.find(c => stringMatches(c.name, mon.name));
 	if (!cl) return null;
 	if (!cl.items.every(id => mon.allItems.includes(id))) return null;
+	if (finishables.find(f => stringMatches(f.name, mon.name))) return null;
```
gc, line 248: use `.some` if youre not using the result of `.find`

gc (changes requested):

PR #4739, opened 2022-11-20
Fix /tools user temp_cl reset:True

themrrobert:
### Description:

Temp CL Reset not working

### Changes:

- Fixes the command so it checks temp_cl.reset === true, instead of temp_cl.

### Other checks:


themrrobert:
already fixed

PR #4748, opened 2022-11-22
Fix NodeJS Freezes in test bot from Canvas

themrrobert:
### Description:

This PR works-around an issue with Canvas exporting to jpg/png.

When the bot is not in production, it sets canvas.gpu = false, and changes nothing for production.

### Changes:

- Centralizes canvas creation to avoid needing to check production any time a canvas is created
- In createCanvas(), if (not production) disable GPU acceleration on Canvas

### Other checks:


gc:
Any more info about this? the server doesnt even have a GPU.

themrrobert:
It doesn't hurt us to disable the GPU in dev.  it's not even enabled in prod because there is no graphics card.

but this simple change will greatly help my ability to continue contributing because i won't have to kill my entire terminal every time to restart the bot.

PR #4757, opened 2022-11-28
Add Nex finishable to BSO

themrrobert:
### Description:

Adds Nex to /finish

### Other checks:

### DIFF
--- a/src/lib/finishables.ts
+++ b/src/lib/finishables.ts
@@ -35,7 +35,7 @@ import {
 import pets from './data/pets';
 import { MoktangLootTable } from './minions/data/killableMonsters/custom/bosses/Moktang';
 import { Naxxus } from './minions/data/killableMonsters/custom/bosses/Naxxus';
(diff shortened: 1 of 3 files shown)


src/lib/finishables.ts
```diff
@@ -161,6 +164,17 @@ export const finishables: Finishable[] = [
...
 	},
+	{
+		name: 'Nex',
+		cl: nexCL.filter(i => i !== itemID('Frozen key')),
+		kill: () => {
+			const loot = new Bank(NexMonster.table.kill(1, {}));
+			if (roll(82)) {
+				loot.add(randArrItem(nexUniqueDrops), 1);
+			}
+			return loot;
```
gc, line 175: This `roll` here is using a magic number, that needs to be abstracted & shared properly with the real nex code. e.g. `export const NEX_UNIQUE_DROPRATE = (teamSize: number) => something * teamSize;`

gc (changes requested):

PR #4770, opened 2022-12-12
Level Up Table Messages

Lajnux:
### Description:

Level up table messages

### Changes:

Added Level up table messages for each skill.

### Other checks:


gc:
I'm not going to close the PR outright, but I have no plans on merging this, and it likely never will be.

PR #4776, opened 2022-12-14
Make /trade consistent, reliable, and remove item deletion bugs (transactional)

themrrobert:
### Description:

The risk for item deletion in the /trade command is relatively large. Several database calls are performed sequentially, committing after each one, which is bad if there's any sort of bot lag/freeze/stall, or similar with the psql server, or even the network stack.

Some of these are much less likely than others, nevertheless, we do see this issue crop up with some regularity. 
(We see it all the time, but usually it doesn't result in item loss.)

Twice just today it resulted in item loss, however, so I'm making this a top priority.

### Changes:

- Removes the 4 separate steps and replaces it with a new function, tradePlayerItems
- Trade player items prepares 2 queries, one to update each user, and runs both of them in a single transaction, **so either both succeed, or neither do.**
- Adds functions to both users `userQueue` simultaneously, so that other operations which act on the bank are properly queued. (Even though this should be unnecessary with a single transaction)

### Other checks:

I even tested the recipients user queue to make sure it properly delays other addItemsToBank() functions, like buy items.

Everything is working as intended.

### DIFF
--- a/src/lib/util/tradePlayerItems.ts
+++ b/src/lib/util/tradePlayerItems.ts
@@ -4,7 +4,7 @@ import { prisma } from '../settings/prisma';
 import { logError } from './logError';
 import { userQueueFn } from './userQueues';
 
-const activeTradeCache = new Map();
+export const activeTradeCache = new Map();
 
 function timeout(ms: number) {
 	return new Promise(resolve => setTimeout(resolve, ms));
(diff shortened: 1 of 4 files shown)


gc:
I'm closing this PR due to 4 months of inactivity, I'm not denying it, just cleaning up some PRs. feel free to re-open if you want to continue working on it. Although, we'd probably need to discuss the implementation of it. I think this would be better off with refactoring `transactItemsFromBank`, so we can have a function which returns the plain `prisma.user.update` promises, so we can update both users in a transaction, and so all bank updates continue to go through one place.

PR #4786, opened 2022-12-16
Update rates - finish pets

TastyPumPum:
Update the finish pets to the more common methods of obtaining the pets.


gc:
These were coded as being the most common training methods, not the most common methods of farming pets, which are usually NOT the most common training methods. And most people just get pets from their regular training methods.

TastyPumPum:
> These were coded as being the most common training methods, not the most common methods of farming pets, which are usually NOT the most common training methods. And most people just get pets from their regular training methods.

Think it's rare people chop magic trees or fish monkfish anymore really but understandable I just made the pr as it was suggested to update them as they're outdated really.

gc:
Mind runes and volcanic mine?

6enny6:
most people do not get the pets from regular training methods, you dont even get close to the rates by having 99 in a skill. So post 99 you would go for the farming pet methods,. 

These rates are taken from the osrs petcord (a server showing the quickest way to get each pet) and excluding any alt or tick manipulation methods. See mind runes and VM below.

image

image

6enny6:
maybe we could add a method: option? or something similar

/finish input:Heron pet method:Karambwans

for example

gc:
Like I said though, these were coded as being the "meta way to hunt pets", they're coded as being the most recognizable, common places that people get them.

PR #4819, opened 2022-12-27
Adjust cannonball rates

TastyPumPum:
Currently the rates for cannonballs are way off, this pr adjusts the rates to match in game. With access to shilo village furnace and double ammo mould, this gives 5.3k cannonballs an hour the current rates are around 7.9k an hour. https://www.youtube.com/watch?v=GIfinzK4wYE


Lajnux:
I think it's better to move cballs to smelting same as ingame. I assume that's one of the reasons it's inaccurate atm.

PR #4821, opened 2022-12-29
Nightmare zone

Lajnux:
### Description:

Added the minigame Nightmare Zone

### Changes:

Added the minigame Nightmare Zone

### Other checks:


src/lib/util/calcMaxTripLength.ts
```diff
@@ -44,6 +44,10 @@ export function calcMaxTripLength(user: MUser, activity?: activity_type_enum) {
 			max *= 2;
 			break;
 		}
+		case 'NightmareZone': {
+			max *= 5;
+			break;
+		}
```
gc, line 50: TBD...

src/lib/types/minions.ts
```diff
@@ -346,6 +346,11 @@ export interface GuardiansOfTheRiftActivityTaskOptions extends MinigameActivityT
 	combinationRunes: boolean;
 }
 
+export interface NightmareZoneActivityTaskOptions extends MinigameActivityTaskOptions {
+	strategy: string;
```
gc, line 350: strategy should be an enum, using a single source/enum for anywhere the strategy is used (e.g. autocomplete)

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,479 @@
...
+		interaction,
+		`Are you sure you want to spend **${cost.toLocaleString()}** Nightmare Zone points to buy **${quantity.toLocaleString()}x ${
+			shopItem.name
+		}**?`
+	);
+
+	await transactItems({
+		userID: user.id,
+		collectionLog: true,
+		itemsToAdd: new Bank(shopItem.output).multiply(quantity)
```
gc, line 434: do the way we normally do:
`const loot = new Bank()`
`Are you sure you want to spend ${x} to buy ${loot}?`

src/tasks/minions/minigames/nightmareZoneActivity.ts
```diff
@@ -0,0 +1,58 @@
...
+import { handleTripFinish } from '../../../lib/util/handleTripFinish';
+import { NightmareZoneActivityTaskOptions } from './../../../lib/types/minions';
+
+export const nightmareZoneTask: MinionTask = {
+	type: 'NightmareZone',
+	async run(data: NightmareZoneActivityTaskOptions) {
+		const { quantity, userID, channelID, duration, strategy } = data;
+		const user = await mUserFetch(userID);
+
+		const [, , attackStyles] = resolveAttackStyles(user, { monsterID: Monsters.Cow.id });
```
gc, line 15: again, this makes no sense. its not even an easter egg. unless im missing something.

gc (changes requested):
seems good to me after these are fixed, and bigly assuming user feedback has been addressed and are accurate. 5x longer max trip length is excessive. Do the same as cannonballs, or maybe a bit longer. Not 5x.

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+import { formatDuration, hasSkillReqs, stringMatches } from '../../../lib/util';
+import addSubTaskToActivityTask from '../../../lib/util/addSubTaskToActivityTask';
+import { calcMaxTripLength } from '../../../lib/util/calcMaxTripLength';
+import getOSItem from '../../../lib/util/getOSItem';
+import { updateBankSetting } from '../../../lib/util/updateBankSetting';
+import { handleMahojiConfirmation } from '../../mahojiSettings';
+import { NightmareZoneActivityTaskOptions } from './../../../lib/types/minions';
+
+export const NMZ_STRATEGY = ['experience', 'points'] as const;
+export type NMZStrategy = typeof NMZ_STRATEGY[number];
```
gc, line 20: It'd be a good idea to move this to `constants.ts`, because it usually creates problems to import things from command files.

src/lib/util/calcMaxTripLength.ts
```diff
@@ -44,6 +44,10 @@ export function calcMaxTripLength(user: MUser, activity?: activity_type_enum) {
 			max *= 2;
 			break;
 		}
+		case 'NightmareZone': {
+			max *= 3;
+			break;
+		}
```
gc, line 50: I'm thinking 2 or 2.5.

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+		defence: 70,
+		strength: 70,
+		attack: 70,
+		hitpoints: 70,
+		prayer: 43
+	};
+
+	const [hasReqs] = hasSkillReqs(user, skillReqs);
+	if (!hasReqs) {
+		return `You not meet skill requirements, you need ${Object.entries(skillReqs)
```
gc, line 295: grammar

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+		return 'The Nightmare Zone minigame requires melee combat for efficiency, swap training style using `/minion train style:`';
+	}
+
+	let percent = round(calcWhatPercent(skillTotal, (attackStyles.length + 1) * 99), 2);
+	percent = Math.min(65, (percent - 70) / 0.46);
+	boosts.push(`${percent.toFixed(2)}% for your trained stats (incudling HP)`);
+	timePerMonster = reduceNumByPercent(timePerMonster, percent);
+
+	// Reduce time for item boosts
+	itemBoosts.forEach(set => {
```
gc, line 334: use `for of`, we never use `forEach`.

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+		itemsToAdd: loot
+	});
+
+	await user.update({
+		nmz_points: {
+			decrement: cost
+		}
+	});
+
+	return `You successfully bought **${quantity.toLocaleString()}x ${shopItem.name}** for ${(
```
gc, line 446: use `${loot}` here

src/tasks/minions/minigames/nightmareZoneActivity.ts
```diff
@@ -0,0 +1,57 @@
...
+import { NightmareZoneActivityTaskOptions } from './../../../lib/types/minions';
+
+export const nightmareZoneTask: MinionTask = {
+	type: 'NightmareZone',
+	async run(data: NightmareZoneActivityTaskOptions) {
+		const { quantity, userID, channelID, duration, strategy } = data;
+		const user = await mUserFetch(userID);
+
+		// No monster ID so use -1 as placeholder
+		const [, , attackStyles] = resolveAttackStyles(user, { monsterID: -1 });
```
gc, line 14: This is a bad thing to do. I'd much rather you refactored `resolveAttackStyles` so the monsterID is optional.

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+		}
+	});
+
+	const maxTripLength = calcMaxTripLength(user, 'NightmareZone');
+	const quantity = Math.floor(maxTripLength / timePerMonster);
+	const duration = quantity * timePerMonster;
+	// Consume GP (and prayer potion if experience setup)
+	const dreamCost = new Bank().add(
+		'Coins',
+		Math.floor(((QP >= MAX_QP ? 16_000 : 26_000) * duration) / ((strategy === 'points' ? 30 : 60) * Time.Minute))
```
gc, line 350: This is one very complex line here. I'd probably make a function just for this, and calculate it a bit more imperatively if you can, e.g. same way we do max trip length

```ts
let base = x;
if (something) base /= 2
else base *= 2 / duration
```
otherwise, this line is far too complex to read

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+	);
+	if (strategy === 'experience') {
+		dreamCost.add('Prayer potion(4)', Math.floor(duration / (Time.Minute * 5)));
+	}
```
gc, line 354: cant add 0 here i assume

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+	const boosts = [];
+
+	let timePerMonster = Time.Minute * 2;
+
+	// combat stat boosts
+	// No monster ID so use -1 as placeholder
+	const [, , attackStyles] = resolveAttackStyles(user, { monsterID: -1 });
+	const skillTotal = sumArr(attackStyles.map(s => user.skillLevel(s))) + user.skillLevel(SkillsEnum.Hitpoints);
+	if (attackStyles.includes(SkillsEnum.Ranged) || attackStyles.includes(SkillsEnum.Magic)) {
+		return 'The Nightmare Zone minigame requires melee combat for efficiency, swap training style using `/minion train style:`';
```
gc, line 325: We could add buttons for switching attack styles, so the user can just click a button here, but you dont have to do that.

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+	if (!hasReqs) {
+		return `You not meet skill requirements, you need ${Object.entries(skillReqs)
+			.map(([name, lvl]) => `${lvl} ${name}`)
+			.join(', ')}.`;
+	}
+
+	const { QP } = user;
+
+	if (QP < 100) {
+		return `The Nightmare Zone minigame requires **100 QP**, and you have ${QP} QP.\n`;
```
gc, line 303: Unnecessary newline at the end.

src/mahoji/lib/abstracted_commands/nightmareZoneCommand.ts
```diff
@@ -0,0 +1,482 @@
...
+		aliases: []
+	},
+	/* TODO: Lock this to max 15 Herb boxes per day
+	{
+		name: 'Herb box',
+		output: new Bank().add('Herb box', 1),
+		cost: 9500,
+		aliases: []
+	},
+	*/
```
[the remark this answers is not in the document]
Lajnux: yes, didn't come up with any good way to lock Herb boxes to be only purchasable 15 times in 24 hours. Also kinda requires a oldschoolJS openable if we want it 100%

src/lib/util/calcMaxTripLength.ts
Lajnux: temp

Also if it's less than 3x, NMZ will be dead content on release since normal combat training methods in bot beat it. The only benefit with NMZ will be to send 1x trip each day before bed time.

PR #4835, opened 2023-01-03
Mahogany Homes, tier contract selection

### LINKED ISSUE #2979: Add a way to use lower-tier planks for Mahogany homes
## 🚀 Feature Proposal

It seems once you hit level 70 Con., Mahogany homes will always use mahogany planks and nothing else. This should be changed to allow players to use Teak planks if they so wish, at the cost of lower exp/hr. You can use Teak planks at level 50, but not at 70.

### LINKED ISSUE #4815: Mahogany homes - option
## 🚀 Feature Proposal

Would it be possible to add an option to mahogany homes start to choose expert, adept etc?

It would make a big difference for ironmen, using mahog homes is many hundreds of mil cheaper than normal con, and adept vs expert is about 400-500m cheaper

would really appreciate it

TastyPumPum:
Allows users to select which tier they want to do for mahogany homes. Defaults to the highest tier they have the level for if no tier is selected.

Closes #4815 
Closes #2979


src/mahoji/lib/abstracted_commands/mahoganyHomesCommand.ts
```diff
@@ -121,15 +134,25 @@ export async function mahoganyHomesBuyCommand(user: MUser, input = '', quantity?
...
+export async function mahoganyHomesBuildCommand(user: MUser, channelID: string, tier?: string): CommandResponse {
 	if (user.minionIsBusy) return `${user.minionName} is currently busy.`;
 
 	const conLevel = user.skillLevel(SkillsEnum.Construction);
 	const kc = await getMinigameScore(user.id, 'mahogany_homes');
 
+	let tierData = contractTiers.find(contractTier => conLevel >= contractTier.level)!;
+
+	if (tier) {
+		tierData = contractTiers.find(contractTier => tier.toLowerCase() === contractTier.name.toLowerCase())!;
```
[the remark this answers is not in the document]
TastyPumPum: Have made it so if no match is found it will use the highest tier, thanks!

src/lib/types/minions.ts
```diff
@@ -229,6 +229,7 @@ export interface MahoganyHomesActivityTaskOptions extends MinigameActivityTaskOp
 	xp: number;
 	quantity: number;
 	points: number;
+	tier: string;
```
gc, line 232: make this `tierID: number`, and give tiers, IDs. Just 1,2,3, etc. It will require a change to fix repeating too probably. otherwise, you will store a string like "Expert (Mahogany Planks)" in the database everytime someone does a trip. a number is smaller to store than even 1 letter, basically.

src/mahoji/lib/abstracted_commands/mahoganyHomesCommand.ts
```diff
@@ -121,15 +134,27 @@ export async function mahoganyHomesBuyCommand(user: MUser, input = '', quantity?
...
 
 	const conLevel = user.skillLevel(SkillsEnum.Construction);
 	const kc = await getMinigameScore(user.id, 'mahogany_homes');
 
+	let tierData = contractTiers.find(contractTier => conLevel >= contractTier.level)!;
+
+	if (tier) {
+		tierData =
+			contractTiers.find(contractTier => tier.toLowerCase() === contractTier.name.toLowerCase()) || tierData;
+		if (!tierData) return 'Error selecting contract tier.';
```
gc, line 148: Here, and below, if this is a scenario which we think should never happen, just throw an error instead. otherwise, by doing this, you're potentially secretly hiding bugs.

gc (changes requested):

Issue #4848, opened 2023-01-07
Shades of Morton - diary string


LuxrayElite:
## 🐛 Bug Report

Currently the firemaking diary boost says it is for the hard morytania diary however it's obtained via elite diary

Brought up in testing by Krisk

LuxrayElite:
Fixed but not commited

PR #4849, opened 2023-01-07
fix shade string returns

osDreamscape:
added user tag, changed fm boost string to Elite to correctly match the function, minor grammar fixes

reference to issues #4848

### Other checks:


osDreamscape:
issue #4848 already closed

PR #4854, opened 2023-01-08
Maniacal monkey

Lajnux:
### Description:

Make Maniacal monkey killable

### Changes:

Make Maniacal monkey killable

### Other checks:

### DIFF
--- a/src/lib/simulation/maniacalMonkey.ts
+++ b/src/lib/simulation/maniacalMonkey.ts
@@ -1,0 +1,20 @@
+import LootTable from 'oldschooljs/dist/structures/LootTable';
+
+export const ManiacalMonkeyTable = new LootTable()
+	.every('Small ninja monkey bones')
+
+	/* Weapons and armour */
+	.add('Steel scimitar', 1, 2)
+	.add('Maple shortbow', 1, 1)
+
+	/* Materials */
+	.add('Oak plank', 1, 3)
+	.add('Grimy guam leaf', 1, 2)
+
+	/* Other */
+	.add('Bass', 2, 4)
+	.add('Banana', 1, 2)
+	.add('Antipoison(2)', 1, 2)
+	.add('Prayer potion(1)', 1, 2)
+	.add('Adamant arrow(p++)', 1, 1)
+	.add('Rope', 1, 1);
--- a/src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
+++ b/src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
@@ -1,22 +1,26 @@
-import { ManiacalMonkeyTaskOptions } from './../../../lib/types/minions';
+import { Time } from 'e';
 
-import { PvMMethod } from './../../../lib/constants';
+import { resolveAttackStyles } from '../../../lib/minions/functions';
 import { Skills } from '../../../lib/types';
 import { formatDuration, hasSkillReqs } from '../../../lib/util';
 import addSubTaskToActivityTask from '../../../lib/util/addSubTaskToActivityTask';
 import { calcMaxTripLength } from '../../../lib/util/calcMaxTripLength';
-import { resolveAttackStyles } from '../../../lib/minions/functions';
-import { Time } from 'e';
+import { PvMMethod } from './../../../lib/constants';
+import { ManiacalMonkeyTaskOptions } from './../../../lib/types/minions';
 
-
-export async function maniacalMonkeyCommand(user: MUser, channelID: string, quantity: number | undefined, method: PvMMethod | undefined) {
+export async function maniacalMonkeyCommand(
+	user: MUser,
+	channelID: string,
+	quantity: number | undefined,
+	method: PvMMethod | undefined
+) {
 	const { minionName } = user;
 
 	const skillReqs: Skills = {
 		hitpoints: 74,
 		prayer: 74
 	};
-	
+
 	const [hasReqs] = hasSkillReqs(user, skillReqs);
 	if (!hasReqs) {
 		return `You not meet skill requirements, you need ${Object.entries(skillReqs)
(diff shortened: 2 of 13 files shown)


src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
```diff
@@ -0,0 +1,315 @@
...
+
+	const [hasReqs] = hasSkillReqs(user, skillReqs);
+	if (!hasReqs) {
+		return `You not meet skill requirements, you need ${Object.entries(skillReqs)
+			.map(([name, lvl]) => `${lvl} ${name}`)
+			.join(', ')}.`;
+	}
+	const magicLevel = user.skillLevel('magic');
+	const rangedLevel = user.skillLevel('ranged');
+	if ((method === 'burst' || method === 'barrage') && magicLevel < 62) {
```
LuxrayElite, line 46: Later in the code you compare magic level to be minimum 70 for training magic (bursting) so it may be better to enforce level 70 here and remove that check later down the line?

src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
```diff
@@ -0,0 +1,315 @@
...
+			boosts.push('5% for Kodai wand');
+		}
+		if (user.hasEquippedOrInBank('Ancestral robe top')) {
+			timeToFinish *= (100 - 3) / 100;
+			boosts.push('3% for Ancestral robe top');
+		}
+		if (user.hasEquippedOrInBank('Ancestral robe bottom')) {
+			timeToFinish *= (100 - 2) / 100;
+			boosts.push('2% for Ancestral robe bottom');
+		}
```
LuxrayElite, line 134: Is there a way that you could prevent repeating the boosts?

src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
```diff
@@ -0,0 +1,315 @@
...
+		// Check what Chinchompa to use
+		const chinchompas = ['Black chinchompa', 'Red chinchompa', 'Chinchompa'];
+		let chinchompa = 'Black chinchompa';
+		for (let chin of chinchompas) {
+			if (user.owns(chin) && user.bank.amount(chin) > 5000) {
+				chinchompa = chin;
+				break;
+			}
+		}
+		const regularVoid = resolveItems([
```
LuxrayElite, line 148: Since this is already a thing in TOB code would it be more beneficial to create it as an export somewhere? Just helps any code relying on void in the future. I know it may go outside of the scope of this pr

src/lib/simulation/maniacalMonkey.ts
```diff
@@ -0,0 +1,20 @@
+import LootTable from 'oldschooljs/dist/structures/LootTable';
+
```
LuxrayElite, line 2: Wouldn't this be better in OSJS so you can export it as a monster as well as a loot-table? I don't see why it would require special loot simulation?
This would make it an alternative for monkeys and slayer compatible.

src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
```diff
@@ -0,0 +1,315 @@
...
+			timeToFinish = reduceNumByPercent(timeToFinish, chinBoostLongRanged);
+			boosts.push(`${chinBoostLongRanged}% for ${chinchompa}`);
+		} else {
+			timeToFinish = reduceNumByPercent(timeToFinish, chinBoostRapid);
+			boosts.push(`${chinBoostRapid}% for ${chinchompa}`);
+		}
+		consumableCosts.push(chinningConsumables);
+	}
+
+	// If no quantity provided, set it to the max.
```
LuxrayElite, line 186: I think everything past this point is a copy and paste of minionKill with a few tweeks for activity tracking and chinning perhaps a different approach could be made.

LuxrayElite (commented):
Just giving a few bits of feedback after reading through the code.

I personally think this would be better as a monster and added under /k. This would reduce repetition of code, make it easier to code boosts, follow existing formats of monsters.

Creature creation is a good for approaching this style of implementation

src/lib/simulation/maniacalMonkey.ts
Lajnux: yea I agree, it should all be as just a killableMonster. I just didn't wanna wait on the oldschoolJS update. :D

src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
Lajnux: I made a killableMonster PR for it now, just need to fix rates when and if oldschoolJS gets updated. 
This was the approach I wanted to go with first just didn't like being locked behind another PR :(

src/tasks/minions/maniacalMonkeyActivity.ts
```diff
@@ -0,0 +1,73 @@
...
+				amount: Math.floor(manicalMonkeyHP * quantity * 1.33),
+				duration
+			})
+		);
+
+		let loot = new Bank();
+
+		for (let i = 0; i < quantity; i++) {
+			loot.add(ManiacalMonkeyTable.roll());
+		}
```
gc, line 48: You're not giving all this loot are you? Nobody gets every single loot from monkeys banked after killing them. Only stackable things?

src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
```diff
@@ -0,0 +1,315 @@
...
+		}
+		const regularVoid = resolveItems([
+			'Void knight top',
+			'Void knight robe',
+			'Void knight gloves',
+			'Void ranger helm'
+		]);
+		const eliteVoid = resolveItems(['Elite void top', 'Elite void robe', 'Void knight gloves', 'Void ranger helm']);
+		if (user.hasEquippedOrInBank(eliteVoid, 'every')) {
+			timeToFinish *= (100 - 10) / 100;
```
gc, line 156: `reduceNumByPercent` fn exists for this.

src/mahoji/lib/abstracted_commands/maniacalMonkeyCommand.ts
gc: yea these are quite a bit of boosts. putting them in a data format would be better.

gc (changes requested):

PR #4858, opened 2023-01-09
Allow the Castle Wars Buyables to be sold back for tickets with /sell

### LINKED ISSUE #4838: castle wars items sell warning
add a warning when using /sell on castle wars items.

warning should say to revert cw items into tickets use /create revert

TastyPumPum:
As the castle wars buyables are under /buy it makes more sense to have the opposite as /sell, people in help and support have accidentally sold for 0 coins and 0 tickets because of this confusion. 

Alternatively we could also leave the revertables in place and just add this to sell command to have both options in place.

Closes #4838 


src/mahoji/commands/sell.ts
```diff
@@ -85,6 +86,30 @@ export const sellCommand: OSBMahojiCommand = {
...
+				interaction,
+				`${user}, please confirm you want to sell ${castleWarsBank} for **${castleWarsLoot}**.`
+			);
+
+			await user.removeItemsFromBank(castleWarsBank);
+			await transactItems({
+				userID: user.id,
+				collectionLog: false,
+				itemsToAdd: castleWarsLoot
+			});
```
gc, line 109: transactItems can remove and add items at the same time. no need to do this separately

gc (changes requested):

PR #4860, opened 2023-01-09
gilded skilling item

### LINKED ISSUE #4859: Gilded axe gives no skilling boost
## 🐛 Bug Report

Gilded axe gives no speed boost while doing woodcutting. Lower tier axes (steel, iron) give a speed boost while having gilded axe equipped and/or in the bank. gilded should give the same boost as rune axe.
975D08A1-A945-4F42-8DEA-52CA2E8C2EE4

Souzer:
gives gilded tools the same speed boosts as rune tools

closes #4859

### Other checks:

### DIFF
--- a/src/lib/data/similarItems.ts
+++ b/src/lib/data/similarItems.ts
@@ -56,8 +56,8 @@ const bowfaCorruptSimilar = resolveItems([
 ]);
 
 const source: [string, (string | number)[]][] = [
-	['Rune Axe', ['Gilded Axe']],
-	['Rune Pickaxe', ['Gilded Pickaxe']],
+	['Rune axe', ['Gilded axe']],
+	['Rune pickaxe', ['Gilded pickaxe']],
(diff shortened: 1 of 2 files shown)


src/lib/data/similarItems.ts
```diff
@@ -56,6 +56,8 @@ const bowfaCorruptSimilar = resolveItems([
 ]);
 
 const source: [string, (string | number)[]][] = [
+	['Rune Axe', ['Gilded Axe']],
+	['Rune Pickaxe', ['Gilded Pickaxe']],
```
gc, line 60: these arent following the capitalization standard, `Rune axe`

gc (changes requested):

gc:
Actually I don't think this is even the right way to do it. Just go to the place where their boost is defined, and change the gilded axe/whatever to match the rune one.

PR #4864, opened 2023-01-11
maniacal monkey as killable monster (chinning)

Lajnux:
### Description:

Requires updated oldschoolJS with Maniacal monkey.
Maniacal monkey as killable monster instead, introduce chinning variable

### Changes:

Maniacal monkey as killable monster instead, introduce chinning variable

### Other checks:


gc:
I'm assuming you got the xp/hr rates balanced. Aside from that, my only concern I think is the loot. Its true in the bot, we let minions keep all the loot, but how much loot do you get from a chinning trip in this?

PR #4741, opened 2022-11-21
Barbarian Potions + Roe/Caviar 

TastyPumPum:
Adds the cutting of leaping fish to the cook command. 
Moved it into its own command and activity to keep it tidy.
Used the same rates as crushing as its a similar method.

Additionally added barbarian mixes to the /mix and /decant commands
Also tidied up the mixables using item: getOSItem()

-   [x] I have tested all my changes thoroughly.

PR #4865, opened 2023-01-12
barbarian mixes

Souzer:
### Description:

add the ability to make barbarian potions 

requires #4741 to be merged because its impossible to get roe/caviar right now

### Changes:

- added all barbarian mixes
_magic essence mix and relicyms mix both work but required potions are not available in the bot_
- added the ability to decant all mixes to their 1 dose and back to 2 dose

### Other checks:


src/lib/skilling/skills/herblore/mixables/barbMixes.ts
```diff
@@ -0,0 +1,390 @@
...
+
+const barbMixes: Mixable[] = [
+	{
+		name: 'Attack mix(2)',
+		aliases: ['attack mix roe', 'attack mix(2)', 'attack mix 2 roe'],
+		id: itemID('Attack mix(2)'),
+		level: 4,
+		xp: 8,
+		inputItems: new Bank({
+			'Attack Potion(2)': 1,
```
gc, line 14: Items need to follow the osrs format for capitalization

src/lib/skilling/skills/herblore/mixables/barbMixes.ts
```diff
@@ -0,0 +1,390 @@
+import { Bank } from 'oldschooljs';
+
+import itemID from '../../../../util/itemID';
+import { Mixable } from '../../../types';
+
+const barbMixes: Mixable[] = [
+	{
+		name: 'Attack mix(2)',
+		aliases: ['attack mix roe', 'attack mix(2)', 'attack mix 2 roe'],
+		id: itemID('Attack mix(2)'),
```
gc, line 10: instead of having name and id, you can have just `item: getOSItem()` and now you have the full item (id and name included), AND you dont have to worry about typos/mismatches on the name or id.

src/lib/skilling/skills/herblore/mixables/barbMixes.ts
```diff
@@ -0,0 +1,390 @@
...
+		inputItems: new Bank({
+			'Extended super antifire(2)': 1,
+			Caviar: 1
+		}),
+		tickRate: 1,
+		bankTimePerPotion: 0.088
+	}
+];
+
+export default barbMixes;
```
gc, line 390: we dont use export defaults anywhere. you were just copying how the other imports do that, so you were doing a smart thing, but anything using export default was written a while ago, and we want to move away from that in the future.

gc (changes requested):
Also: you will need to update/add tests in decanting.test.ts

PR #4869, opened 2023-01-14
Make clue buttons a function

### LINKED ISSUE #4749: Add clue buttons when getting a clue from an openable
Would be helpful for implings especially!

TastyPumPum:
Makes a function to build the clue buttons
Adds the clue buttons to openables
Adds the clue buttons to barb gamble

Closes #4749 


src/mahoji/lib/abstracted_commands/openCommand.ts
```diff
@@ -34,6 +37,17 @@ async function addToOpenablesScores(mahojiUser: MUser, kcBank: Bank) {
...
+export function buildClueButtons(loot: Bank | null, perkTier: number) {
+	const components: ButtonBuilder[] = [];
+	if (loot && perkTier > PerkTier.One) {
+		const clueReceived = ClueTiers.filter(tier => loot.amount(tier.scrollID) > 0);
+		if (clueReceived.length > 0) {
+			clueReceived.map(clue => components.push(makeDoClueButton(clue)));
+		}
+	}
+	return components;
+}
```
gc, line 49: this should probably go in the util file, it could create circular imports and other issues. Generally: we prefer to export things from command files.

gc: you can check if theres a circular import with: `yarn build` and `npx madge --circular ./dist`, if any of the files in the list outputted are ones you changed, then probably you made a circular import

gc (changes requested):

src/mahoji/lib/abstracted_commands/openCommand.ts
TastyPumPum: Moved it to a util file (hopefully a suitable one but let me know if you think it should be moved)
Ran the `npx madge --circular ./dist` and couldn't notice any difference in the return compared to master branch.

Let me know if you feel it should be moved elsewhere though.

TastyPumPum:
Looks like this has one circular dependency now due to how the auto farming contract runs and opens the seed pack:
 lib/util/globalInteractions.js > mahoji/lib/abstracted_commands/farmingContractCommand.js > mahoji/lib/abstracted_commands/openCommand.js
 
 Any advise would be appreciated.

osrsjobbo:
@TastyPumPum See my most recent commits which fix your circular dependency

PR #4891, opened 2023-01-29
Add favourite alchs bank filter

osrsjobbo:
### Description:

Add a `Favourite Alchs` which shows items in bank which are set to be alchable. 

image

### Changes:

- Add new filter called Favourite Alchs

### Other checks:


osrsjobbo:
Part of me thinks the counter of this could be helpful, i.e. show me the items in my bank which aren't set to alchable, so I can easily identify what I need to add to my favourite alchs. Thoughts appreciated.

PR #4923, opened 2023-02-05
Show qty owned when offering and limit to what is owned

### LINKED ISSUE #4362: /offer quantity owned
## 🚀 Feature Proposal

add the amount of items you own for items you /offer similar to how it is in /open

osrsjobbo:
### Description:

This closes #4362.

Change the `/offer` command to only show what is owned, and show the qty owned. 

Have built it so the command will also filter if a value is provided. 

Only showing what is owned is potentially contensious, and can be easily removed, but I think showing qty owned is a good addition. 

### Changes:

- Change offer command to show only what is owned and the qty owned. 

### Other checks:

Tested on my own server

### DIFF
--- a/src/mahoji/commands/offer.ts
+++ b/src/mahoji/commands/offer.ts
@@ -1,3 +1,4 @@
+import { User } from 'discord.js';
 import { randArrItem, randInt, roll, Time } from 'e';
 import { ApplicationCommandOptionType, CommandRunOptions } from 'mahoji';
 import { Bank } from 'oldschooljs';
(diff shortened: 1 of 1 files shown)


src/mahoji/commands/offer.ts
```diff
@@ -60,13 +62,23 @@ export const mineCommand: OSBMahojiCommand = {
...
-						value: i.name
-					}));
+			autocomplete: async (value: string, user: User) => {
+				const botUser = await mUserFetch(user.id);
+
+				let offerables = new Set(
+					[...Offerables, ...specialBones.map(i => i.item), ...eggs, ...Prayer.Bones]
+						.map(i => resolveItems(i.name))
+						.map(i => i[0])
+				);
```
gc, line 72: this can be at the top level?

gc (changes requested):

src/mahoji/commands/offer.ts
osrsjobbo: Hopefully that's what you were after.

PR #4932, opened 2023-02-07
Feature/uncharge shadow

ghosts-are-spooky:
### Description:

Updated minion charge command to allow uncharging of items.

### Changes:
- Added a new variable on degreableitems, this was needed as some items e.g. blood fury return different items to input cost. 
- Updated command to have boolean uncharge option.
- Updated messages to be clear around creating vs charging. Added warning when uncharging if the item would not return 100% of input.

### Other checks:


ghosts-are-spooky:
Fixes #4941 
This increases limit to 100k

gc (commented):
FYI: the concern with this one is dupe bugs, and we'll probably leave this PR for a bit until I feel comfortable to review it carefully and test it. If you wanted to really  go out your way, what we really need is a testable way of handling things like this, which would be fairly hard to code. (somewhat similar to i think, the autoequip function, which doesnt do anything except just tell the bot what database updates to do, so we can test the result easily).


So basically, its a complex change with a risk for dupes/bugs, and I'm gonna leave it for the time being.

PR #4936, opened 2023-02-08
Fix gear command and add tests

themrrobert:
### Description:

Fixes gear command and adds tests that would catch the issues that occured.

### Changes:

Adds quantity where needed
Refunds 2h when a 2h is equipped
Adds tests

### Other checks:


gc:
DM'd you

PR #4942, opened 2023-02-11
Update prisma

gc:
### Other checks:


gc:
```INSERT INTO users(id)
SELECT id FROM new_users WHERE id NOT IN (SELECT id FROM users);```

PR #4959, opened 2023-02-19
Re-add BSO 2x clue boost

ghosts-are-spooky:
### Description:

Added bso 2x boost back in

### Changes:

Copied back from old branch the boost, tested the times with new items and it's as expected

### Other checks:

### DIFF
--- a/src/mahoji/commands/clue.ts
+++ b/src/mahoji/commands/clue.ts
@@ -108,6 +108,9 @@ export const clueCommand: OSBMahojiCommand = {
 			(user.user.openable_scores as ItemBank)[clueTier.id] ?? 1
 		);
 
+		timeToFinish /= 2;
+		boosts.push('👻 2x Boost');
+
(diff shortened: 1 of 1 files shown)


ghosts-are-spooky:
@gc this update seems to match what is expected, I simply added the boost that was there from before. Only thing BSO specific to note is now GMs get 2 new additional boosts:
Ornate jewellery box (10%) > fancy/basic box (5%)
max cape (10%) > construction cape (6%)

So at 81% duration of what it was before. Some examples below:
:whiteMinion: Your minion is now completing 8x Grandmaster clues, it'll take around 35 minutes, 48 seconds to finish.
:whiteMinion: Your minion is now completing 8x Grandmaster clues, it'll take around 33 minutes, 56 seconds to finish.
:whiteMinion: Your minion is now completing 8x Grandmaster clues, it'll take around 32 minutes, 4 seconds to finish.

GMs would have been around 37-39 before.

PR #4961, opened 2023-02-20
Bso clues

ghosts-are-spooky:
### Description:

Updated the order of calculations for boosts, max trip time, moved previous flat boosts into global boosts, adjusted GM individually to match previous times as close as possible.

### Changes:

Removed graceful boost
Moved achievement diary cape out of flat boost into grandmaster, this keeps it with GM and removes double boost for elite master etc.
Moved clue hunter gear and 2x boost to global boosts
Because there is now an additional 10% boost applied to GMs, I set the time per clue to 10% above the previous value.
Moved all the duration calculations above where the max quantity per trip is calculated. This was causing clue trips to be <30 mins.

### Other checks:

I've tested but want to put this on test server for others to test.

### DIFF
--- a/src/lib/clues/clueTiers.ts
+++ b/src/lib/clues/clueTiers.ts
@@ -143,7 +143,7 @@ export const ClueTiers: ClueTier[] = [
 		table: GrandmasterClueTable,
 		id: 19_838,
 		scrollID: 19_837,
-		timeToFinish: Time.Minute * 32.3,
+		timeToFinish: Time.Minute * 32.23,
 		mimicChance: false,
 		allItems: GrandmasterClueTable.allItems,
 		stashUnits: masterStashes,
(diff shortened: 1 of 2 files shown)


src/mahoji/commands/testpotato.ts
```diff
@@ -235,6 +235,50 @@ for (const tier of allStashUnitTiers) {
 	allStashUnitItems.add(tier.cost.clone().multiply(tier.units.length));
 }
 
+const allClueBoostItems = new Bank()
+	.add("zulrah's scales", 50_000)
+	.add('dragon dart', 50_000)
```
TastyPumPum, line 240: Casing on these two items

src/mahoji/commands/testpotato.ts
```diff
@@ -235,6 +235,50 @@ for (const tier of allStashUnitTiers) {
...
+	.add('Bolt of cloth', 20)
+	.add('Gold Leaf', 20)
+	.add('Ring of Dueling(8)', 20)
```
TastyPumPum, line 243: Also on these two

src/mahoji/commands/testpotato.ts
ghosts-are-spooky: removing as should go on master anyway

PR #4963, opened 2023-02-20
Remove lowest value toa drops from #notifications

osDreamscape:
### Description:

Replaced the placeholder scrolls from CoX in purpleButNotAnnounced with the 2 lowest values from ToA to limit spam in #notifications in main channel

### Changes:

Removed Dex and Arcane from const purpleButNotAnnounced that was copied over from CoX activity and replaced with Lightbearer and Elidinis' ward.

### Other checks:

### DIFF
--- a/src/tasks/minions/minigames/toaActivity.ts
+++ b/src/tasks/minions/minigames/toaActivity.ts
@@ -17,7 +17,7 @@ import resolveItems from '../../../lib/util/resolveItems';
 import { updateBankSetting } from '../../../lib/util/updateBankSetting';
 import { userStatsUpdate } from '../../../mahoji/mahojiSettings';
 
-const purpleButNotAnnounced = resolveItems(['Dexterous prayer scroll', 'Arcane prayer scroll']);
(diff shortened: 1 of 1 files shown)


TastyPumPum:
Isn't the reason dex and arcane are removed due to them being the most common, not the value?

So really imo this PR should be fang and lightbearer ring or all three of them

osDreamscape:
I considered fang instead, but it's reasonably valuable (45m), provides a big boost for toa, and at least ingame it's usually a hype drop. Ward is none of those, albeit it being about twice as rare (but still more common than masori/staff)

I'd be fine with whatever changes, the purpose is just to limit the notifications channel sends as it's currently about 90% toa drops.

gc:
Yea it should be the most common ones, so the rarest things are announced.

TastyPumPum:
Magna has done this so closing it [#7b52e1d](https://github.com/oldschoolgg/oldschoolbot/commit/7b52e1df4957974843547444433fb571183c1e21)

PR #4974, opened 2023-02-25
Fix Hammy's appearing needlessly

themrrobert:
### Description:

Fix issue with Hammy appearing instead of Abyssal Pearls.

The issue was the code was doing a blanket check for files < 200 bytes, however Abyssal pearls (and other images can be, too) are less than 200 bytes.

### Changes:

- If the image is less than 200 bytes, look for a PNG header, only if that fails do we return a placeholder.

### Other checks:


gc:
No longer needed

PR #4996, opened 2023-03-11
Ability to do several raids per TOA trip

gc:
### Other checks:

### DIFF
--- a/src/lib/simulation/toa.ts
+++ b/src/lib/simulation/toa.ts
@@ -1,5 +1,5 @@
 import { SimpleTable } from '@oldschoolgg/toolkit';
-import { UserStats, XpGainSource } from '@prisma/client';
+import { Minigame, UserStats, XpGainSource } from '@prisma/client';
 import { bold } from 'discord.js';
 import {
 	calcPercentOfNum,
@@ -15,7 +15,8 @@ import {
 	round,
 	scaleNumber,
 	sumArr,
-	Time
+	Time,
+	uniqueArr
 } from 'e';
 import { CommandResponse } from 'mahoji/dist/lib/structures/ICommand';
 import { Bank, LootTable } from 'oldschooljs';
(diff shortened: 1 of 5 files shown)


src/lib/simulation/toa.ts
```diff
@@ -1155,46 +1169,75 @@ export async function toaStartCommand(
...
+		users.map(async i => ({ user: i, userStats: await i.fetchStats(), minigameScores: await i.fetchMinigames() }))
+	);
+	const baseDuration = createTOATeam({
+		team: toaSimUsers,
+		raidLevel,
+		quantity: 1
+	})[0].duration;
+	const maxTripLength = Time.Hour * 2; // Math.max(...users.map(i => calcMaxTripLength(i, activity_type_enum.TombsOfAmascut)));
+	const maxQuantity = clamp(Math.floor(maxTripLength / baseDuration), 1, 5);
+	const quantity = clamp(quantityInput ?? 1, 1, maxQuantity);
```
themrrobert: shouldn't this be
```ts
const quantity = clamp(quantityInput ?? maxQuantity, 1, maxQuantity);
```
so that it automatically chooses the max number possible and you don't have to guess

src/lib/simulation/toa.ts
```diff
@@ -1155,46 +1169,75 @@ export async function toaStartCommand(
...
+
+	console.log({
+		maxTripLength: formatDuration(maxTripLength),
+		baseDuration: formatDuration(baseDuration),
+		maxQuantity,
+		quantity,
+		quantityInput: quantityInput ?? 1
```
themrrobert: I would probably make this `quantityInput: quantityInput ?? 'NULL"` so that we know if no quantity was specified

src/mahoji/commands/raid.ts
```diff
@@ -120,6 +120,14 @@ export const raidCommand: OSBMahojiCommand = {
 							required: false,
 							min_value: 1,
 							max_value: 8
+						},
+						{
+							type: ApplicationCommandOptionType.Integer,
+							name: 'quantity',
+							description: 'The quantity to do (Defaults to 1).',
```
themrrobert, line 127: default to max now

themrrobert (changes requested):

src/mahoji/commands/raid.ts
osrsjobbo: Ya I'd agree to default to max

PR #4998, opened 2023-03-11
Add extra checks to PR template

osrsjobbo:
### Description:

Tweak the template for PRs to add frequent extra steps (e.g. lint & checking tests).

Tests will fail because they're failing on master atm.

### Changes:

See above. 

### Other checks:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,4 +8,6 @@
 
 ### Other checks:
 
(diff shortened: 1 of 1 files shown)


gc:
These just aren't needed, you don't really even need to run these, the PR CI checks will fail if the code isnt linted or the tests dont pass.

Issue #4983, opened 2023-02-27
Do not show "Auto Farming Contract" button when the required patch is growing


Redquaker:
## 🚀 Feature Proposal

Consider the following case for example:

- You receive a farming contract for _Dragonfruit trees_
- Your fruit tree patches are currently occupied with _Palm trees_

In such a case, the "Auto Farming Contract" button will still show up on a minion status check AND on every trip return, even though you cannot yet advance the required contract via planting/harvesting. Therefore, the button provides no useful info when a case like this occurs. Since it shows up regardless, it can no longer be used to determine when you can perform any action and the player has to check patches manually until the patch is ready.

It would be great if the "Auto Farming Contract" button would not show up if the required patch is currently growing _anything_. It currently does not show while the correct plant is growing, but not the incorrect plant.

Issue #4971, opened 2023-02-25
Auto farming contract button when planting same plant in contract


Redquaker:
## 🐛 Bug Report

The "Auto Farming Contract" button supposedly doesn't appear when you plant the same thing you harvest for a contract. Since this button both opens the seed pack and gets the next contract, you now have to type the contract command. Would be nice to fix.

See below for an example.

![image](https://user-images.githubusercontent.com/59510368/221374361-ed0a7586-d272-4254-bd40-65b06f63346c.png)

Issue #4958, opened 2023-02-19
Add toggle to Autofarming contract button


mylife212:
## 🚀 Feature Proposal

Title. Just like BH buttons being toggled on/off via `user config`, this should be added too for those who don't do contracts anymore - as the button can be a nuisance.

Issue #4956, opened 2023-02-18
ask for easier contract button


turb042o:
## 🐛 Bug Report

It's super nice the ask for easier contract button has been added but it shows up at the wrong time, any chance to get it to appear as a selection when initially receiving the new contract instead of after we manually send the / command for easier contract?

![Screenshot_20230218_061145_Discord.jpg](https://user-images.githubusercontent.com/90441459/219865296-5b0350b3-2d35-4631-9fc1-03b24765c018.jpg)

https://user-images.githubusercontent.com/90441459/219947969-9f055bd5-0778-4061-b352-44b339a912ce.mp4

PR #5001, opened 2023-03-12
Farming contract improvements

osrsjobbo:
### Description:

This closes: 
- https://github.com/oldschoolgg/oldschoolbot/issues/4983
- https://github.com/oldschoolgg/oldschoolbot/issues/4956
- https://github.com/oldschoolgg/oldschoolbot/issues/4958
- https://github.com/oldschoolgg/oldschoolbot/issues/4971

Thanks to @Redquaker for helping me test his bugs.

### Changes:

- Reimplemented `canRunAutoContract` to fix a bug whereby it would return true in the event a patch was planted with something which wasn't your contract, regardless of it being finished. Updated the method to look for the patch type rather than the specific plant to fix this.
- Pulled in components when merging responses. 
- Added a toggle for farming button to display or not similar to birdhouses as a result of lots of user feedback requesting as such (approx 4-5 issues since I added it)

### Other checks:


turb042o:
any way you can look into what is allowing auto farming contract button (and only this button) to work in channels (bso for example) that osb is not able to be interacted with? or is that a integration issue?

Issue #4957, opened 2023-02-18
Add clue button at ToA


Redquaker:
## 🚀 Feature Proposal

When you get an elite clue at ToA, there's no clue button similar to when you get one through /k, birdhouse, etc.

![image](https://user-images.githubusercontent.com/59510368/219884465-435b5fdf-8241-47d3-b0aa-77375c78484d.png)

osDreamscape:
+1

PR #5002, opened 2023-03-12
Allow clue buttons to show when soloing toa

osrsjobbo:
### Description:

Allow clue button to show when included in loot at TOA. 

I think there is a risk of doubling up on GP tracking, so I tried to handle that, but there might be a nicer way of doing that.

Closes https://github.com/oldschoolgg/oldschoolbot/issues/4957

### Changes:

- Pass loot variable into handleTripFinish **only when soloing.**

### Other checks:


osrsjobbo:
@gc What changes would you like to see here?

gc:
something which doesnt require that duplication. we need a more elegant/maintainable solution, im not really sure what that would be just looking at it.

PR #5018, opened 2023-03-20
Add the crafting of Splitbark Armour

### LINKED ISSUE #4940: Splitbark crafting
## 🚀 Feature Proposal

Shades has been out a while now, it would be nice to be able to create split bark armour via crafting this would give access to splitbark boots (the only item not available from the set in the bot)

It also gives all the fine cloth a use.

Blood bark and swamp bark could be added in with creatables.

TastyPumPum:
Adds the crafting of splitbark armour, we can get the relevant materials through shades of Mort'ton and have Bloodbark creating in the game with no way to get the splitbark boots, this resolves that issue.

Closes #4940


src/lib/skilling/skills/crafting/craftables/misc.ts
```diff
@@ -88,6 +88,46 @@ const Misc: Craftable[] = [
 		inputItems: new Bank({ 'Xerician fabric': 5 }),
 		tickRate: 3
 	},
+	{
+		name: 'Splitbark gauntlets',
+		id: itemID('Splitbark gauntlets'),
+		level: 60,
+		xp: 62,
+		inputItems: new Bank({ Bark: 1, 'Fine cloth': 1 }),
```
gc, line 96: From now forward, we no longer want to use this format of passing in names to the constructor, because we want to eventually get rid of it. instead, do: `new Bank().add().add()`

gc (changes requested):

PR #5079, opened 2023-04-20
Ironman price nerf for Bronze axe/pickaxe

mylife212:
The Ironman price for these items were 1gp, while their GE price is 77~ gp. Meaning you could exploit the low price by buying a million pickaxes and sacrificing them for essentially 77x the value. By bringing these prices to 100gp, it fixes that loophole. I'm sure there are other items that can be exploited in a similar way, but these are the biggest offenders by far

### Description:

- Set ironmanPrice for Bronze Pickaxe/Axe from 1gp to 100gp.

### Other checks:


mylife212:
Closing because going to make a bigger PR that goes thru all items that can be abused like this.

PR #5156, opened 2023-06-06
Allow minion to clean herbs on herb runs

### LINKED ISSUE #5061: Cleaning herbs while doing herb run
I was wondering if it would be possible to be able add the ability to clean the herbs as you do a herb run. This is as far as i know how most people do it ingame anyway. Maybe just add like a 10 second timer to the farming trip when you plant herbs so the minions has time to clean them whilst running between the farming patches. 

This would be a good qol update cause you dont have to spend time to clean them/spend money to make zahuur clean them for 200 gp each.

TastyPumPum:
- Adds a bitfield for cleaning herbs during farm runs. This can be toggled by the user.
- Replaces the grimy herbs with the clean herbs in the loot.
- Adds the relevant herblore xp on for cleaning them.
- Only cleans if the user has a high enough herb level.

~~Adds 4.8 seconds per patch onto the herb run. (An average yield of 8 herbs at 1 tick per cleaning)~~ Removed due to cleaning during running instead.

One potential issue is that it doesn't add to the personal herblore stats due to it all being in the farming activity.

Closes #5061 


src/tasks/minions/farmingActivity.ts
```diff
@@ -167,8 +168,17 @@ export const farmingTask: MinionTask = {
 			}
 			checkHealthXp = alivePlants * plantToHarvest.checkXp;
 
+			const cleanHerb =
```
gc, line 171: `cleanHerb` is ambiguous, booleans should almost always start with something like is/was/has/should/did, `shouldCleanHerb` sounds good here

src/tasks/minions/farmingActivity.ts
```diff
@@ -167,8 +168,17 @@ export const farmingTask: MinionTask = {
...
+				user.bitfield.includes(BitField.CleanHerbsFarming) &&
+				user.skillLevel(SkillsEnum.Herblore) >= plantToHarvest.herbLvl!;
+
 			if (plantToHarvest.givesCrops) {
-				if (!plantToHarvest.outputCrop) return;
+				let cropToHarvest = plantToHarvest.outputCrop;
+				if (cleanHerb) {
+					cropToHarvest = plantToHarvest.cleanHerbCrop;
+				}
+				if (!cropToHarvest) return;
```
gc, line 181: In what scenario, or why, are we returning here? Is it an error-case? if so, it should throw an error, or have a test.

src/tasks/minions/farmingActivity.ts
```diff
@@ -299,6 +320,10 @@ export const farmingTask: MinionTask = {
 				skillName: SkillsEnum.Woodcutting,
 				amount: Math.floor(woodcuttingXp)
 			});
+			await user.addXP({
+				skillName: SkillsEnum.Herblore,
+				amount: Math.floor(herbloreXp)
+			});
```
gc, line 326: We should add a xp gain source, for like "Cleaning herbs while farming". The purpose being so we can show users: you gained X amount of xp from cleaning herbs while farming.

gc (changes requested):

PR #5165, opened 2023-06-15
Fix BSO Elder runes bonus quantity

TastyPumPum:
Updates the raiments bonus to be based of the correct quantity for elder runes

~~Unable to test this rn. But reviewing the code it should all be good.~~ Have now tested and it seems good


themrrobert:
You will have to rebase this after my `bso merge` PR is merged to fix the tests.

PR #5174, opened 2023-06-22
Add buyable seed packs

### LINKED ISSUE #2890: Allow spirit seeds to be traded in at guildmaster jane for tier 5 seed packs
## 🚀 Feature Proposal - spirit seeds being offered at guildmaster jane

From the [Wiki page for spirit seeds](https://oldschool.runescape.wiki/w/Spirit_seed), it says "As there are limited spots the player can plant spirit trees, any excess spirit seeds the player might have can be traded in at Guildmaster Jane for tier 5 Seed packs which are worth an average of 101,452.09. Tier 5 seed packs have an approximately 1/22 chance of containing another spirit seed which can itself be traded in for an additional pack. Meaning the average value of a seed pack when trading in all spirit seeds obtained for additional packs converges to the slightly higher value of 106,206.95."

this would be useful for irons and mains on the bot as it would allow them to get some useful seeds they from some otherwise potentially useless seeds. 

Kind Regards

LuxrayElite:
### Description:

This is a rehash of a very old PR which I did. #3353

Which also closes & resolves #2890

Allows you to exchange spirit seeds for tier 5 seed packs
Loot is auto-added to the bank so it doesn't mess with getting a new seed pack
Nor depend on the current contract seed quality
Loot isnt disclosed until the user purchases the items to prevent rerolling.

### Changes:

Added a new abstracted buy command
Allow the exchanging of spirit seeds for seedpacks
Generates a bank image if exchanging over 5 seeds at once

### Other checks:

Adds to their openables total and associated leaderboard
Doesnt impact contracts completed
Checked with and without favour
Checked with and without farming level
Checked over buying


gc (changes requested):
Don't we need to add this to their openables scores? Perhaps try calling `finalizeOpening` directly.

LuxrayElite:
> Don't we need to add this to their openables scores? Perhaps try calling `finalizeOpening` directly.

I chose to just add it to the openables score via `addOpenablesScore` due to not requiring any additional modification of functions.

If I were to use `finalizeOpening`, it would have made all the bank image and text but would end up being a multiple bank item transactions ,in order to generate the correct bank image, which I think could be more prone to bugs. 

I would have to give them the seedpacks first in order to use `finalizeOpening` without any modifications.

Seedpack items are tied to the last farming contract completed.

Hopefully this is an acceptable solution :)

hispazn:
Someone merge this >:O

PR #5186, opened 2023-06-26
Fix tithe farm leaderboards

### LINKED ISSUE #5122: Tithe Farm 0kc & nobody on LB [BSO&OSB]
## 🐛 Bug Report

it seems everyone's `/minion kc name:tithe farm` says 0 on both bots causing the lb to say nobody is on it

but it shows completions in the cl

LuxrayElite:
### Description:

Resolves #5122

Whilst this "fixes" the issue it doesnt resolve the underlying problem.  We currently have two database variables for storing tithe farm completions, one which isnt utilized at all but was the primary fetch for the KC & LB command.

I did consider using the tithe_farm database entry within the minigames table but felt it could potentially introduce some database bugs. (The implemented fix would consist of 5 lines of code in a single file than these two file changes)

### Changes:

Point the leaderboard to the current place where tithefarmcompletions are stored
Prevent minion KC looking at the same wrong place and getting caught as a "Minigame"
Restrict the new fetch to require 10 completions and log the first 10 entries similar to other minigames
### Other checks:

### DIFF
--- a/src/mahoji/commands/leaderboard.ts
+++ b/src/mahoji/commands/leaderboard.ts
@@ -193,7 +193,7 @@ async function minigamesLb(user: MUser, channelID: string, name: string) {
 					   FROM user_stats
 					   WHERE "tithe_farms_completed" > 10
 					   ORDER BY "tithe_farms_completed"
-					   DESC LIMIT 2000;`
+					   DESC LIMIT 10;`
 		);
 		doMenu(
 			user,
--- a/src/mahoji/commands/leaderboard.ts
+++ b/src/mahoji/commands/leaderboard.ts
@@ -187,28 +187,50 @@ async function minigamesLb(user: MUser, channelID: string, name: string) {
 		return `That's not a valid minigame. Valid minigames are: ${Minigames.map(m => m.name).join(', ')}.`;
 	}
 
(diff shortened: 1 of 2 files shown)


src/mahoji/commands/leaderboard.ts
```diff
@@ -187,28 +187,50 @@ async function minigamesLb(user: MUser, channelID: string, name: string) {
...
+			user,
+			channelID,
+			chunk(titheCompletions, LB_PAGE_SIZE).map((subList, i) =>
+				subList
+					.map(({ id, amount }, j) => `${getPos(i, j)}**${getUsername(id)}:** ${amount.toLocaleString()}`)
+					.join('\n')
+			),
+			'Tithe farm Leaderboard'
+		);
+	} else {
```
gc, line 208: add a return after the doMenu here, so the existing code can be unchanged.

gc (changes requested):

src/mahoji/commands/leaderboard.ts
LuxrayElite: Ive changed it back now.

I only initially wrapped it in an else statement to prevent repeating the return line

PR #5192, opened 2023-07-02
Add ability to use super attack and super strength if super combat is not in user's bank for TOA

KennySchl:
### Description:

Added logic to check if user has a super combat, if available add that to `minimumSuppliesNeeded` and then continue like before. Else add a super attack and super strength and then continue like before with all of the requirement checks.

### Changes:
Changes in toa.ts only. To minimum supplies and condition check before checking minimumSuppliesNeeded against the bank. As well as adding the supplies to the cost accordingly.

### Other checks:

### DIFF
--- a/src/lib/simulation/toa.ts
+++ b/src/lib/simulation/toa.ts
@@ -328,7 +328,7 @@ const toaRequirements: {
 	{
 		name: 'Supplies',
 		doesMeet: ({ user, quantity }) => {
-			if (user.owns(new Bank().add('Super combat potion(4)', 1))) {
+			if (user.owns('Super combat potion(4)')) {
 				minimumSuppliesNeeded = minSuppliesWithScmb;
 			} else {
 				minimumSuppliesNeeded = minSuppliesWithAtkStr;
--- a/src/lib/simulation/toa.ts
+++ b/src/lib/simulation/toa.ts
@@ -183,8 +183,7 @@ const minTOAStats: Skills = {
 const minimumSuppliesNeeded = new Bank({
 	'Saradomin brew(4)': 10,
 	'Super restore(4)': 5,
-	'Ranging potion(4)': 1,
-	'Super combat potion(4)': 1
+	'Ranging potion(4)': 1
 });
 
 const miscBoosts = [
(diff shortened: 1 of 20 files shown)


LuxrayElite:
Just some feedback from testing which I've done:

/raid toa help
- The default value keeps incrementing each time which you run the command
- The first call requires 1x super attack, 1 super strength then increments by +2 each subsequent call
- If you own the super combat potion it doesn't/ state you have the correct supplies

image
After spawning super combats
image
image

If you run /toa start invocation: 450
It is checking for the incremented value of potions and if you dont have enough increments it further.

image
image

src/lib/simulation/toa.ts
```diff
@@ -320,6 +319,12 @@ const toaRequirements: {
 	{
 		name: 'Supplies',
 		doesMeet: ({ user, quantity }) => {
+			if (user.owns(new Bank().add('Super combat potion(4)', 1))) {
```
LuxrayElite, line 322: The minimum amount of supplies needed should be static, this increments the amount with each toa command that is run

src/lib/simulation/toa.ts
```diff
@@ -953,7 +958,12 @@ async function calcTOAInput({
 }> {
 	const cost = new Bank();
 	const kc = kcOverride ?? (await getMinigameScore(user.id, 'tombs_of_amascut'));
-	cost.add('Super combat potion(4)', quantity);
+	if (minimumSuppliesNeeded.has('Super combat potion(4)')) {
+		cost.add('Super combat potion(4)');
```
LuxrayElite, line 962: Missing quantity here

src/lib/simulation/toa.ts
```diff
@@ -953,7 +958,12 @@ async function calcTOAInput({
...
+	} else if (minimumSuppliesNeeded.has('Super attack(4)') && minimumSuppliesNeeded.has('Super strength(4)')) {
```
LuxrayElite, line 963: Doesn't necessarily require an else if, I think an else would suffice here. 
If neither are present what would be the cost for example?

src/lib/simulation/toa.ts
KennySchl: Once I saw the main comment I figured this would be the issue will take a second look at this and make sure it's static instead of adding each time

src/lib/simulation/toa.ts
KennySchl: My thoughts here were the same. Pretty much checking if either are available before adding to the cost in order to avoid null errors but it should return higher up before if gets here if neither is available anyways.

KennySchl:
Appreciate the quick feedback. Will make those changes!

src/lib/simulation/toa.ts
LuxrayElite: btw you can just use `user.owns('item name')` instead of `user.owns(new bank())`

PR #5200, opened 2023-07-06
Adds "Top Sacrificer" role for Rank 1 Ironman Sac Value

themrrobert:
### Description:

Gives Top Sac role to Rank 1 ironman.

### Other checks:

- [] I have tested all my changes thoroughly.


gc:
FYI just not sure if I want this as a separate ironman thing. It doesnt effect the economy and it isnt something i particularly want to encourage, and isnt in itself that impressive either I think.

LuxrayElite:
There are currently 3 ranks for each bot assigned to non-ironmen for having the most value sacrificed. It is quite common for people to swap gp between the two bots, and people can be donated items.  (which seems less impressive)

Generally the most value sacrificed by an ironman reflects their dry-streaks such as Nex, Corp, and others. 

An ironman will never be able to compete with an account that can trade similar to the pr which added most unique sacrificed items an ironman roll which likely gave precedence to this pr.

Currently rank 1 sac value (ironman)  rank 10 sac value (overall) has 13k nex kc without a nexling rather than farming gp at a boss.

themrrobert:
Yes, and @gc said we could add it for ironman when I asked a while ago, as long as it was only for Rank 1.

And considering I would personally be affected by this PR, I want it, I definitely worked hard to earn and maintain rank 1 sac value.

PR #5207, opened 2023-07-12
Karambwan and wine update

Lajnux:
### Description:
Updated rates for karambwan and wine making to be more accurate to ingame techniques.

### Changes:
480k xp/hr for wines
915k xp/hr for karambwan at lvl 92 forwards.

### Other checks:


gc:
You have extraneous changes to the creatables.txt file.

LuxrayElite (commented):
Is there a reason you split 2.85 into 2.4 and 0.45 instead of just changing it from 2.88 to 2.85

Lajnux:
> Is there a reason you split 2.85 into 2.4 and 0.45 instead of just changing it from 2.88 to 2.85

yea, makes it easier to see what is the 4 ticks action vs the extra added time.

PR #4909, opened 2023-02-01
Fix armadyl plate/component clash

ghosts-are-spooky:
### Description:
Fix bug where you couldn't make armadyl plate on BSO

### Changes:
Updated the clashing create item names, added either (Components) or (Plate) so it's clear what you get.

### Other checks:

-   [x] I have tested all my changes thoroughly.

PR #5213, opened 2023-07-17
Fix Revert to components not adding CL (BSO)

themrrobert:
### Description:

- Reverting Masori/Armadyl/Ancestral/Bandos materials on BSO doesn't add to CL
- Not possible to revert Armadyl into plates

### Changes:

- Adds forceAddToCl on all revert to components
- Adds distinguishers on the Armadyl reverts so you can actually revert to plates, and know which you're doing.

### Other checks:

### Notes:

The Armadylean plates do NOT have `forceAddToCL` added because that needs to be merged from OSB; so just be mindful when merging OSB => BSO that in case there is a conflict, we need to ensure `forceAddToCL` is merged to BSO.


LuxrayElite:
Extends/ resolves #4909

PR #5283, opened 2023-08-26
Updated item ids for DT2 Rings

DarkWorldsArtist:
Updated item ids from the beta world variants (untradeable) to the main game variants (tradeable). This does not address the issue of previously acquired rings being untradeable.


gc:
These files aren't supposed to be edited. The fix requires we do a change on all the database, and remove them from OSJS.

PR #5296, opened 2023-08-29
abyssal lantern

Lajnux:
### Description:

Added Abyssal lantern to buyables, recent update ingame

### Changes:

Added Abyssal lantern to buyables, recent update ingame

### Other checks:


gc:
Boosts?

PR #5327, opened 2023-09-07
rc fix + gp sink

Lajnux:
### Description:

Fixed logic for runecrafting and added another GP sink. 

### Changes:

Changed some logic so that runes that don't require stamina potions are not defaulting to true.
Also added a trillion GP sink in the form of runners, which is a common training form for specifically runecrafting (40m gp/h, 2x ingame prices). Resulting in around 40B gp drained per 200M rc xp. 

### Discussion:
A lot of discussion about runners was brought up, both positive and negative. Which I try to summarize here.

Pros: 
It's a massive GP sink. 200M rc would cost 40B gp (2.7B gp for lvl 99) . Only 6 ppl in OSB have the cash on hand for 200M xp.  Making it a substantial sink if used both for 99 and 200M. 
It could be a direct sink or go towards other ppl, the down side with other ppl is the convenience factor, where ppl wouldn't bother spending more gp for increased xp if it also result in them having to find other ppl.
It is one of few skills in bot that doesn't have have EHP method implemented yet which goes in line with previous Skilling updates. 
It also goes well with other contents where we decided to use none real minions to assist in various activities from raids, shooting star, volcanic mine and guardians of the rift. 

Cons:
Concerns regarding relating none real minions with alts, which could open up possibilities of alt methods.
People who already did 200M xp feel like it's unfair if others get faster methods, This is something that happens in osrs all the time.
A few people raised that it's against the spirit of the bot, which might be related to the concerns that they relate none real minions with alts.
Concerns regarding the actual gp sink amount was brought up, both ppl thinking it is too cheap with 40B and others thinking it's too expensive to the point where none would use it or afford it. 

### Other checks:


J4stMart:
my man

src/mahoji/commands/runecraft.ts
```diff
@@ -71,9 +77,15 @@ export const runecraftCommand: OSBMahojiCommand = {
...
+	}: CommandRunOptions<{
+		rune: string;
+		quantity?: number;
+		usestams?: boolean;
+		runners?: boolean;
+		daeyalt_essence?: boolean;
+	}>) => {
 		const user = await mUserFetch(userID.toString());
-		let { rune, quantity, usestams, daeyalt_essence } = options;
+		let { rune, quantity, usestams, runners, daeyalt_essence } = options;
```
TastyPumPum, line 89: Needs to be an ironman check somewhere and force runners to false if the minion is ironman (also this is a bad pr)

src/mahoji/commands/runecraft.ts
```diff
@@ -122,46 +134,55 @@ export const runecraftCommand: OSBMahojiCommand = {
...
-			boosts.push('3% for Runecraft cape');
+			if (
+				user.skillLevel(SkillsEnum.Runecraft) >= 99 &&
+				user.hasEquippedOrInBank('Runecraft cape') &&
+				inventorySize > 28
+			) {
+				tripLength *= 0.97;
+				boosts.push('3% for Runecraft cape');
+			}
+		} else if (runners) {
```
LuxrayElite, line 183: Doesnt require Else if (runners)

else would suffice

src/mahoji/commands/runecraft.ts
```diff
@@ -122,46 +134,55 @@ export const runecraftCommand: OSBMahojiCommand = {
 
 		let { tripLength } = runeObj;
 		const boosts = [];
-		if (userHasGracefulEquipped(user)) {
-			tripLength -= tripLength * 0.1;
-			boosts.push('10% for Graceful');
+		if (user.isIronman && runners) {
+			runners = false;
+			boosts.push('Ironmen stand on their own');
```
LuxrayElite, line 139: Not really necessary to push it to boosts and doesnt really explain why this message is present.

I understand that its because it involves trading but maybe more useful to say "Ironmen can't utilize runners." and not include it in the boost string or throw a generic return like most Ironman clauses.

src/mahoji/commands/runecraft.ts
Lajnux: does that work even if runners can be undefined?

src/mahoji/commands/runecraft.ts
Lajnux: fixed

src/mahoji/commands/runecraft.ts
```diff
@@ -200,6 +221,17 @@ export const runecraftCommand: OSBMahojiCommand = {
 
 		const totalCost = new Bank();
 
+		const userGP = user.GP;
+		if (runners) {
+			if (userGP < (duration / Time.Hour) * 40_000_000) {
```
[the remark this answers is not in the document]
Lajnux: good feedback, fixed

src/mahoji/commands/runecraft.ts
Lajnux: resolved

src/mahoji/commands/runecraft.ts
```diff
@@ -122,46 +134,55 @@ export const runecraftCommand: OSBMahojiCommand = {
 
 		let { tripLength } = runeObj;
 		const boosts = [];
-		if (userHasGracefulEquipped(user)) {
-			tripLength -= tripLength * 0.1;
-			boosts.push('10% for Graceful');
+		if (user.isIronman && runners) {
+			runners = false;
```
TastyPumPum, line 138: Any point in this if it's just returning?

gc:
If you want to add runners, we will need to discuss it in a #osb-discussion thread first. Put your arguments/justification for it. I personally don't know if it's a good idea, unless you're paying other actual (non-alt-account) players' minions to run for you.

Lajnux:
Information added in the discussion section

gc:
I'm taking the review label off this until we've reached a consensus agreement in discord. Currently, I don't think there is one other person who supported the idea. So it needs to be changed or you need to more work on convincing.

gc:
Closing unless we reach some consensus decision in discord, and since there's no chance this will be merged as is without more information/discussion/etc.

PR #5341, opened 2023-09-15
Fix "Ancient blood ornament kit" drop rate

TastyPumPum:
Currently you have to do 5 Awakened kills to get the ancient blood ornament kit, this updates it so you can get it on the 4th kill.


gc:
Just based on reading the code, this seems like it would give it on the 1st kill? I could be wrong

TastyPumPum:
> Just based on reading the code, this seems like it would give it on the 1st kill? I could be wrong

It worked in testing
Tried killing 7 of one awakened boss caused no blood kit
Tried killing 1 of each and gave blood kit on last kill
Tried killing 1 of each and then sacrificing the kit and getting another kill and gave a kit again

PR #5342, opened 2023-09-15
Adds Forgotten Brews

TastyPumPum:
Allows the reverting of the ancient icon and the venator shard to get ancient essence. Adds forgotten brews as potions.

https://oldschool.runescape.wiki/w/Forgotten_brew#4_dose
https://oldschool.runescape.wiki/w/Ancient_essence


src/lib/skilling/skills/herblore/mixables/potions.ts
```diff
@@ -400,6 +400,15 @@ const Potions: Mixable[] = [
 		tickRate: 2,
 		bankTimePerPotion: 0.5
 	},
+	{
+		item: getOSItem('Forgotten brew (4)'),
+		aliases: ['forgotten brew (4)', 'forgotten brew', 'forgotten'],
+		level: 91,
+		xp: 145,
+		inputItems: new Bank({ 'Ancient brew (4)': 1, 'Ancient essence': 80 }),
```
gc, line 408: You dont have to change it, but for future reference, try not use this format for making banks anymore. do `new Bank().add().add()`

src/lib/data/createables.ts
```diff
@@ -726,6 +726,26 @@ const Reverteables: Createable[] = [
...
 		noCl: true
 	},
+	{
+		name: 'Revert ancient icon',
+		inputItems: {
+			[itemID('Ancient icon')]: 1
+		},
+		outputItems: {
+			[itemID('Ancient essence')]: 5000
+		},
```
gc, line 736: You dont have to change it, but for future reference, these should be just banks. `new Bank().add()`

PR #5374, opened 2023-09-23
Fix Nex Combat Achievements in BSO

ghosts-are-spooky:
### Description:
BSO specific nex fixes, merging to BSO as it's updating NEX_ID const to fix CAs that are working in OSB

### Changes:

Updated NEX_ID const to be correct id
Fixed Nex Survivors

### Other checks:


gc:
Changing this ID will surely cause a few issues, id's are meant to never change. i.e all KCs will be reset to 0.

themrrobert:
I was going to say the same thing.

In this case, the combat achievements for Nex don't apply to BSO.

BSO Nex is completely separate content from OSB nex: OSB Torva doesn't exist in BSO, and the Nex Boss is completely different, and anything related to Nex in OSB cannot be transferred 1:1 to BSO.

Instead, new Combat Achievements for Nex should be written, or at the very least they should be changed to use the BSO Nex Identifier.

ghosts-are-spooky:
Re-tested again in full on the test bot server starting with 1 kc, 0 ca done. https://discord.com/channels/940758552425955348/1031143698752360478

Issue #5388, opened 2023-09-26
Shooting Stars not spawning


TastyPumPum:
Myself and several other members haven't seen a shooting star spawn in over a month, could be related to #5276 but not too sure.

PR #5389, opened 2023-09-26
Fix shooting stars not spawning

J4stMart:
### Description:

There were multiple reports of shooting stars not spawning anymore. 
Seems like this would fix it, ~~but it's untested!~~ and has been confirmed by @nwjgit


TastyPumPum:
Closes #5388

nwjgit:
I tested this and it would appear to fix shoot stars:

Discord_JFAuSqQc6r

PR #5397, opened 2023-09-29
Divination

gc:
### Other checks:


TURB042O:
I see this is from September, but I changed the divination cape, hood & cape (t) info in #5641

PR #5402, opened 2023-10-03
Forestry basket, Woodcutting boost message + fix

nwjgit:
1. Add Forestry basket to creatable items
2. Correct issue with +5min trip time not working when log basket is equipped.
3. Make forestry basket give same trip boost as log basket.
4. Add a boost message in the trip send message.


nwjgit:
Will re-post with Forestry Part 2 edits.

Issue #5007, opened 2023-03-14
Add a function to pick favourite bh seeds


TastyPumPum:
Like we have favourite alchs, add a way to pick a favourite seed for birdhouse runs and when you run out of the seed give a return message stating your seed has ran out. And to clear the favourite setting or pick a new seed etc. Also update the seeds we can use in birdhouses as there was a recent osrs update that allows any seed now. This would be very useful for irons and people that don't want to accidentally use herb seeds

PR #5409, opened 2023-10-04
Favourite bh seeds

Superwhammybar:
Closes https://github.com/oldschoolgg/oldschoolbot/issues/5007

### Description:

Adding the ability to favourite seeds to use for birdhouse runs so that precious seeds, like herbs, can be avoided.

### Changes:

Added a new favorite_bh_seeds array on the user to store favourited seeds
Added `/config user favorite_bh_seeds` to allow for modification of favourites
Birdhouse activity uses the favourites list, if configured, and won't send a trip if there aren't any favourited seeds available

### Other checks:


themrrobert (changes requested):

src/lib/skilling/skills/hunter/birdHouseTrapping.ts
```diff
@@ -178,4 +179,182 @@ const birdhouses: Birdhouse[] = [
...
+		amount: 10
+	},
+	{
+		name: 'Woad seed',
+		itemID: itemID('Woad seed'),
+		amount: 10
+	},
+	{
+		name: 'Limpwurt seed',
+		itemID: itemID('Limpwurt seed'),
```
gc, line 270: Item name shouldn't be kept like this, manually written out (its hard to maintain and prone to mistakes). Instead, just do `item: getOSItem("Limpwurt seed")`. and then you get the item.name and item.id.

Superwhammybar: Updated the code to use getOSItem and changed consumers to use `item.id`

src/lib/skilling/skills/hunter/birdHouseTrapping.ts
```diff
@@ -178,4 +179,182 @@ const birdhouses: Birdhouse[] = [
 	}
 ];
 
+export const birdhouseSeeds = [
+	{
+		name: 'Hammerstone seed',
```
gc, line 184: Im not sure if youre still working on it, but the names are still here.

Superwhammybar: Sorry, completely missed this part! I've removed the variable and referenced `item.name` instead.

Issue #5305, opened 2023-09-01
Kits on Scythe Disappear


shelixd:
## 🐛 Bug Report

Write a description of the bug here, give all the information you can about what it is, how it happened.

When your scythe of vitur runs out of charges and gets unequipped, if there is a kit on it (holy/sanguine) it disappears and the regular scythe gets put in the bank.

mylife212:
This is still a bug, lwy just lost a sang kit due to scythe running out of charges.

Lewis7777:
i got owned by this pls fix

PR #5422, opened 2023-10-11
return correct uncharged item when charges run out

nwjgit:
closes: https://github.com/oldschoolgg/oldschoolbot/issues/5305

### Description:

This addresses the issue of charged items reverting to their normal variant when minion charges reach zero. i.e. when you have a Sanguine Scythe of vitur and you run out of charges you are given a Scythe of vitur (uncharged) and the kit is lost.

### Changes:

Changes how the bot determines which item to give back to the user. This returns the correct unchanged version of the item that has reached zero charges.

### Other checks:

I have done extensive testing in my private server. It may be beneficial to move this over to the test server and let the community do more testing to ensure no bugs.


themrrobert (changes requested):
Too many problems as-is.

- the name `run` is too generic, it should be `breakOnRefund`
- The code is bugged, and won't remove the degraded item unless it also refunds an item
- The code doesn't take into consideration items degraded from the bank
- A union type should be used to avoid creating anonymous functions to return a constant value
- The constant Bank()'s should end with .freeze(), which i know isn't your bug, but this would be a good place to fix that.
- I don't think anonymous functions are the right solution here.... it should probably be some kind of `refundVariants` array which is easier to maintain:

Something like:
`refundVariants: [ ['Holy scythe', 'Holy scythe (uncharged)' ], [ 'Sang scythe', 'Sang scythe uncharged'] ];`

or maybe even
```ts
    refundVariants: [
        {
            variant: getOSItem('Holy scythe of vitur'),
            refund: new Bank().add('Holy scythe of vitur (uncharged')
        }, {
            variant: getOSItem('Sanguine scythe of vitur'),
            refund: new Bank().add('Sanguine scythe of vitur (uncharged)').freeze();
        }
    ],
```

src/lib/degradeableItems.ts
```diff
@@ -285,22 +325,22 @@ export async function degradeItem({
...
-				await user.addItemsToBank({ items: degItem.itemsToRefundOnBreak, collectionLog: false });
-			}
 		} else if (hasInBank) {
 			// If its in bank, just remove 1 from bank.
-			let itemsToAdd = undefined;
-			if (degItem.itemsToRefundOnBreak) {
-				itemsToAdd = degItem.itemsToRefundOnBreak;
+			if (degItem.run !== undefined) {
+				const itemsToAdd = degItem.run(user); // Call the function to get the actual Bank value
+				await user.transactItems({ itemsToRemove: new Bank().add(item.id, 1), itemsToAdd });
```
themrrobert, line 342: This is where the bug I mentioned is; notice it only removes the degraded item if there's an item to refund.

This may not always be the case if something degrades into dust, and should be accounted for, even if not used currently.

src/lib/degradeableItems.ts
```diff
@@ -125,7 +141,9 @@ export const degradeableItems: DegradeableItem[] = [
 	{
 		item: getOSItem('Amulet of blood fury'),
 		settingsKey: 'blood_fury_charges',
-		itemsToRefundOnBreak: new Bank().add('Amulet of fury'),
+		run: () => {
+			return new Bank().add('Amulet of fury');
```
themrrobert, line 145: There is no reason to use anonymous arrays here; these should remain constant Bank() objects, and a union type + type checking should be used to support both types.

nwjgit:
Thanks for all the feedback, I believe I made all the adjustments you wanted.

Something that may need addressed to is the "unchargedItem" not taking into account variants when trying to charge items. Currently it isn't an issue because the only to variant items are sang staff and scythe and you can just do /create and charge them without errors. But this won't work if for example the shadow gets a varient since it's charged version isn't under /create.
Discord_ZmBbh794K1

I imagine something could be done with /degradableItemsCommand similar to what I've done here but I didn't want to bog down the pr with more changes unless you want to put it all in one pr or make a new one.

nwjgit:
Could this be reviewed again? People are still losing their kits.

PR #5429, opened 2023-10-13
Rebalance wildy pk encounters and death chance

Lajnux:
### Description:

Added back the initial intended rates for pk and death chance but using the new wilderness experience overall

### Changes:

Added back the initial intended rates for pk and death chance but using the new wilderness experience overall

### Other checks:


gc:
What exactly does this do? And are people happy with it?

Lajnux:
> What exactly does this do? And are people happy with it?

It adds back the initial intended values for death rate, peak times and minimum item requirements which, I balanced the whole function around to make it accurate to ingame.

Lajnux:
> > are people happy with it?

People normally don't like negative implementations so I assume not, but it's balanced based of actual ingame rates. I don't know why the numbers was adjusted in the first place.

PR #5455, opened 2023-10-23
BSO - Dragon hunter crossbow/lance boost fix

TastyPumPum:
Unsure if it's intentionally this way, but the Dragon hunter lance/crossbow boost on BSO was 20% rather than the 15% that is on OSB. 

This PR puts the BSO boost back up to the 20%.


gc:
I cant think of any reason this would be different from osb.

gc:
Closing as its been 5-6 months without a response. It seems like the current state is the right state, and as I said above, I have no idea why it would be different in the first place.

PR #5459, opened 2023-10-24
Fix passive implings league task

themrrobert:
### Description:

- Currently the passive impling task will still 'succeed' if you haven't got any of a certain impling.

IE. If you only have 100 young implings, and 100 earth implings, but NONE others, it will still think you have >= 100 of each.

### Changes:

- Checks each impling instead of assuming the userStats passive impling bank represents all implings

### Other checks:

image


ghosts-are-spooky:
@themrrobert you changed the 100 of every impling task to require 100 mystery instead of 50, I know this logically sounds correct, but I remember it was set to 50 as it's such a huge requirement at 100. This effectively made it like 12 months longer for anybody that didn't have the task, but it didn't remove it from those that already claimed it at 50 and don't have 100.

Issue #3104, opened 2021-11-05
Phosani’s Nightmare Shouldn’t Use Fire Surge if No Harmonised Nightmare Staff


shelixd:
Currently you have to use wrath runes/fire surge to kill Phosani’s Nightmare. In game, you would only use fire surge if you are also using a harmonised nightmare staff. It is just as viable to use powered staffs like a sanguinesti or toxic trident. 

We should be able to use a toxic trident to kill phosani and and use the appropriate supplies per kill (chaos/death/fire/coins/zulrah’s scales) if we dont have wrath runes. Currently the assumption that we are fire surging without a harmonised staff if someone doesn’t own one doesn’t make sense.

PR #5512, opened 2023-11-12
Nightmare QOL

nwjgit:
closes: https://github.com/oldschoolgg/oldschoolbot/issues/3104

### Description:
Various Nightmare QOL changes

### Changes:
-Allow the use of the quantity option.
-Make Tumeken's shadow & Sanguinesti staff usable at Nightmare.
-Use appropriate staff charges in place of Air, Fire, Wrath runes at Phosani's Nightmare.
-Use the proper staff % boost.
-Add a 3% buff if sang staff is being used (similar to Harmonised nightmare staff).
-Add loot & charges used message to the start of each type of nightmare trip.
-Add xp received message to end of every nightmare trip.

### Other checks:
- Removed some redundant checks.
- Removed food check from checkReqs since it is already handled.
- Made checkReqs a basic if chain instead of a loop that checks user 4 times.


src/mahoji/lib/abstracted_commands/nightmareCommand.ts
```diff
@@ -109,20 +110,41 @@ async function checkReqs(
 	}
 }
 
-function perUserCost(isPhosani: boolean, quantity: number) {
+function perUserCost(user: MUser, quantity: number, isPhosani: boolean, hasShadow: boolean, hasSang: boolean) {
 	const cost = new Bank();
+	const tumCharges = shadowChargesPerKc * quantity;
+	const sangCharges = sangChargesPerKc * quantity;
+	if (hasShadow && user.user.tum_shadow_charges < tumCharges) {
```
LuxrayElite, line 117: You are only providing a shadow or sang boost during phosani, however you are checking regardless of type fought

src/mahoji/lib/abstracted_commands/nightmareCommand.ts
```diff
@@ -133,9 +155,6 @@ export async function nightmareCommand(user: MUser, channelID: string, name: str
 		type = 'mass';
 	}
 
-	const err = await checkReqs([user], NightmareMonster, 2, isPhosani);
```
LuxrayElite, line 136: This allows for early code termination/ error messages. qty variable is ultimately reassigned after trip length so checking an single kill would suffice if later error messages about trip length/supplies catch the rest of the conditions

LuxrayElite (commented):
Just a few comments to think about, I didn't get around to testing it due to an error with the toolkit dependancy.

So they may be incorrect assumptions from glancing over the code :)

PR #5515, opened 2023-11-13
Speed Combat Achievments

nwjgit:
### Description:
- Make speed combat achievements use trip time instead of rng.

### Changes:
- Create a new check for speed CA that use trip duration.
- Only 1 speed CA is currently changed as an example.

### Other checks:
- In testing I did about 100+ infernos and still couldn't get lower than 90 minutes. Maybe some re-balancing is needed for certain trips before something like this can be utilized properly.

### DIFF
--- a/src/lib/combat_achievements/combatAchievements.ts
+++ b/src/lib/combat_achievements/combatAchievements.ts
@@ -51,6 +51,12 @@ export type CombatAchievement = {
 			};
 	  }
 	| {
+			time: {
+				maxTimeAllowed: number;
+				hasChance: activity_type_enum | ((data: ActivityTaskData, user: MUser) => boolean);
+			};
+	  }
+	| {
 			notPossible: true;
 	  }
 );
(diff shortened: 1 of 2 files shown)


LuxrayElite:
Would suggest marking as draft to prevent unattainable CA completions.

nwjgit:
> Would suggest marking as draft to prevent unattainable CA completions.

It's obtainable you can check inferno time lb. just suggesting that inferno times should probably be adjusted

nwjgit:
Also in it's current state I highly doubt Magna would even consider merging this. If anything I was hoping more for an inferno rebalancing. If we are saying it only takes roughly 15 kc to get a 65 min time then by 15kc we should already be seeing times between 60-70 minutes. This can also be said for alot of things that could use an adjustment to trip times.

nwjgit:
After some talks in the dev chat i'm going to close this. I don't see a place for this in the bot right now.

Issue #5106, opened 2023-05-06
Tokkulshop Rates


6enny6:
Tokkulshop is currently insanely slow and takes 1 minute to buy 1 uncut sapphire, or 30 minutes to buy 30

You should be able to buy around ~430 uncut sapphires every 30 minutes (source: https://www.youtube.com/watch?t=226&v=F3ZBue6eWw8&feature=youtu.be)

PR #5521, opened 2023-11-15
Update Tokkul Shop

nwjgit:
closes: https://github.com/oldschoolgg/oldschoolbot/issues/5106

### Description:
-Updates Tokkul Shop to reflect in game trip times. Currently the bot only checks gem/ore shop outside of Mor Ul Rek. Having killed jad once you gain access to the upgraded gem/ore shop.

Sources:
Shop any user can access: https://oldschool.runescape.wiki/w/TzHaar-Hur-Lek%27s_Ore_and_Gem_Store#Default
Shop accessed after 1 jad kc: https://oldschool.runescape.wiki/w/TzHaar-Hur-Rin#Default

### Changes:
-Update Ore/Gems prices in Tokkul shop.
-Add a new stock amount for if the user has completed jad.
-Add a new stock amount for if the user has completed jad and is an ironman. (This affects uncut sapphire, emerald, ruby, and diamond)

### Other checks:

### DIFF
--- a/src/mahoji/commands/tokkulshop.ts
+++ b/src/mahoji/commands/tokkulshop.ts
@@ -95,10 +95,12 @@ export const tksCommand: OSBMahojiCommand = {
 		if (user.minionIsBusy) return `${user.minionName} is currently busy and cannot go to the Tzhaar shops.`;
 		const [hasKaramjaDiary] = await userhasDiaryTier(user, KaramjaDiary.easy);
 		const item = TokkulShopItems.find(i => stringMatches(i.name, options.buy?.name ?? options.sell?.name ?? ''));
+		const hasKilledJad: boolean = (await user.getKC(TzTokJad.id)) >= 1 ? true : false;
+		const isIronman = user.user.minion_ironman ? true : false;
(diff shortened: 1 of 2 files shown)


src/mahoji/commands/tokkulshop.ts
```diff
@@ -91,27 +91,62 @@ export const tksCommand: OSBMahojiCommand = {
...
+		if (options.buy) {
+			if (hasKilledJad) {
+				if (isIronman) {
+					shopStock = item.rinIMShopStock
+						? item.rinIMShopStock
+						: item.rinShopStock
+						? item.rinShopStock
+						: item.stock
+						? item.stock
+						: null;
```
gc, line 132: You can simplify this into: `shopStock  = item.rinIMShopStock ?? item.rinShopStock ?? item.stock ?? null;`

gc (changes requested):

src/mahoji/commands/tokkulshop.ts
nwjgit: Updated, I also simplified the two other shopStock using the same format.

PR #5548, opened 2023-12-19
Add Lava scale shard to /mix

nwjgit:
### Description:
- Add Lava scale shards to /mix with the proper multipliers
- Mixing one Lava scale results in 3-6 lava scale shard
- If the user has 99 herblore there is a 98% chance the lava scale gives 6 shards
- If the user has completed the hard wilderness diary the scales are increased by 50% rounded down
- If the user uses wesley to mix their lava dragon scales the result will always be 3 shards per scale regardless of other multipliers

Source: https://oldschool.runescape.wiki/w/Lava_scale

### Changes:
- Add Lava scale shard to crush.ts
- Add wesley variable to interface HerbloreActivityTaskOptions (helps to properly calculate scales to give the user)
- Add the special code for lava dragon scale shards to herbloreActivity.ts

### Other checks:

### DIFF
--- a/src/lib/skilling/skills/herblore/mixables/crush.ts
+++ b/src/lib/skilling/skills/herblore/mixables/crush.ts
@@ -91,6 +91,16 @@ const Crush: Mixable[] = [
 		wesley: true
 	},
 	{
+		item: getOSItem('Lava scale shard'),
+		aliases: ['lava dragon scale', 'lava dragon dust', 'lava dragon', 'lava shard'],
+		level: 1,
+		xp: 0,
+		inputItems: new Bank({ 11_992: 1 }),
+		tickRate: 1,
+		bankTimePerPotion: 0,
+		wesley: true
+	},
+	{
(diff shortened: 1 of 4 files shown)


src/lib/skilling/skills/herblore/mixables/crush.ts
```diff
@@ -90,6 +90,16 @@ const Crush: Mixable[] = [
 		bankTimePerPotion: 0,
 		wesley: true
 	},
+	{
+		item: getOSItem('Lava scale shard'),
+		aliases: ['lava dragon scale', 'lava dragon dust', 'lava dragon', 'lava shard'],
+		level: 1,
+		xp: 0,
+		inputItems: new Bank({ 11_992: 1 }),
```
gc, line 98: Whys this using the item id?

nwjgit: No real reason. I replaced the ID with the item name, I also changed the code in herbloreAcitivity to also use the item name

PR #5557, opened 2023-12-27
Fishing master cape boost

nwjgit:
### Description:
Add a 2x speed boost for fishing master cape, similar to other master capes.
Community poll: https://discord.com/channels/342983479501389826/1032668754561224734/1170807749324587150
firefox_nMnW00pn5u

### Changes:
- Reduce scaledTimePerFish by 50% if the user has Fishing mater cape equipped.
- Reduction for cape is done after the Shelldon check.
- Change the Shelldon boost code to match the format.

### Other checks:


gc (changes requested):
Lets make this not needed to be equipped. A comparison of xp/hr rates before/after would be nice.

DarkBugz:
The average xp/hr with this change would go from
3,177,556 to 6,355,122 (with 1a)

themrrobert:
Let's make this +100% instead, along with shelldon.

So Shelldon + Master Fishing cape =3x instead of 4x.

(See my Cooking boost update PR for example)

themrrobert:
If Magna's OK with it, then I am, too. 

This is in line with many other skills, but we could also do what we do with Fletching/Farming, and give +100% for each Shelldon + Master cape; for a total of +200% (that would make it 3x instead of 4x).

PR #5562, opened 2024-01-02
Broken Pickaxe

Lajnux:
### Description:

Added Ore Pack and Broken Pickaxe to VM

### Changes:

Added Ore Pack and Broken Pickaxe to VM

### Other checks:

### DIFF
--- a/src/lib/data/CollectionsExport.ts
+++ b/src/lib/data/CollectionsExport.ts
@@ -1488,7 +1488,12 @@ export const troubleBrewingCL = resolveItems([
 	'Red rum (trouble brewing)',
 	'Blue rum (trouble brewing)'
 ]);
-export const volcanicMineCL = resolveItems(['Ash covered tome', 'Large water container', 'Volcanic mine teleport']);
+export const volcanicMineCL = resolveItems([
+	'Ash covered tome',
+	'Large water container',
+	'Volcanic mine teleport',
+	'Dragon pickaxe (broken)'
+]);
 export const anglerOutfit = resolveItems(['Angler hat', 'Angler top', 'Angler waders', 'Angler boots']);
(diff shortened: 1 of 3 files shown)


nwjgit:
I would recommend this be handled in a different way. The Ore Pack from volcanic mine is an actual item in-game: https://oldschool.runescape.wiki/w/Ore_pack_(Volcanic_Mine) . This code auto gives the loot from the ore pack when purchased. 

This would remove alot of functionality and user experience such as:
- `/lb opens openable:`
- `/data name:Personal Open Stats`
- Bank log
- Sac log
- No functionality to the ore pack in BSO when received from a mystery box.

Having an openable function differently than 99% of other openables probably isn't the way to go. 
There is already a way to easily handle openables with oldschooljs. Trying to re-invent the wheel here would just add to player confusion and frustration in my opinion.

Lajnux:
I don't agree with this,
It adds extra unnecessary steps for the user.
We already have Ore Pack from Giant's Foundry and would have to separate them.
The Ore Pack doesn't have a CL itself, Giant's Foundry one does.
There is plenty of items that are unopenable in BSO, making extra work for a mystery box item to be openable isn't really a priority.

nwjgit:
> I don't agree with this, It adds extra unnecessary steps for the user. We already have Ore Pack from Giant's Foundry and would have to separate them. The Ore Pack doesn't have a CL itself, Giant's Foundry one does. There is plenty of items that are unopenable in BSO, making extra work for a mystery box item to be openable isn't really a priority.

Why make this one openable function differently than 99% of the other ones though?

TastyPumPum:
Dragon Pickaxe broken has been added in: https://github.com/oldschoolgg/oldschoolbot/pull/5709

PR #5571, opened 2024-01-12
Degradable Items fixes

nwjgit:
### Description:
Fixes and corrections to degradableitems. Mainly allowing for variant items such as "Holy scythe of vitur" to be charged and converted even if the user doesn't own a normal scythe of vitur.
This PR is a follow up for: https://github.com/oldschoolgg/oldschoolbot/pull/5422
### Changes:
- Changed `/minion charge` to properly check for variant items instead of only the original item.
- Add a new variable for variants to handle returning the proper items. Example: Abyssal tentacle (or) breaks into two items so this variable is best handled as Bank()
- Fixed Abyssal tentacle not checking for kraken tentacle before charging
- Abyssal tentacle now properly converts on charge
- Updated the message if the user doesn't have the item they want to charge
- Change some variable naming
- Make more variables use Item instead of relying on Bank
- Corrected some code to properly remove variant items from the users bank (not currently used by any items but good for future use if for example ash sanctifier gets a cosmetic kit)
- Corrected tracking of "degraded_items_cost"
### Other checks:
- I did extensive testing here: https://discord.com/channels/940758552425955348/1162793245194072074


src/lib/degradeableItems.ts
```diff
@@ -302,51 +302,68 @@ export async function degradeItem({
...
 			}
-			// Unequip and delete the users item.
+			// Unequip and delete the item from the users gear setup
 			const gear = { ...user.gear[degItem.setup].raw() };
 			gear[item.equipment!.slot] = null;
 			await user.update({
 				[`gear_${degItem.setup}`]: gear
 			});
-			// Give the user the uncharged version of their charged item.
+			console.log(removeItems);
```
LuxrayElite, line 332: left some logging in unsure if that was intentional?

src/lib/minions/functions/degradeableItemsCommand.ts
```diff
@@ -17,54 +17,80 @@ export async function degradeableItemsCommand(
...
+			unchargedItem = variant.unchargedVariant;
+			chargedItem = variant.chargedVariant;
+			break;
+		}
+	}
+
+	// Show error message if the user doesn't have the charged, uncharged, or variants of the item in the bank
+	if (!user.hasEquippedOrInBank(chargedItem.id) && !user.owns(unchargedItem.id)) {
+		return `You don't own a ${chargedItem.name}${
+			unchargedItem!.name === chargedItem.name ? '' : `, ${unchargedItem!.name}`
```
LuxrayElite, line 57: I think this conditional could have been handled better, if the uncharged and charged are the same name that property can be removed due to it not being required.

https://github.com/oldschoolgg/oldschoolbot/blob/5f7e55b97acace2635919ae3bb3b076df44f8919/src/lib/degradeableItems.ts#L12-L39C2

This would mean modifying how the break code works so that it gives back the item if no uncharged version exists

src/lib/degradeableItems.ts
```diff
@@ -302,51 +302,68 @@ export async function degradeItem({
 	const newCharges = Math.floor(currentCharges - chargesToDegrade);
 
 	if (newCharges <= 0) {
-		// If no more charges left, break and refund the item.
+		// If the user runs out for charges for a degradeable item, break and refund the item.
 		const hasEquipped = user.gear[degItem.setup].hasEquipped(item.id, false);
-		const hasInBank = user.owns(item.id);
+		const hasInBank = user.hasEquippedOrInBank(item.id);
```
LuxrayElite, line 307: owns is preferred due to exclusively checking the bank (unless equipment variable is passed)

In this scenario owning a single charged scythe (equipped) would make it true for "hasEquipped" and "hasInBank". IF statements later filter this variable out due to equipped taking priority.

LuxrayElite (commented):
During my testing I found an error as well as a few things that I'm curious about.

## Error
Error (Application didnt respond) - Easily repeatable, will affect large amount of new users (presumably) 
Unable to charge blood essence (active) unless already owning the item (once made it is permanently restricted to 1 charge minimum)
image

## Non-issues / curiosities
Unsure if this was in the bot prior to this pr however when charging a scythe with no charged scythes it will prioritize holy scythes over the default scythe(if you own multiple, I saw this in your testing) its a non-issue due to /create but highlighting it. Breaking still works as intended.

Unsure why aliases have been changed even though they're unused so a non-issue really

Left some code related feedback on a few places

src/lib/degradeableItems.ts
nwjgit: This is .hasEquippedOrInBank since its the only bank function I could find that checks for similar items. There isn't a scenario where this wouldn't work that I can think of. The user has to have the equipable chargeditems equipped for it to lose charges.

src/lib/minions/functions/degradeableItemsCommand.ts
nwjgit: Yeah I was trying to make an all in one error for if the user didn't have an item, like ash sanctifier. I just made a new error for that case since this seems to break too many things and lacks readability.

src/lib/degradeableItems.ts
nwjgit: fixed

src/lib/minions/functions/degradeableItemsCommand.ts
nwjgit: This is also what caused the blood essence issue, and has been fixed.

nwjgit:
Updated a few things:
- Add a new variable for variants to handle returning the proper items. Example: Abyssal tentacle (or) breaks into two items so this variable is best handled as Bank()
- Fixed Abyssal tentacle not checking for kraken tentacle before charging
- Abyssal tentacle now properly converts on charge

PR #5578, opened 2024-01-15
Allow Tame to collect more items & dt2 tablet fix

nwjgit:
### Description:
- Allows Igne tame to collect "Special items" that are handled by unique code, inaccessible to tames such as: 
  - specialLoot under Vladimir Drakan(outfit, Echo, Plushie)
  - effect under dt2 bosses(ring of shadow tablets)
  - effect under Phantom Muspah(Charged ice)
- Add Phosani Nightmare to tame killable (Slepey tablet & Parasitic egg)
- Correct an issue preventing users from being able to collect the dt2 tablet cl slot
### Changes:
- SpecialLoot interface to tames.ts to handle the monster, items, and chance
- Code to handle giving items from SpecialLoot (Currently only allows for 1 of each item). This code will work for any future special/unique code that isn't handled with a loottable
- Phosani Nightmare to tame killable
- Special case for Slepey tablet, similar to minion to only allow for 1
- Added a second check to dt2 tabs to prevent making the cl inaccessible if the user has one in bank (obtained through trading or tame killing)
- Removed Avernic defender hilt boost from awakened duke/vard due to scythe being a 2h weapon and required to kill.
### Other checks:

### DIFF
--- a/src/lib/tames.ts
+++ b/src/lib/tames.ts
@@ -22,6 +22,7 @@ import { PHOSANI_NIGHTMARE_ID } from './constants';
 import { getSimilarItems } from './data/similarItems';
 import { trackLoot } from './lootTrack';
 import killableMonsters, { NightmareMonster } from './minions/data/killableMonsters';
+import { BSOMonsters } from './minions/data/killableMonsters/custom/customMonsters';
 import { customDemiBosses } from './minions/data/killableMonsters/custom/demiBosses';
 import { Planks } from './minions/data/planks';
 import { KillableMonster } from './minions/types';
@@ -169,8 +170,7 @@ export const specialLoot: SpecialLoot[] = [
 		]
 	},
 	{
-		// Vladimir Drakan
-		monster: [291_242],
+		monster: [BSOMonsters.VladimirDrakan.id],
 		specialItems: [
 			{
 				item: [
--- a/src/lib/tames.ts
+++ b/src/lib/tames.ts
@@ -18,6 +18,7 @@ import { ChambersOfXeric, TheatreOfBlood } from 'oldschooljs/dist/simulation/mis
 
 import { collectables } from '../mahoji/lib/abstracted_commands/collectCommand';
 import { mahojiUsersSettingsFetch } from '../mahoji/mahojiSettings';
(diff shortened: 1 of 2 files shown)


gc:
Preferably, in the future split things up into several PRs rather than grouping unrelated things together in 1 PR. It makes it much harder to review.

nwjgit:
> Preferably, in the future split things up into several PRs rather than grouping unrelated things together in 1 PR. It makes it much harder to review.

No problem. How would you like this one split up? I figured since its all tame related I would keep it to 1 PR.

gc:
The DT2 stuff seemed separate, maybe im wrong.

src/lib/tames.ts
```diff
@@ -113,6 +115,86 @@ export const seaMonkeySpells: SeaMonkeySpell[] = [
...
+				item: [itemID('Vampyric plushie')],
+				chance: [250]
+			},
+			{
+				item: [itemID('Echo')],
+				chance: [1000]
+			}
+		]
+	}
+];
```
gc, line 196: I'm not sure im a fan of any of how this works. The thing I hate more than anything in code is having several sources of truth like this (e.g. if we change echo droprate, we now need to change it in 2 places and nobody will remember to do that in both).

For vladimir drakans case, you can use the original code just calling specialLoot. Also maybe do a quick check over all the other specialLoots to make sure nothing crazy will happen letting tames get that.

gc (changes requested):

nwjgit:
> The DT2 stuff seemed separate, maybe im wrong.

The dt2 fix is what sort of stemmed the entire PR. The tablets are currently unobtainable because they got removed from oldschooljs and added as an effect under killable monster. So it gives Top Tame cl role a bit of an advantage. I'll split it into different PRs to address the different issues tho.

src/lib/tames.ts
nwjgit: At least for echo I could use the globalDroprates.ts rate. I'll try some more stuff out and see what I can come up with.

gc: You dont need too, just call monster.specialLoot(). Check monsterActivity

PR #5579, opened 2024-01-15
Block blacklisted people receiving gp from /pay

TastyPumPum:
Adds a blacklist check to the /pay command.
Informs the sender that they can't send gp to someone blacklisted, similar to how it works for /trade


themrrobert:
Could you change the error text to say, 'cannot receive money' to be consistent with the other failure modes?

Issue #5523, opened 2023-11-17
Mysterious tree contracts only give 1 seed pack


gpRollMaster:
Is there a reason mysterious tree contracts only give 1 seed pack instead of the standard 3 (4 if arcane harvester is enabled)?
If arcane harvester is enabled, still only gives 1

![Screenshot_20231115-182127](https://github.com/oldschoolgg/oldschoolbot/assets/97349844/ee17ac23-9b7f-4158-873f-dcaf89c6b3ea)
![Screenshot_20231115-182711](https://github.com/oldschoolgg/oldschoolbot/assets/97349844/6551bf0d-e607-430b-9440-87bb75b7001e)
![Screenshot_20231115-185048](https://github.com/oldschoolgg/oldschoolbot/assets/97349844/d8bc742e-e7de-4626-a088-4cf081b8ad0c)

nwjgit:
The jane message also isn't showing the correct amount of seed packs

nwjgit:
Fixed.

PR #5604, opened 2024-01-25
Fix mysterious tree contracts

nwjgit:
### Description:
Fixes an issue causing mysterious tree contracts to not properly multiple the seed packs & non mystery box loot. 
Fixes the wording in the Jane message

### Changes:
- remove plantsNotUsedForArcaneHarvester
- Make Mysterious tree contracts give an extra seed pack by default to make them give four total with max boosts to match other contracts.
- Add a return to the final handleTripFinish to remove the eslint warning
- Fix the Jane chatHeadImage() for farming contracts to show the correct amount of seed packs received. 
- Remove outdated plopper message "You recieved 100% bonus loot from Plopper" (seen below) since the "100% for Plopper" message is already shown

### Other checks:
closes: https://github.com/oldschoolgg/oldschoolbot/issues/5523

Discord_viM1SQmy4A

### DIFF
--- a/src/tasks/minions/farmingActivity.ts
+++ b/src/tasks/minions/farmingActivity.ts
@@ -41,7 +41,7 @@ async function farmingLootBoosts(user: MUser, plant: Plant, loot: Bank, messages
 		bonusPercentage += 100;
 		messages.push('100% for Farming master cape');
 	}
-	if (user.hasEquippedOrInBank(['Arcane harvester']) && !(plant.name === 'Mysterious tree')) {
+	if (user.hasEquippedOrInBank(['Arcane harvester']) && plant.name !== 'Mysterious tree') {
 		const boostRes = await inventionItemBoost({
 			user,
 			inventionID: InventionID.ArcaneHarvester,
--- a/src/tasks/minions/farmingActivity.ts
+++ b/src/tasks/minions/farmingActivity.ts
@@ -44,18 +44,19 @@ async function farmingLootBoosts(user: MUser, plant: Plant, loot: Bank, messages
 		bonusPercentage += 100;
 		messages.push('100% for Farming master cape');
 	}
-	if (plantsNotUsedForArcaneHarvester.includes(plant)) return;
-	if (user.hasEquippedOrInBank(['Arcane harvester'])) {
-		const boostRes = await inventionItemBoost({
-			user,
-			inventionID: InventionID.ArcaneHarvester,
-			duration: plant.level * Time.Second * 30
-		});
-		if (boostRes.success) {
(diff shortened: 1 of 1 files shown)


themrrobert:
Just take out the plantsNotUsedForArcaneHarvester code entirely.

The mystery boxes from Mysterious trees  are handled after the loot multiplication anyway, with the current code, mysty trees will still only give 3 instead of 4 seed packs.

So just remove the:
```ts
const plantsNotUsedForArcaneHarvester = ['Mysterious tree'].map(i => Farming.Plants.find(p => p.name === i)!);
assert(!(plantsNotUsedForArcaneHarvester as any[]).includes(undefined));
```
from the top of the file, and then the surrounding logic referencing that const in your modified code.

The farming code still needs more refactoring before we should be using code like this, since it's not doing it's intended job, and only having negative side effects.

nwjgit:
> Just take out the plantsNotUsedForArcaneHarvester code entirely.
> 
> The mystery boxes from Mysterious trees are handled after the loot multiplication anyway, with the current code, mysty trees will still only give 3 instead of 4 seed packs.
> 
> So just remove the:
> 
> ```ts
> const plantsNotUsedForArcaneHarvester = ['Mysterious tree'].map(i => Farming.Plants.find(p => p.name === i)!);
> assert(!(plantsNotUsedForArcaneHarvester as any[]).includes(undefined));
> ```
> 
> from the top of the file, and then the surrounding logic referencing that const in your modified code.
> 
> The farming code still needs more refactoring before we should be using code like this, since it's not doing it's intended job, and only having negative side effects.

I thought the same thing but didn't want to remove pre-existing code, I'll make the change now.

nwjgit:
> Just take out the plantsNotUsedForArcaneHarvester code entirely.
> 
> The mystery boxes from Mysterious trees are handled after the loot multiplication anyway, with the current code, mysty trees will still only give 3 instead of 4 seed packs.
> 


Also we may want to rethink the solution here because since mystery tree loot is pretty worthless, and people only do it for the boxes. Removing this code will make arcane harvester work with mystery trees and remove mats unnecessarily. This could be circumvented by toggling it off but that would drastically hurt qol doing that each farm run.

nwjgit:
I changed it around so that Mysterious trees don't use Arcane harvester since this is how it's worked for awhile and the base loot is really bad and it just wastes invention materials for no real gain. Added new code to make Mysterious trees give an extra seed pack for contracts. This makes it so that even without Arcane harvester mysterious tree contracts still get 4 seed packs at max boost. This makes the default 2 seed packs per mysterious tree contract.

PR #5660, opened 2024-02-06
XP from Tears of Guthix

nwjgit:
### Description:
Data command for XP received from tears of Guthix
### Changes:
- Add `/data name: XP from Tears of Guthix`
### Other checks:
Discord_UbXx1nth97

### DIFF
--- a/src/mahoji/lib/abstracted_commands/statCommand.ts
+++ b/src/mahoji/lib/abstracted_commands/statCommand.ts
@@ -956,12 +956,13 @@ GROUP BY "bankBackground";`);
 	{
 		name: 'XP from Tears of Guthix',
 		perkTierNeeded: PerkTier.Four,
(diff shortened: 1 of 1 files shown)


gc:
This is global XP from tears of guthix. If thats intentional, it should say global. If its meant to be per-user, it should be changed to do that.

nwjgit:
> This is global XP from tears of guthix. If thats intentional, it should say global. If its meant to be per-user, it should be changed to do that.

Thanks for the catch, I added the check for user id. It's intended to be per-user and not global. Although I could add another one in for global if you'd like.

gc (changes requested):
The title needs to be changed to say "personal" too

nwjgit:
> The title needs to be changed to say "personal" too

done.

PR #5663, opened 2024-02-07
Adding tgb for alias & removed lowercase only cl searching

TURB042O:
### Description:

- made it so when you search on the cl, and you start with a capital letter it will register it. ex: `Halloween` instead of `halloween` both will now work.
- added tgb for alias to tester gift box

### Other checks:


themrrobert:
I'm not sure what you did here, but you merged in changes from master, including my most recent PR, so it seems like the way you updated your PR wasn't right.

PR #5672, opened 2024-02-08
Remove extra TOB clue scrolls

nwjgit:
### Description:
Stops TOB loot image from showing the user they got more than 1 clue scroll (OSB only)
### Changes:
- Add two new checks to make sure the user is only ever shown 1 clue scroll from TOB (OSB only) and isn't shown any if one is already in their bank.
### Other checks:
Prevents this from happening:
Discord_KlKE62BqLx

### DIFF
--- a/src/lib/simulation/tob.ts
+++ b/src/lib/simulation/tob.ts
@@ -83,10 +83,6 @@ const NonUniqueTable = new LootTable()
 	.add('Yew seed', 3)
 	.add('Magic seed', 3);
 
-if (BOT_TYPE === 'BSO') {
-	NonUniqueTable.tertiary(25, 'Clue scroll (elite)');
-}
-
 const HardModeExtraTable = new LootTable()
 	.tertiary(275, 'Sanguine dust')
 	.tertiary(150, 'Sanguine ornament kit')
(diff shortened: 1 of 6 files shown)


src/lib/simulation/tob.ts
```diff
@@ -96,6 +96,11 @@ export class TheatreOfBloodClass {
 		const loot = new Bank();
 		for (let i = 0; i < 3; i++) {
 			loot.add(NonUniqueTable.roll());
+
+			// OSB only: Remove dupe elite clue scrolls
+			if (loot.amount('Clue scroll (elite)') > 1) {
+				loot.remove('Clue scroll (elite)', 1);
+			}
```
gc, line 103: - Seems like it wont remove all extra clues, unless its guaranteed you can only get {0,1,2}. Perhaps remove `numClues - 1`
- If you want code to definitely only run in OSB, you can check `if (BOT_TYPE === 'osb')`. The upside is no need to worry about removing it in BSO when merging. You don't have to do this though.

src/tasks/minions/minigames/tobActivity.ts
```diff
@@ -141,6 +141,11 @@ export const tobTask: MinionTask = {
 				// Refund initial 100k entry cost
 				userLoot.add('Coins', 100_000);
 
+				// OSB only: Remove elite clue scroll if user has one in bank
+				if (user.owns('Clue scroll (elite)')) {
+					userLoot.remove('Clue scroll (elite)', 1);
+				}
```
gc, line 147: Fine as long as its not possible to get 3+

src/lib/simulation/tob.ts
nwjgit: Not sure what you're referring to with the `numClues -1`. But I think I found a solution that should work for all scenarios, I'll add it soon.

src/tasks/minions/minigames/tobActivity.ts
nwjgit: Correct, this is after the loot sim from tob.ts so at this point the users loot should only have max 1 elite clue scroll. Then if they have one in the bank it will delete it.

PR #5678, opened 2024-02-10
Fixed only lowercase searches in CL issue

TURB042O:
### Description:

right now searching any CL if you start your search using a capital letter for example `Halloween`.. no results show.   but when you type `halloween` all the halloween results show. this should fix that.

### Changes:

- changed `(value)` to `(value.toLowerCase)`

### Other checks:

### DIFF
--- a/src/mahoji/commands/cl.ts
+++ b/src/mahoji/commands/cl.ts
@@ -39,7 +39,7 @@ export const collectionLogCommand: OSBMahojiCommand = {
 							];
 						})
 						.flat(3)
-				].filter(i => (!value ? true : i.name.toLowerCase().includes(value)));
+				].filter(i => (!value ? true : i.name.toLowerCase().includes(value.toLowerCase())));
 			}
 		},
 		{
--- a/src/mahoji/commands/cl.ts
+++ b/src/mahoji/commands/cl.ts
@@ -30,16 +30,16 @@ export const collectionLogCommand: OSBMahojiCommand = {
 					{ name: 'Overall+', value: 'overall+' },
 					...Object.entries(allCollectionLogs)
 						.map(i => {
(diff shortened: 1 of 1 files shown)


nwjgit:
What were you trying to accomplish here exactly? `/cl name:` already isn't case sensitive and you can do `/cl name: nex` or `/cl name: Nex` or even `/cl name: NeX` and all three return the Nex collection log.

DayV-git:
This change is for the autocomplete function, if you type `/cl name: nex`, nothing will come up at the moment in the search

nwjgit:
> This change is for the autocomplete function, if you type `/cl name: nex`, nothing will come up at the moment in the search

Ah I see that now, thanks for clarifying.

TURB042O:
> What were you trying to accomplish here exactly? `/cl name:` already isn't case sensitive and you can do `/cl name: nex` or `/cl name: Nex` or even `/cl name: NeX` and all three return the Nex collection log.
> 
> If there is a certain cl that isn't working properly then it is more likely an issue with it's aliases. 

sorry I am just now seeing this comment. idk why but github mobile has never notified for comments :( so thank you @DayV-git for responding :) 

but not sure if this has gone live yet or not haven't had chance to check but here are screen shots of what I'm hoping to achieve.



this is with capital H (just like the cl is labeled) nothing shows.



now I retype it with lowercase h and here's all halloweens.. thats just the example I choose to use.

TURB042O:
so I'm guessing this didn't fix anything because the issue I'm trying to fix still exists.

maybe somebody will know the answer? @nwjgit @gc 


the issue is that I've noticed only in /cl the searching is case sensitive 


because anywhere else capitals don't matter 


what's causing this?

PR #5684, opened 2024-02-11
Fix dt items for tames

themrrobert:
### Description:

Fixes DT one-of drops for tames

### Changes:

updates tames to call effects and pass it's CL

### Other checks:


nwjgit:
Muspah also uses the effect for charged ice if you wanted to add that in aswell.

nwjgit:
sorry closed the wrong one

src/lib/minions/types.ts
```diff
@@ -113,8 +113,9 @@ export interface KillableMonster {
 	effect?: (opts: {
 		messages: string[];
 		user: MUser;
+		tame?: Tame;
```
gc, line 116: Change to use `MTame`, so you can do `tame.totalLoot` instead of  `new Bank(tame.max_total_loot as ItemBank)`. Merge at your discretion when its ready.

PR #5686, opened 2024-02-12
Add functionality to Ghommal's lucky penny

nwjgit:
### Description:
"When equipped, it offers a 5% chance to not use a charge when using items that consume charges (such as jewellery and weapons)." https://oldschool.runescape.wiki/w/Ghommal%27s_lucky_penny
### Changes:
- Add a 5% chance to not use a charge when the user has Ghommal's lucky penny equipped.
- Reduce amount of blood runes to charge scythe from 300 to 200 (This was implemented in-game osrs 17Jan24)
### Other checks:

### DIFF
--- a/src/lib/degradeableItems.ts
+++ b/src/lib/degradeableItems.ts
@@ -212,7 +212,7 @@ export const degradeableItems: DegradeableItem[] = [
 		setup: 'melee',
 		aliases: ['scythe of vitur'],
 		chargeInput: {
-			cost: new Bank().add('Blood rune', 300).add('Vial of blood').freeze(),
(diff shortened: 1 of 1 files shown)


src/lib/degradeableItems.ts
```diff
@@ -297,6 +298,17 @@ export async function degradeItem({
 	const degItem = degradeableItems.find(i => i.item === item);
 	if (!degItem) throw new Error('Invalid degradeable item');
 
+	// 5% chance to not consume a charge when Ghommal's lucky penny is equipped
```
gc, line 301: If its a 5% percent chance, we should use `percentChance(5)`

nwjgit: Switched it to percentChance(5)

PR #5692, opened 2024-02-16
Remove Hell Cat Ears from Createables CL

TastyPumPum:
As cat ears are a discontinued item and can't be traded, remove them from the creatables CL like other holiday items.


themrrobert:
Hell cat ears can be created at any time if you have the original components. Only the Cat ears themselves are discontinued.

The Red  dye used to be discontinued, but it was (unintentionally?) re-released into the /buy command

PR #5702, opened 2024-02-19
important updates to CPETS & DCPETS 

TURB042O:
### Description:

made all pets except leagues pet bank value 1m, changed some pets from `Coal` to `Herbi` to register as a pet.

### Changes:

- Changed `Coal` to `Herbi` on Rudolph & bv 1m
- ~~made doopy tradeable GE only~~
- doopy is tradeable via player & ge.
- all pets in customitems that are labeled `Herbi` are now 1m bv
- Removed tradeable: true & tradeable_on_ge: true from Balloon cat due to it is ment to be untradeable and has the super untradeable code also.
- moved Balloon cat out of discontinued section in pets.ts above dyed
- ~~removed `cantBeSacrificed: true` from the guthix, saradomin & zamorak pets.~~
- ~~removed `cantBeSacrificed: true` from Buggy~~
- turned Skip from `Coal` to `Herbi` because it is a growable pet.

- Added `isDiscontinued: true` to casper, doopy & mini mortimer

### Other checks:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9798 @@
+{
+	"name": "osb",
+	"lockfileVersion": 3,
+	"requires": true,
+	"packages": {
+		"": {
+			"dependencies": {
+				"@discordjs/builders": "^1.5.0",
+				"@fastify/cors": "^8.2.0",
+				"@fastify/helmet": "^10.1.0",
+				"@fastify/rate-limit": "^8.0.0",
+				"@fastify/sensible": "^5.2.0",
+				"@napi-rs/canvas": "0.1.38",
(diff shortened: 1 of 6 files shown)


TURB042O:
@gc I was having dumb issues with the last PR. and branch. but this one seems to have passed everything perfectly!

gc:
For future PRs like this, you should make them smaller. There is atleast 1 issue I can see, doopy is not discontinued.

TURB042O:
> For future PRs like this, you should make them smaller. 
alright. thank you for the insight. 


> There is atleast 1 issue I can see, doopy is not discontinued. 
oooh you're right. good catch I'll edit that.

Issue #3825, opened 2022-04-22
Add a method of uncharging glories for amulet of eternal glory


Redquaker:
## 🚀 Feature Proposal

Currently, the only way to uncharge glories is to do gnome restaurant for a 1/3 chance to use some glory charges. Needless to say, this is extremely slow. To hunt for the amulet of eternal glory, it's basically required to make fresh glories.

There should be a command to uncharge Amulet of glory (6) to Amulet of glory (5) so that they can be charged again for more chances at the amulet of eternal glory. For example, consider one teleport to take 3 ticks (1.8 seconds) -> round up to 2 seconds per amulet uncharged, or even make it a bit slower than that.

This would also likely need the +chargeglories command changed to accept Amulet of glory (5) for charging.

RLChristof:
Would love to see this feature as I come closer to having to do that grind for CL

nwjgit:
> Would love to see this feature as I come closer to having to do that grind for CL

This is something i plan to work on soon unless someone beats me to it.

nwjgit:
This wasn't completed since #5732 was closed and not merged. If @TastyPumPum doesn't want to make a new PR then i'll tackle it eventually.

Redquaker:
Completed by #6772

PR #5732, opened 2024-02-27
Uncharge glory

TastyPumPum:
Adds the ability to uncharge amulet of glory (6)'s. This will be useful for players chasing the eternal glory at the fountain of rune.

Closes [#3825](https://github.com/oldschoolgg/oldschoolbot/issues/3825)

### DIFF
--- a/prisma/robochimp.prisma
+++ b/prisma/robochimp.prisma
@@ -57,6 +57,9 @@ model User {
 
     store_bitfield Int[]
 
+    testing_points         Float @default(0)
+    testing_points_balance Float @default(0)
+
     @@map("user")
 }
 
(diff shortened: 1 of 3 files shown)


src/mahoji/lib/abstracted_commands/chargeGloriesCommand.ts
```diff
@@ -65,3 +65,47 @@ export async function chargeGloriesCommand(user: MUser, channelID: string, quant
...
+
+	await addSubTaskToActivityTask<ActivityTaskOptionsWithQuantity>({
+		userID: user.id,
+		channelID: channelID.toString(),
+		quantity,
+		duration,
+		type: 'GloryUncharging'
+	});
+
+	await user.removeItemsFromBank(new Bank().add('Amulet of glory (6)', quantity));
```
gc, line 106: Remove before creating the activity. Also the cleaner way to do it is: `const cost = ...` `if (!user.owns(cost))`

PR #5733, opened 2024-02-27
Add Crystal saw

### LINKED ISSUE #5311: [OSB] Add crystal saw as buyable, and use in /poh build
## 🚀 Feature Proposal

We should add the crystal saw as a buyable, given the minion has the skill requirements for the relevant quest, https://oldschool.runescape.wiki/w/The_Eyes_of_Glouphrie

This would be namely:
* Completed The Grand Tree quest
  * Requires 25 agility (boostable)
* Construction 5 Construction (not boostable)
* Magic 46 Magic (not boostable)

When in the bank, the saw would give it's usual +3 boost to the relevant POH builds, (e.g not garden, but to jewelry boxes, etc).

TastyPumPum:
Adds crystal saw to the buyables provided they have the relevant stats.
Give the relevant +3 construction boost for using for the saw.

Closes #5311 


src/lib/data/buyables/buyables.ts
```diff
@@ -717,6 +717,16 @@ const questBuyables: Buyable[] = [
...
+	{
+		name: 'Crystal saw',
+		gpCost: 75_000,
+		qpRequired: 7,
+		skillsNeeded: {
+			agility: 25,
+			construction: 5,
+			magic: 46
+		}
 	}
```
gc, line 730: Isnt it made using a seed? and uses charges?

TastyPumPum: Yes it's made using a seed, I could make the seed the buyable and make a saw a createable but that seems very long winded.

Also charges are used yes but it's free to recharge, so again, would just seem a long route about it.

TastyPumPum: As a thought, maybe it could be taken out the build command as its not viable, and just kept in the pohcommand and stash unit command?

PR #5735, opened 2024-02-28
Make monster activity work like everything else with 2x loot.

themrrobert:
### Description:

- Currently, `monsterActivity` gives 2x kill quantity, instead of doubling the loot like everything else.
- This fixes that.

### Changes:

- Removes old 2x code from monster activity and replaces it with `.multiply(2)`

### Other checks:

### DIFF
--- a/src/tasks/minions/monsterActivity.ts
+++ b/src/tasks/minions/monsterActivity.ts
@@ -379,9 +379,10 @@ export const monsterTask: MinionTask = {
 				tertiaryItemPercentageChanges: user.buildCATertiaryItemChanges()
 			}
 		};
-		const finalQuantity = isDoubleLootActive(duration) ? quantity * 2 : boostedQuantity;
-		// Regular loot
-		const loot = (monster as KillableMonster).table.kill(finalQuantity, killOptions);
+
+		const loot = (monster as KillableMonster).table.kill(boostedQuantity, killOptions);
+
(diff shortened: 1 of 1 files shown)


DarkBugz:
This should be polled?

DarkBugz:
You say "like everything else" but there's only four cases. Bso bosses get 2x loot, raids get 2x loot, all osb monsters get 2x rolls. Bso demi-bosses get 2x rolls. 

Would this causes ori +25% rolls to stack with double loot which it currently doesn't?

themrrobert:
> You say "like everything else" but there's only four cases. Bso bosses get 2x loot, raids get 2x loot, all osb monsters get 2x rolls. Bso demi-bosses get 2x rolls.
> 
> Would this causes ori +25% rolls to stack with double loot which it currently doesn't?

Double loot applies to:

Monkey rumble - this also has speed + increased odds boosts,, should those not count? It gets the loot actually doubled.
KG / KK / Nex / Moktang / Vasa Magus / Ignecarus (all separate activities) (loot actually doubled)
DG Agility - Gorajan shards (loot actually doubled)
Dungeoneering - Gorajan shards (etc..)

So it's really 9 other tasks that all behave the same way, the loot is doubled, including stuff from boosts.

Every time we have a double loot, someone reports 'its broken I didn't get 2x of whatever drop,'

PR #5737, opened 2024-02-29
Forestry update

nwjgit:
### Description:
Forestry part two & forestry updates.

### Changes:
- new `/chop` flags
  - Twitcher gloves flag (gloves only work equipped)
  - Forestry flag (stops you from doing events, also removes wc guild boost when doing events)
- Adds items & features from Forestry: Part Two.
- Unique forestry events with loot, xp and end trip message.
- Boost for using Felling axe with Forester's ration.
- Add Forestry basket.
- Adds leafs to trees.
- Boost message for log basket. 
- Twitcher's gloves with proper functionality.
- Fox & pheasant pet.
- Removed: Clover insignia, Powdered pollen, Leprechaun charm, Bee on a stick
- Add 10% more nests from having woodcutting cape or a completed forestry basket (need cape pouch & 99 cape in bank)
- Tracking forestry events and xp for future use in `/data`

TODO: **Plan to add all these in their own PR since this PR is already fairly big**
- Ability to create Forester's rations: https://oldschool.runescape.wiki/w/Forester%27s_ration
- Add utility for Secateurs attachment, Nature offerings, sawmill voucher
- Add data command to show xp and number of events completed

### Other checks:


src/mahoji/commands/chop.ts
```diff
@@ -98,13 +99,37 @@ export const chopCommand: OSBMahojiCommand = {
...
+		{
+			type: ApplicationCommandOptionType.String,
+			name: 'twitchers_gloves',
+			description: "Change the settings of your Twitcher's gloves. (default egg nests, optional)",
+			required: false,
+			choices: [
+				{ name: 'Egg nest (Default)', value: 'egg' },
+				{ name: 'Ring nest', value: 'ring' },
+				{ name: 'Seed nest', value: 'seed' },
+				{ name: 'Clue nest', value: 'clue' }
```
gc, line 118: These should be strictly typed, something like:

````
const twitchers = [...] as const;
type TwitcherType = typeof twitchers[number].id;
```

src/lib/types/minions.ts
```diff
@@ -193,6 +193,8 @@ export interface WoodcuttingActivityTaskOptions extends ActivityTaskOptions {
 	fakeDurationMax: number;
 	fakeDurationMin: number;
 	powerchopping: boolean;
+	forestry: boolean;
+	twitchers: string;
```
gc, line 197: This should be strictly typed.

Sidenote: Ideally, they would be optional, and the key doesnt exist at all if its false or not in use, as it will save a bit of storage in the db.

For example, instead of storing `{ forestry: false }` in the db, we can just not have that key in the object at all, so `forestry?: true`. For twitchers, it can be the same thing. And the default would go in the task, instead of the default being stored in the data here.

src/tasks/minions/woodcuttingActivity.ts
```diff
@@ -1,35 +1,262 @@
...
+		uniqueXP: SkillsEnum.Woodcutting
+	},
+	{
+		id: 9,
+		name: 'Pheasant Control',
+		uniqueXP: SkillsEnum.Thieving
+	}
+];
+
+const leafTable = new LootTable()
```
gc, line 69: We generally write loottable names like this in TitleCase.

src/tasks/minions/woodcuttingActivity.ts
```diff
@@ -1,35 +1,262 @@
...
+	.add('Leaves', 20)
+	.add('Oak Leaves', 20)
+	.add('Willow Leaves', 20)
+	.add('Maple Leaves', 20)
+	.add('Yew Leaves', 20)
+	.add('Magic Leaves', 20);
```
gc, line 75: Item names in our code are always written following the osrs convention (only first word capitalized)

gc (changes requested):

gc:
Apart from the few comments, it looks good as long as its tested and feedback has been addressed

src/mahoji/commands/chop.ts
nwjgit: fixed, i added it constants.ts since i saw that's where a good majority of them were, let me know if they should be somewhere else.

src/mahoji/commands/chop.ts
```diff
@@ -229,7 +225,7 @@ export const chopCommand: OSBMahojiCommand = {
 			iQty: options.quantity ? options.quantity : undefined,
 			powerchopping: powerchop,
 			forestry: forestry_events,
-			twitchers: twitchers_gloves || 'egg',
+			twitchers: twitchers_gloves as TwitcherGloves,
```
gc, line 228: This should definitely not need a typecast. You have it typed wrong in the options above, probably.

src/tasks/minions/woodcuttingActivity.ts
```diff
@@ -306,7 +306,7 @@ export const woodcuttingTask: MinionTask = {
 
 		// Check for twitcher gloves
 		if (twitchersEquipped) {
-			twitcherSetting = twitchers;
+			twitcherSetting = twitchers as TwitcherGloves;
```
gc, line 309: same here

src/lib/types/minions.ts
nwjgit: I believe I fixed this with the proper solution. I made them both optional and used the typed instead of string for twitchers gloves.

src/tasks/minions/woodcuttingActivity.ts
```diff
@@ -236,7 +236,7 @@ export const woodcuttingTask: MinionTask = {
 
 		let strungRabbitFoot = user.hasEquipped('Strung rabbit foot');
 		let twitchersEquipped = user.hasEquipped("twitcher's gloves");
-		let twitcherSetting = '';
+		let twitcherSetting: TwitcherGloves | undefined = undefined;
```
gc, line 239: I dont know if it should be or not, but you dont have the default here.

gc (changes requested):

nwjgit:
> Apart from the few comments, it looks good as long as its tested and feedback has been addressed

I did a round of testing with some testers a few days ago and all the feedback & bugs found were fixed.

src/tasks/minions/woodcuttingActivity.ts
nwjgit: I still have it as string in options. I fixed it. thanks for the catch

nwjgit: oh i see what you're saying now, i just removed the default entirely so that no setting is selected. Do you think I should set this to a default? like `let twitcherSetting: TwitcherGloves | undefined = 'egg';`

src/tasks/minions/woodcuttingActivity.ts
```diff
@@ -93,6 +93,7 @@ async function handleForestry({ user, duration, loot }: { user: MUser; duration:
 		const event = ForestryEvents[eventIndex];
 		const defaultEventXP = 5 * (randInt(85, 115) / 100); // used for unverified xp rates
 		let eggsDelivered = 0;
+		let beeHive = 0;
```
gc, line 96: This variable should be given a better name, just reading it, i have 0 idea what it is for or does.

gc (changes requested):

src/tasks/minions/woodcuttingActivity.ts
nwjgit: I renamed it beeHiveRepairs to more accuratley reflect what it is. It's purpose is to get a more accurate xp rate to ingame xp rates. 

`for (let i = 0; i < randInt(5, 7); i++)`
- accounts for the average of 6 beehives per event, 5-7 for variance depending on how "busy" the event is

`beeHiveRepairs += randInt(5, 10);`
- accounts for the users interaction with repairing the beehive, 5-10 for variance depending on how "busy" the event is

`eventXP[event.uniqueXP] += user.skillLevel(event.uniqueXP) * 0.3 * beeHiveRepairs;`
- beeHiveRepairs is then used here to give an accurate amount of construction xp per event based on the users con level

PR #5756, opened 2024-03-07
[TESTED/POLL INFO]Allow wesley to turn athelas into paste

I-am-TURBO:
### Description:

There's already no xp gain from turning athelas into athelas paste by having our minions do it, so it was brought to my attention that there is no reason why wesley can't do it for us.

### Changes:

turned wesley from false to true for athelas

### Other checks:

- **poll started on 04/25**

- **test bot results**


themrrobert:
I think the reason is to increase the time cost for making heat res. potions, we'll have to see what Magna things on this one; I think it was intentionally made non-Wesley'able on creation.

TURB042O:
> I think the reason is to increase the time cost for making heat res. potions, we'll have to see what Magna things on this one; I think it was intentionally made non-Wesley'able on creation.

alright. the thought was brought up to me by a few people so I figured why not. in the end you're right and magna will decide.

gc:
I have no opinion. If it passes a poll im happy to change it.

I-am-TURBO:
> I have no opinion. If it passes a poll im happy to change it.

I-am-TURBO:
> I have no opinion. If it passes a poll im happy to change it.

has the poll passed?

PR #5757, opened 2024-03-07
[TESTED]Farming spirit weed cleans only if toggled on & added subtext for Grimy

I-am-TURBO:
### Description:

Grimy spirit weed has been coded in as a item since released but was not coded correctly for farming so instead of harvesting Grimy spirit weed we harvested the cleaned version. now that isn't so. we will harvest Grimy and if toggled on it will be cleaned while farming.

### Changes:

- changed outputCrop to Grimy spirit weed
- added `cleanHerbCrop: itemID('Spirit weed'),`
- added herbLvl & herbXp
- added `spirit weed` to aliases on grimy.ts
- added subtext for Grimy spirit weed

### Other checks:


TURB042O:
> Doesn't there need to be a herb level and xp added to this?

now u mention this I do see that is needed. where do I get the #s for Korulsi & Spirit weed?

TURB042O:
> Doesn't there need to be a herb level and xp added to this?

thank you for the info :)

PR #5763, opened 2024-03-08
Updated fishing contest button to check for the items of both locations

TURB042O:
### Description:

The button just calls the normal fishing contest command with no location option causing the bot to to respond you need so many of bait to fish at location instead of checking if you have items for the other location

i updated the `fishingContestStartCommand()` function to check for the items of both locations

all-in-all this will get rid of having to manually send the command for the location you do have items for.

mad thanks to @themrrobert for this.

### Changes:

- changed `if (!loc)...` to iterate over all validLoc items putting current value in `location`
- changed (user.bank.has(location.fishing_location)) to (user.bank.amount(location.bait.id)

### Other checks:


TURB042O:
@themrrobert bro what went wrong...

PR #5765, opened 2024-03-09
[TESTED]Updated fishing contest button to check for the items of both locations

I-am-TURBO:
### Description:

The button just calls the normal fishing contest command with no location option causing the bot to to respond you need so many of bait to fish at location instead of checking if you have items for the other location

This image is an example of what it currently does now if you don't have bait before the location that it searches by default

i updated the `fishingContestStartCommand()` function to check for the items of both locations

all-in-all this will get rid of having to manually send the command for the location you do have items for.

mad thanks to @themrrobert for this.

### Changes:

- changed `if (!loc)...` to iterate over all validLoc items putting current value in `location`
- moved the section of code that registers quantity as one above the location code so that way it reads quantity instead of a number when it comes to the amount of bait needed

- linked the buy command to BSO so people can tap/click it and it'll open buy.
- added `Beginner's` to the return message if you don't own a contest rod or Beginner's tackle box (because people may have gotten confused)

### Other checks:

in this screenshot it shows he has neither item for either location, so he spawns the item needed for the 2nd location only and sends command and it went directly there as hoped for.

changed the message you get when you don't own rod or tackle box.

thank you sooo much @themrrobert for all you have done to help me on this.

### DIFF
--- a/src/mahoji/lib/abstracted_commands/fishingContestCommand.ts
+++ b/src/mahoji/lib/abstracted_commands/fishingContestCommand.ts
@@ -65,7 +65,7 @@ export async function fishingContestStartCommand(user: MUser, channelID: string,
 	}
 
 	const result = await getUsersFishingContestDetails(user);
-	if (result.catchesFromToday.length > 0) {
+	if (0) {
 		return `You already participated in the Fishing Contest today. You caught: ${result.catchesFromToday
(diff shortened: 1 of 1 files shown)


I-am-TURBO:
this edit worked in my test. but I have no clue how to turn off the daily cooldown to test the command.

PR #5771, opened 2024-03-11
Add Crystal Impling Hunting

TastyPumPum:
Adds a way to seek out crystal implings and hunt them. Currently we have to do priff content and the rate of obtaining them is quite rare. This gives players a way to specifically hunt them.


gc (commented):
Not sure about the `if (crystalImpling)`'s. It's bad for maintainability.

TastyPumPum:
> Not sure about the `if (crystalImpling)`'s. It's bad for maintainability.

Have you got any ideas? I think it would be best for users to be under /hunt, but maybe it could have its own abstracted command file and activity file then?

It's just quite different to any other hunted creatures really.

PR #5772, opened 2024-03-12
Add option to hide/unhide tame

themrrobert:
### Description:

- Adds option to hide/unhide your tame. (from /tames list)
- This is because you can easily acquire a large-ish number of tames trying to hatch several for good stats and/or shiny tames.

### Changes:

- Adds `hidden` column on Tame table
- Adds `/tames hide` command to toggle/set the `hidden` value
- Adds `hidden` option to `/tames list` command to include the hidden tames in case you do want to see the big graphic :)

### Other checks:

image
image
image


gc:
Not sure this is the best way to solve this issue of too many tames (manually hiding each one, which might be a lot). We could do pages.

PR #5776, opened 2024-03-12
Balloon cat - Removed tradeable & tradeable on ge

I-am-TURBO:
### Description:

fixed small things with Balloon cat code

### Changes:

- Removed `tradeable: true` & `tradeable_on_ge: true` from Balloon cat due to it is ment to be untradeable & has `isSuperUntradeable: true`

### Other checks:

### DIFF
--- a/src/lib/customItems/pets.ts
+++ b/src/lib/customItems/pets.ts
@@ -418,19 +418,6 @@ setCustomItem(
 	1_000_000
 );
 
-setCustomItem(
-	52_687,
-	'Balloon cat',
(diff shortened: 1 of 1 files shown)


gc:
Dont move it please. It ruins the diff, and the file is already wildly unorganized and not really meant to be anymore.

TURB042O:
> Dont move it please. It ruins the diff, and the file is already wildly unorganized and not really meant to be anymore.

gotcha
 I will not touch it. I was confused why it was down there but you have a reason

PR #5778, opened 2024-03-13
Update QP and Slayer Block Formula

TastyPumPum:
Updates the slayer block formula to correctly check for the lumbridge diary. Also updates the max_qp to the same as osrs. (With the addition of the dt2 5 points)


src/mahoji/lib/abstracted_commands/questCommand.ts
```diff
@@ -8,7 +8,7 @@ import addSubTaskToActivityTask from '../../../lib/util/addSubTaskToActivityTask
 import { minionIsBusy } from '../../../lib/util/minionIsBusy';
 import { userHasGracefulEquipped } from '../../mahojiSettings';
 
-export const MAX_GLOBAL_QP = 293;
+export const MAX_GLOBAL_QP = 303;
```
gc, line 11: This is the max QP excluding the "specific quests", which DT2 is. So you want to use `MAX_QP` (look how its defined to see what i mean)

TastyPumPum: Have updated to add the new quests similar to how dt2 is done

src/lib/data/buyables/buyables.ts
```diff
@@ -1,12 +1,12 @@
 import { Bank } from 'oldschooljs';
 
-import { QuestID } from '../../../mahoji/lib/abstracted_commands/questCommand';
 import { chompyHats } from '../../constants';
 import { CombatCannonItemBank } from '../../minions/data/combatConstants';
 import { MinigameName } from '../../settings/settings';
 import { soteSkillRequirements } from '../../skilling/functions/questRequirements';
 import { MUserStats } from '../../structures/MUserStats';
 import { Skills } from '../../types';
+import { QuestID } from '../../util/specificQuests';
```
gc, line 9: Whats with this change?

TastyPumPum: I thought it would be tidier having the specific quests in its own file rather than in the command file, I can move back that's fine

TastyPumPum: I've now moved all the specific quests back to the questcommand file, so ready for a review again

gc: It's a good idea and nice change, but in reality stuff like this can make it harder to review and create annoying merge conflicts. So refactorings are generally best left out, or down on their own separate from a feature update.

PR #5784, opened 2024-03-17
[TESTED]Display DG Buyables list by sending command without item flag

I-am-TURBO:
### Description:

to get the dg buyables list you had to type something random in the `item:` flag in order for the bot to tell you that's not a buyable item and display the items. this gets rid of the having to send a random command.
SmartSelect_20240317_111539_Discord

### Changes:

- made the `item` not required to send command
- removed the "that is not a buyable item" message

### Other checks:


I-am-TURBO:
@gc tested & works.

PR #5798, opened 2024-03-24
Allow users to exclude ironmen from MostDrops Command

Felris:
### Description:
This has been requested for a long time, since irons were giving enhanced rates for crates.
This only updated the mostdrops command for now.
image

the format of the ironmanPart is what eslint wants, assuming it's ok.
also my first pull request, lemme know if I'm doing something stupid or missed something. I did web dev for 5 years but did more backend, so my javascript skills aren't great and never worked with typescript.
### Changes:

Changed the ironmanOnly flag from bool to string
Changed param from 'ironman' to "filter"
Added "Both" (doesn't actually do anything, just defaults to ''), "Irons Only", and "Main Only" choices.

### Other checks:

### DIFF
--- a/src/mahoji/commands/tools.ts
+++ b/src/mahoji/commands/tools.ts
@@ -644,9 +644,14 @@ async function dryStreakCommand(monsterName: string, itemName: string, ironmanOn
 		.join('\n')}`;
 }
 
-async function mostDrops(user: MUser, itemName: string, ironmanOnly: boolean) {
+async function mostDrops(user: MUser, itemName: string, filter: string) {
 	const item = getItem(itemName);
-	const ironmanPart = ironmanOnly ? 'AND "minion.ironman" = true' : '';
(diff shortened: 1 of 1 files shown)


src/mahoji/commands/tools.ts
```diff
@@ -842,10 +847,11 @@ export const toolsCommand: OSBMahojiCommand = {
...
 						{
-							type: ApplicationCommandOptionType.Boolean,
-							name: 'ironman',
-							description: 'Only check ironmen accounts.',
-							required: false
+							type: ApplicationCommandOptionType.String,
+							name: 'filter',
+							description: 'Filter by account type.',
+							required: false,
+							choices: ['Both', 'Irons Only', 'Mains Only'].map(i => ({ name: i, value: i }))
```
gc, line 854: Ideally this would be a typesafe enum or array defined elsewhere.

PR #5802, opened 2024-03-25
Shorten imports and removes duplicate imports

themrrobert:
### Description:

- Shortens imports, and fixes several spots where the same file[s] is/are imported multiple times due to mismatched import paths.

### Other checks:

### Notes:

This will NOT merge to BSO easily, so allow me to handle the merge.

(Note to self: I will just accept everything from BSO, and then apply the same method used to fix master)


gc:
Did you do this manually?

themrrobert:
> Did you do this manually?

My IDE can find all instances of this and auto correct them, so it's fairly easy to do, although i did manually verify each instance on this occasion.

gc:
Extremely out of date and likely not beneficial at this point

PR #5806, opened 2024-03-28
Scroll of cleansing saves 10% of herblore Secondaries

Felris:
### Description:

Scroll of Cleansing was added with dungeoneering 3 years ago, and was never given a use. This use isn't too OP, but makes it do something a little useful at least. Scroll of cleansing saves 10% of secondaries for /mix command. Also added a /data to track how much the scroll has saved.

### Changes:

src/mahoji/commands/mix.ts - Add a check for scroll of cleansing, and if found removes 10%(rounded down) of secondaries (using the secondaries filterable). Then updates cleansing_scroll_bank, and appends a short message saying the scroll of cleansing saved ingredients.

prisma/schema.prisma, src/lib/structures/MUserStats.ts - Added cleansing_scroll_bank

src/mahoji/lib/abstracted_commands/statCommand.ts - added /data command to view cleansing_scroll_bank

### Other checks:


src/mahoji/commands/mix.ts
```diff
@@ -171,6 +187,9 @@ export const mixCommand: OSBMahojiCommand = {
 		if (boosts.length > 0) {
 			str += `\n**Boosts:** ${boosts.join(', ')}`;
 		}
+		if (hasScroll) {
+			str += ' Your scroll of cleansing enables you to save 10% of your secondaries: ';
```
nwjgit, line 191: should make this end with ` secondaries.`

Felris: Good catch, was originally planning on including a qty saved, hence the colon, but ended up not doing that. should be fixed now.

src/mahoji/commands/mix.ts
```diff
@@ -144,7 +146,21 @@ export const mixCommand: OSBMahojiCommand = {
...
 		const finalCost = requiredItems.clone().multiply(quantity);
+
+		if (hasScroll) {
+			for (const [item, qty] of finalCost.items()) {
+				if (secondaries.includes(item.id)) {
+					const saved = Math.floor(calcPercentOfNum(10, qty));
+					finalCost.remove(item.id, saved);
+					const savedBank = new Bank();
+					savedBank.add(item.id, saved);
+					userStatsBankUpdate(user.id, 'cleansing_scroll_bank', savedBank);
```
gc, line 159: This is probably unsound (prone to overwriting updates due to not awaiting), and you should do it with just 1 call.

Possibly like:

```ts
const savedBank = new Bank();
for (const item of items) {
const savings = new Bank();
cost.remove(savings);
savedBank.add(savings);
}
await userStatsBankUpdate(user.id, 'cleansing_scroll_bank', savedBank);
```

src/mahoji/commands/mix.ts
```diff
@@ -171,6 +187,9 @@ export const mixCommand: OSBMahojiCommand = {
 		if (boosts.length > 0) {
 			str += `\n**Boosts:** ${boosts.join(', ')}`;
 		}
+		if (hasScroll) {
+			str += ' Your scroll of cleansing enables you to save 10% of your secondaries.';
+		}
```
gc, line 192: Preferably just put this in the boosts array. We're moving away from the `str += ` because its hard to keep track of formatting them.

gc (changes requested):
Perhaps also just go through the BSO-specific herblore things and see if theres any situations this would be OP or shouldnt apply.

PR #5807, opened 2024-03-29
changed flag to work from bank for Drygore axe, Wisp-buster & Divine hand

I-am-TURBO:
### Description:

title says all

### Changes:

title says all.

### Other checks:


themrrobert:
The respective code where these boosts are applied also needs to be looked at / updated to account for banked items.

Also, I'm not sure if we decided how to handle the wisp buster/divine hand conflict yet, so for now it might be best to leave that one alone.

Spoke with turbo on this one and came to resolution.

PR #5809, opened 2024-04-02
[SQL] Change some sums to bigint where it's possible they will go out of range

themrrobert:
### Description:

Some data / stats are failing to return due to prisma out of range errors.

Corrected this, and also changed a few xp_totals where they could possibly exceed 2.14b

### Changes:

Changed ::int to ::bigint in a few places

### Other checks:


nwjgit:
changing these from int to bigint is causing an error with toKMB()

"TypeError: Cannot mix BigInt and other types, use explicit conversions"

PR #5821, opened 2024-04-08
Add a blackjack command to gamble

TastyPumPum:
Adds the blackjack game to the gamble command.
Allows players to win up to 2x their stake by beating the dealer/bot.


AvariceBrah:
you madlad, i love it

gc:
Big tip: Try to code this in a fully abstracted way, so that its a lot easier to test (which is important so we know exactly for a fact the odds/rates of this, and just generally to test it for bugs), e.g. such as how `hunterActivity.ts` is in bso. I like the blackjack idea

src/mahoji/lib/abstracted_commands/blackjackCommand.ts
```diff
@@ -0,0 +1,211 @@
...
+		for (const value of values) {
+			deck.push({ suit, value });
+		}
+	}
+	return deck;
+}
+
+// Function to shuffle the deck
+function shuffleDeck(deck: Card[]): void {
+	for (let i = deck.length - 1; i > 0; i--) {
```
themrrobert, line 50: Needs to be i >=0 otherwise the loop where i = 0 is not executed, and you're just hoping the shuffle chooses card zero.

Basically if you don't do >=0, then the 0th card ends up unshuffled a disproportionate amount of the time.

src/mahoji/lib/abstracted_commands/blackjackCommand.ts
```diff
@@ -0,0 +1,211 @@
...
+		const j = Math.floor(Math.random() * (i + 1));
```
themrrobert, line 51: Should use a nodeCrypto random, try cryptoRand() from util.ts, and you don't need to floor it.

Alternatively, randFloat() from util.ts also uses nodeCrypto, but why use two functions when you can use one? :D

src/mahoji/lib/abstracted_commands/blackjackCommand.ts
```diff
@@ -0,0 +1,211 @@
...
+	for (const suit of suits) {
+		for (const value of values) {
+			deck.push({ suit, value });
+		}
+	}
+	return deck;
+}
+
+// Function to shuffle the deck
+function shuffleDeck(deck: Card[]): void {
```
themrrobert, line 49: The shuffle looks pretty good, but I would still shuffle the deck at least 2-3 times.  (either multiple calls to shuffle deck, or more likely just run the whole loop 2-3 times)

src/mahoji/lib/abstracted_commands/blackjackCommand.ts
```diff
@@ -0,0 +1,211 @@
...
+	// Create and shuffle the deck
+	const deck = createDeck();
+	shuffleDeck(deck);
+
+	// Deal initial hands
+	const playerHand = [dealCard(deck), dealCard(deck)];
+	const dealerHand = [dealCard(deck), dealCard(deck)];
+
+	// Check for player blackjack
+	if (calculateHandValue(playerHand) === 21) {
```
themrrobert, line 128: Blackjack should pay 2.5x  (1.5x their bet in profit). 

Or a push if dealer also has blackjack. 

You could get crazy with insurance, even-money on blackjack vs A, etc, but that's probably not necessary :D

src/mahoji/lib/abstracted_commands/blackjackCommand.ts
```diff
@@ -0,0 +1,211 @@
...
+];
+
+interface Card {
+	suit: string;
+	value: string;
+}
+
+// Define constants for suits and values
+const suits = ['Hearts', 'Diamonds', 'Clubs', 'Spades'];
+const values = ['2', '3', '4', '5', '6', '7', '8', '9', '10', 'Jack', 'Queen', 'King', 'Ace'];
```
themrrobert, line 35: It MIGHT make formatting the game easier if you use a single char for card values (and suits?).

eg. 10 = 'T', J, K , Q, A

It would be REALLY cool to dynamically generate the result like slots. You don't have to draw the entire card, just a simple box around painted unicode letters, 3♠, A♥  etc. Paint spades/clubs black, and hearts/diamonds red. T♣

Look at slots for simple graphics stuff, and the bank code has examples of drawing text (including unicode symbols) onto canvas)

themrrobert (commented):
I know this is still a work in progress, and I haven't gone thru all of it, but here are some good notes :)

PR #5835, opened 2024-04-17
Add toggle for Size Matters task

themrrobert:
### Description:

Add toggle for 'Size Matters' task similar to the Scroll of Longevity toggle.

### Changes:

- Adds disabled bitfield, and updates slayer code to check for it.

### Other checks:


gc:
Good to merge assuming this is something people want.

PR #5856, opened 2024-04-27
Fix crystal imp lb and status

TastyPumPum:
Changes the ID of crystal impling so it doesn't conflict with bluegill of aerial fishing.

Also updates the minion status message to remove the quantity.


themrrobert:
Important note: The crystal impling PR was not compatible with the BSO hunter codebase, too many conflicts etc.

So we should try to avoid merging these changes to BSO until the crystal implings are implemented there.

Issue #5850, opened 2024-04-26
Drygore Axe not counting as dwarven axe for wintertodt


Felris:
## 🐛 Bug Report

Dwarven great axe gives a 2x boost to Wintertodt. Drygore axe does not, even though it's an upgrade.

PR #5863, opened 2024-04-29
Wintertodt drygore axe

Felris:
### Description:

Gives drygore axe the same boost at wintertodt as dwarven greataxe as it's an upgrade. Does not use materials. This is consistent with how the drygore axe functions with /chop.
Fixes https://github.com/oldschoolgg/oldschoolbot/issues/5850

### Changes:

Added drygore axe to hasEquippedOrInBank check. If you'd prefer similaritems let me know. Just not sure if this would be needed anywhere else.

It still shows "2x for Dwarven greataxe" regardless of which axe you have. But this is consistent with how /chop works when drygore axe is disabled. So don't think it's really an issue.

Code_2024-04-29_16-13-27

### Other checks:


themrrobert:
I think making Drygore Axe a similarItem of Dwarven greataxe is the way to go here.

Felris:
> I think making Drygore Axe a similarItem of Dwarven greataxe is the way to go here.

Done

PR #5873, opened 2024-05-11
Filter out completed quests

TastyPumPum:
When doing activities quest and picking a quest, currently it just shows all the quests and it's hard to keep track of what's completed and what isn't.

Following the #bso-vote poll, this hides the quests you have completed from the list and only shows ones you haven't completed. Then when every quest is completed it shows (completed) at the end instead of providing an empty list.

image


I-am-TURBO:
this didn't work for BSO

PR #5874, opened 2024-05-12
Enigmar boss - New Elder clue dropping boss!

themrrobert:
### Description:

- Adds Enigmar boss!

### Changes:

- `/create name:Enigma tokens` to convert clue scrolls into tokens used to kill the boss.
- Get the `Enigmatic orb` from the boss!
- Create `Quantum transmuter` to increase your Farming yield! (Restores the pre-Plopper nerf max rate)
- Cannot be killed with DTD's currently

### Misc
- Removes duplicate ownedMaterials() function
- Fixes Venatrx & Akumu `allItems` CL

### Other checks:


src/tasks/minions/farmingActivity.ts
```diff
@@ -61,6 +61,10 @@ async function farmingLootBoosts(
 			);
 		}
 	}
+	if (user.owns('Quantum transmuter')) {
+		bonusPercentage += 80;
```
gc, line 65: Should this be 40% so we can put the other 40$ elsewhere?

themrrobert: The original boost was 1.2x, which stacked with the 4x to yield to an additional 80% boost on top of the +300%.
- 100 x 4 => 400 x 1.2 => 480
- 100 + 380% = 480

So it's actually not as big as it seems, and this would already make 4 different farm boosts (plopper / master farming cape / arcane harvester, quantum transmuter), just to finally get back to the 2  boosts we had previously (Plopper & harvester).

So if we even want another farming boost, we can add it sure, but I think this should be 80%.

PR #5884, opened 2024-05-18
Removed doubled entries in /cl & Collections

I-am-TURBO:
### Description:

Title says it.

### Changes:

- removed Camdozaal, Chompy, fossil Island notes & creature creation from 'other' cl
- removed chaos druids & champ challenge from 'skilling' cl

### Other checks:


themrrobert:
This is going to affect the number of CL slots + CL completion, I don't think we should do this.

What is causing the problem? Is the default monsters in OSJS being loaded in?

Instead of removing these, we should probably just do a filter when building the autocomplete list, to avoid putting the same entry in there twice. This would solve the visual issue, without affecting the CL / gameplay.

I-am-TURBO:
> This is going to affect the number of CL slots + CL completion, I don't think we should do this.
> 
> What is causing the problem? Is the default monsters in OSJS being loaded in?
> 
> Instead of removing these, we should probably just do a filter when building the autocomplete list, to avoid putting the same entry in there twice. This would solve the visual issue, without affecting the CL / gameplay.

PR #5936, opened 2024-07-03
Add quantity to Wintertodt

TastyPumPum:
Allows users to specify the quantity of Wintertodt. If no quantity is provided do a max trip as always. Also adds "Loot from x Wintertodt" into the loot image.

### DIFF
--- a/src/lib/util/minionStatus.ts
+++ b/src/lib/util/minionStatus.ts
@@ -322,7 +322,8 @@ export function minionStatus(user: MUser) {
 			}. ${formattedDuration} Your ${Emoji.Cooking} Cooking level is ${user.skillLevel(SkillsEnum.Cooking)}`;
 		}
 		case 'Wintertodt': {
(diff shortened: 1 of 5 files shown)


src/lib/util/minionStatus.ts
```diff
@@ -322,7 +322,8 @@ export function minionStatus(user: MUser) {
 			}. ${formattedDuration} Your ${Emoji.Cooking} Cooking level is ${user.skillLevel(SkillsEnum.Cooking)}`;
 		}
 		case 'Wintertodt': {
-			return `${name} is currently fighting the Wintertodt. ${formattedDuration}`;
+			const data = currentTask as ActivityTaskOptionsWithQuantity;
+			return `${name} is currently fighting Wintertodt  ${data.quantity}x times. ${formattedDuration}`;
```
gc, line 326: There's 2 spaces here

TastyPumPum: Fixed good spot

Issue #5950, opened 2024-07-12
DT2 item incorrect data


nwjgit:
## 🐛 Bug Report
https://github.com/oldschoolgg/oldschoolbot/commit/6b02a88d7ccdca31a82ec71526c363b0a248c450 introduced item switches which switched the dt2 rings from beta id to the proper id but the price and associated wiki page seem to use the beta id information.

- before:
![Discord_jluJI2sbdl](https://github.com/user-attachments/assets/b4b1a1aa-e882-4283-abf2-bdbf9b8bfa38)
- after:
![Discord_XVLAK1AOSv](https://github.com/user-attachments/assets/56b4d8d0-7a47-47dc-be44-51850c6a4ff9)

- Another issue is we can still create the beta versions of these items. as seen below. These beta ids should probably be added to the item ignore list in OSJS.
![Discord_Nf2U30d87P](https://github.com/user-attachments/assets/4e092fb8-18f2-44bb-b4b3-d8493d80c057)

other issues:
- `/gear stats gear_setup: Venator ring` shows the beta icon and not the current icon
- Soulreaper axe still using the incorrect id

Follow up to: https://github.com/oldschoolgg/oldschoolbot/issues/5284

gc:
The IDs have not changed. The beta versions now just match the data of the real ones. Removing the beta IDs would (same as before) require all the existing items to be migrated to the new one, which is a pain to do.

nwjgit:
> The IDs have not changed. The beta versions now just match the data of the real ones. Removing the beta IDs would (same as before) require all the existing items to be migrated to the new one, which is a pain to do.

It's a pain to do but its probably the only real solution. The one currently implemented isn't doing the trick as seen in the original post.

Issue #5284, opened 2023-08-26
DT2 Rings are using the beta version


nwjgit:
## 🐛 Bug Report

Using the Bellator ring as an example:

This is the current ring we have in OSB/BSO: https://oldschool.runescape.wiki/w/Bellator_ring_(beta)

This is the current ring in the live game: https://oldschool.runescape.wiki/w/Bellator_ring

![Discord_w4ItNGHSCs](https://github.com/oldschoolgg/oldschoolbot/assets/69014816/4a1bda9b-63b2-4f53-9e87-d77314c59004)

This also is whats causing the rings to not be tradable: https://github.com/oldschoolgg/oldschoolbot/issues/5282#issue-1867774779

6enny6:
fixed in #5283

nwjgit:
> fixed in #5283

that doesn't solve the issue of rings that have already been created.

mylife212:
Bump.  hope this is fixed soon and not months/years from now, considering the untapped market demand for the bis rings.

nwjgit:
Updating this here because it is in the realm of DT2 content. The Soulreaper Axe also has the same issue of using the beta ID.

Soulreaper axe in OSB/BSO: https://oldschool.runescape.wiki/w/Soulreaper_axe_(beta)

Soulreaper axe in live game: https://oldschool.runescape.wiki/w/Soulreaper_axe

gc:
Fixed in next update

PR #5953, opened 2024-07-14
Fix bank with item swaps

nwjgit:
### Description:
Add a method to swap items with the wrong ids with the proper id & properly update bank/gear/sac/cl data. 
This fixes the dt2 beta id issue and any future similar issues. 

Currently this is only under `/tools user fixbank` I would recommend maybe having this auto run under some command such as when the user runs `/bank` or `/cl`

### Changes:
- Add `/tools user fixbank` to OSB
- Add code inside fixbank that handles swapping item ids for bank/gear/sac/cl data
- Use the real id for the ring creation to prevent future beta id rings from entering the game
- Set the rings to use itemAliases
- Make `/sacrifice` `/sell` `/trade` use item aliases to always select the correct item.

### Other checks:
closes: https://github.com/oldschoolgg/oldschoolbot/issues/5950
closes: https://github.com/oldschoolgg/oldschoolbot/issues/5284
related: https://github.com/oldschoolgg/oldschoolbot/commit/6b02a88d7ccdca31a82ec71526c363b0a248c450

### DIFF
--- a/src/lib/data/Collections.ts
+++ b/src/lib/data/Collections.ts
@@ -1100,7 +1100,7 @@ export const allCollectionLogs: ICollection = {
 					...Monsters.Vardorvis.allItems,
 					...Monsters.DukeSucellus.allItems
 				],
-				items: [...theLeviathanCL, ...theWhispererCL, ...vardorvisCL, ...dukeSucellusCL]		
+				items: [...theLeviathanCL, ...theWhispererCL, ...vardorvisCL, ...dukeSucellusCL]
 			},
 			Creatables: {
 				counts: false,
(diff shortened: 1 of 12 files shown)


gc (changes requested):
I'm not 100% sure on this yet, we would benefit from repairItems function in osb (although I just redid that function entirely on bso), but repairItems is only used for removing items, not swapping. I believe it may be more beneficial to do the item swapping entirely in SQL scripts, so that its fixed for everyone on the bot all at once, and people don't have to run a command or get errors. There's also the G.E that has the items in it. Is there a more practical issue this is fixing or is the goal just to have the IDs correct?

PR #5998, opened 2024-08-12
QBD Boost fixes + Dragon Boost Mislabel fix

DaughtersOfNyx:
### Description:

Three different bugs fixed here relating to Dragons/QBD specifically.

### Changes:

1. Changed QBD's base monster from Vorkath -> KBD. This removes the incorrect application of Undead boosts to QBD (Salve amulet etc)
2. Added a check before applying generic dragon boost to QBD similar to Vorkath that allows the Lance to actually provide its boost. Previously it was impossible to get the boost from lance as the boss requires range. This also involved explicitly adding the Dragon Hunter item boosts to QBD.
3. Fixed the incorrect info in the show_info subcommand of /k from 20% -> 15% as this wasn't changed back when the boost was reduced I suppose. Already fixed in OSB.

### Other checks:


gc:
Has conflicts. Looks good to merge though

PR #6005, opened 2024-08-16
Fishing rework 

Arodab:
### Description:

The fishing skill in the bot currently works nothing like in the real game, where fishing spots can have different types of fish. This PR makes the skill work exactly like in OSRS, changing it to a tick system, which allows methods like 2t swordfish and 3t barb to be added. It also allows spirit flakes to be used for a chance to catch more fish.

### Changes:

- Grouped fish into fishing spots (e.g. Tuna/Swordfish).
- Added optional `powerfish` and `spirit_flakes` parameters to the `fish` command. If **powerfish** is `True`, the minion will receive more exp at the cost of not banking the fish, and will use tick manipulation methods if applicable.
- Removed the +9min boost from having a fish barrel **only if** powerfishing, trips are still longer if banking the fish.
- Removed other nonsense boosts like pearl fishing rods, and added dragon/crystal harpoon boosts only for fish that they actually affect.

### Other checks:


gc:
I pushed some changes and discussed in discord about stuff that should be changed, feel free to ping me in discord for any question.

PR #6010, opened 2024-08-20
Fix elite clue drop rates for TOB

Arodab:
### Description:

TOB currently uses the entry mode drop rate for elite clues, this PR adjusts the rates to match OSRS.

### Changes:

-Changed elite clue rate to 3/25 for normal mode and 3.5/25 for HM.

### Other checks:

### DIFF
--- a/src/lib/simulation/tob.ts
+++ b/src/lib/simulation/tob.ts
@@ -109,7 +109,7 @@ class TheatreOfBloodClass {
 
 			clueRate = 3.5 / 25;
 		}
-		console.log(clueRate)
+
 		if (Math.random() < clueRate) {
 			loot.add('Clue scroll (elite)');
 		}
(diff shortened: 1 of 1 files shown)


src/lib/simulation/tob.ts
```diff
@@ -98,16 +98,19 @@ class TheatreOfBloodClass {
...
 				loot.bank[Number.parseInt(itemID)] = Math.ceil(loot.bank[Number.parseInt(itemID)] * 1.15);
 			}
 			// Add HM Tertiary drops: dust / kits
 			loot.add(HardModeExtraTable.roll());
+
+			clueRate = 3.5 / 25;
 		}
 
-		if (roll(25)) {
+		if (Math.random() < clueRate) {
```
nwjgit, line 113: You could just do `if (roll(clueRate)) {` here

nwjgit: roll() returns true or false so a comparison to a number wouldn't work. You could convert 3/25 and 3.5/25 to percentages and use percentChance()

nwjgit: also correct me if I'm wrong but doesn't Math.random() return a float between 0 and 1? so this would always be true.

Issue #5913, opened 2024-06-04
Mystery Box item search issue


nwjgit:
https://github.com/oldschoolgg/oldschoolbot/pull/5846 expanded the mysterybox icon to include all items from mystery boxes when doing an item search.n 

Non-ironmen shouldn't have a mysterybox icon displayed for ironman only PMB pets. Something with the code here is broken https://github.com/oldschoolgg/oldschoolbot/blob/bso/src/lib/events.ts#L153.

Good first issue if anyone wants to try and fix, if not i'll troubleshoot & fix sometime in the future.

nwjgit:
fixed.

PR #6024, opened 2024-08-28
Dynamic PMB & item search fix

nwjgit:
### Description:
Create PMB dynamically based off all pets cl. This prevents the delay of OSB pets being added to PMB table in the future. 

### Changes:
- Dyanmic PMB
- Rename `allMbTables` to `combinedTmbUmbEmbTables` for better understanding of whats in it.
- use `allMbTables` to represent the main 6 and `allIronmenMbTables` to represent the main 6 + ironmen unique pets on PMB
- correct the item search to only show mysterybox for unique ironmen PMB pets to ironmen only
- Add Brain lee to leagues cl

### Other checks:
- Nexling has never been on PMB, i'm guessing since its the pet for Nihilz, but this PR would add it. I can remove it if there was a reason to leave it off. Also running a poll here: https://discord.com/channels/342983479501389826/1032668754561224734/1278203213488717855

closes: https://github.com/oldschoolgg/oldschoolbot/issues/5913


gc:
I think the delay is is desirable, so that the first person to get the pet has to earn it legitimately. I'd rather it get polled just to be sure people actually want this, if people want it then its fine

nwjgit:
> I think the delay is is desirable, so that the first person to get the pet has to earn it legitimately. I'd rather it get polled just to be sure people actually want this, if people want it then its fine

I would generally agree, the issue is that when osjs item DB updated it added the coliseum pet to UMB since the pet wasn't on the all pets cl. So about 10ish people got the pet before coliseum was even coded or in PMB. A good example of this is right now is Quetzin.
Discord_YjbcCESIGY

I'll run a poll to get feedback though.

PR #6029, opened 2024-08-29
bso unique minionKill.ts fixes

nwjgit:
### Description:
Fix ability to chin & use Superior dwarf multicannon
### Changes:
- Change the check to not remove chin from combat options when not on task.
- Update Superior dwarf multicannon code to properly work with combat update
  - Update the message to reflect single or multi combat
### Other checks:


nwjgit:
Closing since this should be handled with the minionKill refactor

Issue #5831, opened 2024-04-17
No return trip on champions challenge 


DarkBugz:
No return trip at all let alone a ping

PR #6036, opened 2024-09-01
Bug fix for cl completion & maybe roles?

nwjgit:
### Description:
- Sometimes when a minion attempts to return from a trip and the loot contains the last item of the cl this query causes the trip to not return properly and not post to notifications.
- I also noticed that fetchMultipleCLLeaderboards() is utilized for sync roles for top collector. I can't verify due to lack of perms in discord but I wouldn't be surprised if this is the cause of alot of the role issues for `/rp action sync_roles` since top collector is ran first. 

### Changes:
- correct the query in fetchMultipleCLLeaderboards()
### Other checks:

closes: https://github.com/oldschoolgg/oldschoolbot/issues/5831


gc:
Can you explain how this query is corrected?

nwjgit:
> Can you explain how this query is corrected?

I traced this issue through console logs and found that this query is the reason some trips fail to return if the trip is returning with the collection log's final unique item. Restructuring the query no longer makes the trip fail.

Steps to re-create the scenerio
- Start with a fresh minion `/minion buy` or run this twice: `/minion ironman` 
- run: `/testpotato spawn collectionlog: true items: 1 Earth warrior champion scroll, 1 Ghoul champion scroll, 1 Giant champion scroll, 1 Goblin champion scroll, 1 Hobgoblin champion scroll, 1 Imp champion scroll, 1 Jogre champion scroll, 1 Lesser demon champion scroll, 1 Skeleton champion scroll, 1 Zombie champion scroll`
  - This adds all 10 champ scrolls to your cl and bank
- run: `/activities other activity: Champions Challenge`

Your minion never returns and you are hit with this console error:
Code_HcVwqzcvfk

gc:
I believe those errors are just because you don't have the intarray extension installed in your postgres.  Run this query in your postgres and it should fix it: `CREATE EXTENSION IF NOT EXISTS intarray;`

PR #6048, opened 2024-09-14
Add a way to trade Scurrius' spines for combat xp

Arodab:
### Description:

Scurrius' spines can be traded for antique lamps, which can be used for combat xp.
Book of arcane knowledge was updated to give more xp in osrs months ago.

### Changes:

-Added Antique lamp (Historian Aldo) to /lamp, giving 3.5k xp to prayer or 5k exp to other combat skills. Can be created with /create, costing 1 Scurrius' spine.
-Updated the magic/rc xp from Book of arcane knowledge from 11->15 and 4->5.

### Other checks:


nwjgit:
Don't you have to turn the spine in for a lamp first?

Arodab:
> Don't you have to turn the spine in for a lamp first?
> 
> https://oldschool.runescape.wiki/w/Antique_lamp_(Historian_Aldo)

Yes, but the lamp is not even bankable, has anyone ever not used it immediately? It was probably done this way in osrs to prevent people from accidentally breaking their builds, I don't see a reason to make it more inconvenient in the bot by adding an extra step.

nwjgit:
> > Don't you have to turn the spine in for a lamp first?
> > https://oldschool.runescape.wiki/w/Antique_lamp_(Historian_Aldo)
> 
> Yes, but the lamp is not even bankable, has anyone ever not used it immediately? It was probably done this way in osrs to prevent people from accidentally breaking their builds, I don't see a reason to make it more inconvenient in the bot by adding an extra step.

I would still opt for adding the lamp, can easily be done with /create. Skipping steps like that ruins the fun for some people, myself included going for unique sac others bank log etc etc. I recently went through something similar with another contrib over if ore pack's from volcanic mine should be auto open or not for "convenience sack".  I don't see a reason to deviate from how it works in game.

Arodab:
> > > Don't you have to turn the spine in for a lamp first?
> > > https://oldschool.runescape.wiki/w/Antique_lamp_(Historian_Aldo)
> > 
> > 
> > Yes, but the lamp is not even bankable, has anyone ever not used it immediately? It was probably done this way in osrs to prevent people from accidentally breaking their builds, I don't see a reason to make it more inconvenient in the bot by adding an extra step.
> 
> I would still opt for adding the lamp, can easily be done with /create. Skipping steps like that ruins the fun for some people, myself included going for unique sac others bank log etc etc. I recently went through something similar with another contrib over if ore pack's from volcanic mine should be auto open or not for "convenience sake". The bank thing isn't really an issue either since we have alot of non-bankable items in bank and unlimited bank space. I don't see a reason to deviate from how it works in game.

The bank thing was just to say everyone uses it immediately, but the sac log is fair, I forgot it was even a thing. So many other parts of the bot skips steps like this when there's actually a reason not to, so I just didn't think anyone would care if there was no real reason in game.

src/lib/data/itemAliases.ts
```diff
@@ -29,9 +29,6 @@ export function setItemAlias(id: number, name: string | string[], rename = true)
 			id
 		});
 	}
-	if (Items.get(id)!.name !== firstName) {
-		throw new Error(`Failed to set item alias for item ${id}`);
-	}
```
gc, line 34: Removed this why?

Arodab: That wasn't part of the code when I made this branch, from what I can tell. Added it though, and fixed conflicts.

PR #6061, opened 2024-09-21
Fix ash sanctifier

nwjgit:
### Description:
- Fix for ash sanctifier on BSO (The same logic can be applied to OSB so long as the multiplier is turned to false)
### Changes:
- Change ash sanctifier multiplier to true, to receive BSO 5x boost to xp.
- Pass ashesSanctified back to monsterActivity for removal from updateBank so the user doesn't retain ashes that are sanctified.
### Other checks:

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -27,9 +27,10 @@ RUN cp src/config.example.ts src/config.ts
 ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.9.0/wait /wait
 RUN chmod +x /wait
 
-CMD /wait && \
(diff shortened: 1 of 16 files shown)


gc:
If this bug effects OSB, the fix should go to osb, then get merged to bso. Otherwise we need to manually PR this same change to osb, then it gets remerged back to BSO anyway.

nwjgit:
> If this bug effects OSB, the fix should go to osb, then get merged to bso. Otherwise we need to manually PR this same change to osb, then it gets remerged back to BSO anyway.

I does affect both. This was made since the update went live on bso before osb so wasn't 100% sure. I'll remake for osb.

Issue #6072, opened 2024-09-27
Phantom Muspah & BOFA


oHSugge:
## 🚀 Feature Proposal
Currently unable to use BOFA at muspah due to getting an "invalid weapon" error.
can we get this fixed so the BOFA negates arrow req, and is able to be used at this boss?

PR #6073, opened 2024-09-27
Don't use arrows when using bowfa(c)

nwjgit:
### Description:
- Bow of faerdhinen (c) doesn't require arrow usage
### Changes:
- update checkRangeGearWeapon() for bowfa
- update `if (monster.projectileUsage?.required)` to work with bowfa logic
### Other checks:
- BSO RELATED: This may require special attention for BSO to not allow negation of hellfire arrow usage for some custom content and other custom bso mechanics.
- closes: https://github.com/oldschoolgg/oldschoolbot/issues/6072


src/lib/util.ts
```diff
@@ -385,6 +393,7 @@ export function checkRangeGearWeapon(gear: Gear) {
 			projectileCategory[0]
 		}-based weapons, you can use: ${projectileCategory[1].items.map(itemNameFromID).join(', ')}.`;
 	}
+	console.log(`weapon ${weapon}, ammo ${ammo}`);
```
gc, line 396: Log left here

nwjgit: Fixed

PR #6086, opened 2024-10-07
Gear Images for New Wiki

nwjgit:
### Description:
- Script to generate and automatically update gear images for wiki when `yarn wiki` is ran. Allows for use with variables already inilitzed elsewhere in the bot or you can construct your own right in geatSetups.ts (See exampleGear)

### Changes:
- generateGearImages() added to `yarn wiki` script
- change generateGearImage() to handle an undefined user
- Change the cox.mdx script as an example

### Other checks:
Updated Wiki:
chrome_uf8zLFFdfx


gc:
I feel like this could be implemented in a more automatic way, like

```wiki.md
[[gear:Twisted bow,Dragon boots, Bandos chestplate
```

in our plugin, we get that list of items, and generate an image from it.

```ts
const items = match.split(":")[1].split(",").map(getItemOrThrow);
```

Then we save that image with a certain file name, like all the item ids sorted and put together, e.g. `353-5931-3932.png`, then the plugin can put an image element there directly (like we do for item icons)

So that no manual work is needed to add new images.

nwjgit:
> I feel like this could be implemented in a more automatic way, like
> 
> ```lisp
> [[gear:Twisted bow, Dragon boots, Bandos chestplate]]
> ```
> 
> in our plugin, we get that list of items, and generate an image from it.
> 
> ```ts
> const items = match.split(":")[1].split(",").map(getItemOrThrow);
> ```
> 
> Then we save that image with a certain file name, like all the item ids sorted and put together, e.g. `353-5931-3932.png`, then the plugin can put an image element there directly (like we do for item icons)
> 
> So that no manual work is needed to add new images.

This would only work for dynamically generated stuff inside wiki.ts though right? That would probably work fine for something like rendeCoxMarkdown() but what about the stuff that isn't generated with wiki.ts

gc:
So heres some explanation:

1. [[]] features don't have to be in generated content. The wiki script literally just edits the .md files same as a human would, they don't do any special thing at all. Literally just generates a markdown string, edits it into the file.
2. The items.ts file is my plugin for my custom [[]] codes. The way it works is dead simple, you get the string of what's inside it, then you can convert the [[]] into anything you want, i.e html. So the way [[Twisted bow]] works is: i see there's some string inside the brackets, i check if the string happens to be the name of an item, if it is, i replace it with an  html tag. that's it.
3. Note that markdown (anywhere on the internet), gets converted to html so you can view it in the browser. So a #Title becomes a <h1> etc, and the plugin interacts here and decides how to turn certain stuff, the text with [[]], into html. 

If you check the plugin file, you can copy the others pretty easily, except yours will have to generate & save an image, then put the img html. (Async should work?)

The images it generates would be saved to git. 

If you do implement it youre welcome to ask in discord for any help/question.

Issue #6092, opened 2024-10-08
Grandmaster Araxxor CA's are way too rare, compared to in-game.


mylife212:
## 🐛 Bug Report
The current rate for these Combat Achievements are much rarer than others, and when compared to how difficult they are in-game with max gear - are not accurate.
![image](https://github.com/user-attachments/assets/5a760777-62f8-4e79-845b-ca7b04c1d34d)

I believe the rates should be changed to: 1/150, 1/200 and 1/250 respectively. This reflects their difficulty while ensuring you don't go thousands of kills without a CA that would have realistically been obtained already.

Issue #6091, opened 2024-10-07
Krystallia can assign Araxxor boss tasks despite not being found in wildy


mylife212:
## 🐛 Bug Report

As title says, Krystallia boss tasks are wilderness-only because shes a wildy master. She can currently assign Araxxor - rendering the task uncompletable. 
![image](https://github.com/user-attachments/assets/12115c64-b478-4e64-821b-61393fd3af80)
![image](https://github.com/user-attachments/assets/66c49002-610c-474a-bae8-fbfa61163eae)

Issue #6089, opened 2024-10-07
Amount of extended Araxyte is too high


mylife212:
## 🐛 Bug Report

In game, extended Araxytes are 200-250 per task. They are currently above this: 
![image](https://github.com/user-attachments/assets/d1064e3e-0f23-4889-8922-70b8df7bf6a8)

PR #6095, opened 2024-10-08
Araxxor adjustments

nwjgit:
### Description:
- Araxxor adjustments

### Changes:
- Adjust all CA to be closer to ingame rates
- Add times for all time based arraxor CA
- Remove a duplicated CA
- Remove araxxor from wilderness slayer
- Correct extended araxyte from going outside of 200-250
- Adjust the Noxious halberd logic to always make the piece given the one thats missing from the next full halberd
  - ex: you have 6 blade, 6 pommel, 7 point and you get 3 blades to drop this trip. It will filter it so you have 8 blade 7 pommel 7 point. next trip you get a point and it filters to a pommel making it 8 blade 8 pommel 8 point.
### Other checks:
closes: https://github.com/oldschoolgg/oldschoolbot/issues/6089
closes: https://github.com/oldschoolgg/oldschoolbot/issues/6091
closes: https://github.com/oldschoolgg/oldschoolbot/issues/6092


gc:
To fix the conflict, merge master then:
```
git rm --cached -r . 
git add .
```

I feel like the new noxious pieces code could probably be simpler. e.g. sort the 3 items ascending by how many of each they own, and always give the first item in that list (the one they own least of)?

nwjgit:
> To fix the conflict, merge master then:
> 
> ```
> git rm --cached -r . 
> git add .
> ```
> 
> I feel like the new noxious pieces code could probably be simpler. e.g. sort the 3 items ascending by how many of each they own, and always give the first item in that list (the one they own least of)?

Fixed, and I took another go at the noxious piece code, I believe it's simpler now and I just tested it in testing server so its good to go.

Arodab:
Can you remove the
```
customMonsterHP: 575,
combatXpMultiplier: 1.2,
```
from Araxxor? I believe these were just copied from Scurrius. Not sure how the slayer xp is handled, but since that's correct I assume the bot knows about the +67.5% bonus xp, so my guess is removing those will make the combat xp accurate too (should be 4x the slayer xp, around ~240k/hr, which is now the best melee xp in osrs without alts).

src/lib/minions/data/killableMonsters/bosses/misc.ts
```diff
@@ -704,14 +706,28 @@ const killableBosses: KillableMonster[] = [
 			if (loot.has('Coagulated venom') && (ownedItems.has('Coagulated venom') || ownedItems.has('Rax'))) {
 				loot.set('Coagulated venom', 0);
 			}
+
+			loot.addItem(getOSItem('Noxious point').id, 4);
+			loot.addItem(getOSItem('Herbi').id, 3);
```
gc, line 711: ?

nwjgit: I was using that to debug, I removed it.

src/lib/minions/data/killableMonsters/bosses/misc.ts
```diff
@@ -704,14 +705,25 @@ const killableBosses: KillableMonster[] = [
...
+						.sort((a, b) => a.owned - b.owned);
+
+					const targetPiece = sortedPieces[0].piece;
+					loot.set(targetPiece, (loot.amount(targetPiece) || 0) + 1);
+					loot.set(noxPieces[i], loot.amount(noxPieces[i]) - 1);
+
+					ownedCount[noxPieces.indexOf(targetPiece)]++;
+					lootCount[i]--;
+				}
+			}
```
gc, line 726: I'm still not super happy with how this is coded, but its not a big deal if it works and I dont want to make you redo it again. BTW: The `|| 0` is not doing anything (e.g. its doing `0 || 0`).

PR #6104, opened 2024-10-09
added Long bone and Curved bone to prayer

00justas:
### Description:

- added Long bone and Curved bone to prayer

### Changes:

- added Long bone and Curved bone to prayer

### Other checks:

### DIFF
--- a/src/lib/skilling/skills/prayer.ts
+++ b/src/lib/skilling/skills/prayer.ts
@@ -41,6 +41,18 @@ const bones: Bone[] = [
 		inputId: itemID('Jogre bones')
 	},
 	{
+		name: 'Long bone',
+		level: 1,
+		xp: 15.1,
+		inputId: itemID('Long bone')
+	},
+	{
+		name: 'Curved bone',
+		level: 1,
+		xp: 15.1,
+		inputId: itemID('Curved bone')
+	},
+	{
 		name: 'Zogre bones',
 		level: 1,
 		xp: 22.5,


gc:
What is the purpose/reason for this?

00justas:
Its a missing option/feature from osrs. Divine waters also use this array list in bso and it was requested to add these to the divine water list

DayV-git:
This would put long and curved bones twice in /offer with two different meanings

nwjgit:
> This would put long and curved bones twice in /offer with two different meanings

This wouldn't be the case,
`offerables` under offer.ts is a set so it will only list long and curved bone once. long and curved bone are also listed under `specialBones` and are handled first.

gc:
Can you check that its not the case? My worry is just that people will accidentally offer/bury these bones, and theres literally no purpose ever for this in OSB. If someone comes along and accidentally does a /bury trip of all their saved up long bones, they wont be happy. So maybe its better off just in bso.

nwjgit:
> Can you check that its not the case? My worry is just that people will accidentally offer/bury these bones, and theres literally no purpose ever for this in OSB. If someone comes along and accidentally does a /bury trip of all their saved up long bones, they wont be happy. So maybe its better off just in bso.

I tested it and my comment above is correct, but you do bring up a good point. It might be more beneficial to just add to the divine water array manually in bso rather than this approach.

gc:
With no response from the creator, and it seeming like this is only for BSO, im closing this PR. As jonesey said, if this is only for BSO, it should be PR'd in a way to bso.

PR #6117, opened 2024-10-13
Fix ash sanctifier

nwjgit:
redo of https://github.com/oldschoolgg/oldschoolbot/pull/6061 since it affects osb & bso

### Description:
- Fix ash sanctifier not giving proper xp and removing ashes
### Changes:
- Added a check for multiplier by checking BOT_TYPE
- pass updateBank to ashSanctifierEffect() to handle removing ashes
- Fix monsterActivity & zalcano ashSanctifierEffect() usage
- update zalcano return message to incorporate ashSantifierEffect()
- update zalcano to use updateBank to handle giving loot
### Other checks:


gc (changes requested):
I believe eppisis's comment is correct here. You took the loot removing out of this function (not sure why), which I believe now will cause zalcano to give xp without taking away the ashes.

gc:
`ashSanctifierEffect` should instead take an updateBank, and add/change it, instead of making its own one.

nwjgit:
> `ashSanctifierEffect` should instead take an updateBank, and add/change it, instead of making its own one.

The issue is that `updateBank.merge(ashSanctifierResult.updateBank);` isn't sufficient at removing the ashes. `ashSanctifierResult.updateBank` has them removed but when you merge that with the original updateBank the ashes return.

you could do something like `updateBank = ashSanctifierResult.updateBank;` and make updateBank a let instead of const. Not sure if thats something you want to do though.

gc:
Seems like the code hasnt been linted. I always run `yarn dev` before committing, which I recommend.

nwjgit:
> Seems like the code hasnt been linted. I always run `yarn dev` before committing, which I recommend.

Fixed.

PR #6120, opened 2024-10-14
Fix broken Combat achievment IDs

nwjgit:
### Description:
https://github.com/oldschoolgg/oldschoolbot/pull/6095 removed a duplicated CA but some users already completed and have that id stored in completed_ca_task_ids. This PR aims to correct that without manipulating the db.
### Changes:
- Create a new array with all the broken CA id
  - Currentley only one `1135`
- filter these out from `completedTaskIDs` for use in displaying ca information to the user.
### Other checks:

Visual of the current bug: `506/505` this PR will filter out the broken CA to properly display 505/505.
Discord_T6FdJvFO3x


gc:
I've just now removed this task from everyone in the database, so this should no longer be needed.

PR #6121, opened 2024-10-14
Cleaning Scroll

Felris:
### Description:
Scroll of Cleansing was added with dungeoneering 3 years ago, and was never given a use. This use isn't too OP, but makes it do something a little useful at least. Scroll of cleansing saves 10% of secondaries for /mix command. Also added a /data to track how much the scroll has saved.

### Changes:

src/mahoji/commands/mix.ts - Add a check for scroll of cleansing, and if found removes 10%(rounded down) of secondaries (using the secondaries filterable). Then updates cleansing_scroll_bank, and appends a short message saying the scroll of cleansing saved ingredients.

### Other checks:
I've gone through the list of secondaries to make sure it's not too crazy anywhere, and I think it should be fine. The big items it saves are:
Abyssal Dragon bones and Ignecarus scales for dragons fury (But 10% saved is very minor)
Crushed superior dragon bones (Used in super anti-fires which are used in heat res pots)
Divine water for enhanced pots (again I don't think the 10% saved will change much)
Heat Res Vials are considered a secondary (But again, don't think it would change much)

- [ x] I have tested all my changes thoroughly.


gc:
Just to comment, I'm unsure if this is the best use of the item, or the best place a perk like this should come from - although there's nothing wrong with your PR. So the PR is on a "thinking about it" status for now.

PR #6127, opened 2024-10-15
Fix alch agility items choice

DayV-git:
### Description:

Currently bot chooses alch item for alch agility based on 3 second alch time, causing incorrect choice when you have more of a less expensive item to alch. Added agility flag to choose 13s alch agility timer instead.

### Changes:

Add agility parameter to `favAlchs` and `alchPrice`, causing them to take agility alch timer `timePerAlchAgility` instead.

### Other checks:


src/mahoji/lib/abstracted_commands/alchCommand.ts
```diff
@@ -27,6 +27,7 @@ const unlimitedFireRuneProviders = resolveItems([
 
 // 5 tick action
 export const timePerAlch = Time.Second * 3;
+export const timePerAlchAgility = Time.Second * (3 + 10);
```
gc, line 30: Can you make it so the code that calculates the time also uses this same variable so theres not 2 places with this same value?

PR #6136, opened 2024-10-16
Add missing Graceful creates / reverts

DayV-git:
### Description:

Add Brimhaven and Dark graceful revert and individual item reverts for Brimhaven, Dark and Varlamore - same as every other graceful recolour. 

### Changes:

- Add Brimhaven graceful revert.
- Add reverts for each Brimhaven / Varlamore piece.
- Convert every graceful create to `inputItems: Bank()` syntax

### Other checks:


src/lib/data/creatables/gracefulOutfits.ts
```diff
@@ -1226,42 +1226,183 @@ export const gracefulOutfitCreatables: Createable[] = [
...
-		})
+		inputItems: {
+			[itemID('Graceful crafting kit')]: 1,
+			[itemID('Graceful hood')]: 1,
+			[itemID('Graceful top')]: 1,
+			[itemID('Graceful legs')]: 1,
+			[itemID('Graceful gloves')]: 1,
+			[itemID('Graceful boots')]: 1,
+			[itemID('Graceful cape')]: 1
+		},
```
gc, line 1237: Why change these?

DayV-git: @nwjgit asked for these changes to make the code style consistent with other graceful commands.
 https://discord.com/channels/342983479501389826/648196527294251020/1296200249932779532

gc: This format of `{[itemID()]: 1}` is something we don't want to use anymore (although it still may exist in current code)

DayV-git: All commands have been altered from `itemID` to `Bank`. Was easiest to just find and replace on the whole file. Also noted that dark graceful was missing, so added it.

PR #6150, opened 2024-10-20
Add corruptedYoungllef creatable

Felris:
### Description:

Adds Corrupted Youngllef as a creatable

### Changes:

Added corrupted youngllef as creatable/revertable.
Create checks for 1 corrupted gauntlet KC.

### Other checks:


nwjgit:
You should put these under their respective sections: `metamorphPets` and `revertMetamorphPets`

Felris:
> You should put these under their respective sections: `metamorphPets` and `revertMetamorphPets`
> 
> also minor edit: you could end the fail message with a period and remove the else statement and just leave the if and return null at the very end, similar to other customreqs.
> 
> like:
> 
> ```
> 			const [kcName, kcAmount] = await user.getKCByName('Corrupted gauntlet');
> 			if (kcAmount < 1) {
> 				return `You need to have at least 1 ${kcName} KC.`;
> 			}
> 			return null;
> 		}
> ```

Yeah I had done that originally, but I personally prefer explicit else's so I did it my own way. But it's not my code, so you're right, and I updated it to match.

PR #6152, opened 2024-10-25
Fix Bingo Notifications

TastyPumPum:
Currently notifications for the osb bingo are only going to the loot games server which a lot of people aren't in. This adds the notifications to the bingo-notifications channel.

I appreciate this is probably the worst way to do this, but just thought id put the pr up.


gc:
I've fixed this in the db

PR #6157, opened 2024-10-28
Add Varlamore Wealthy Citizen Thieving

DayV-git:
### Description:

Added wealthy citizens to `/steal`. Requires corresponding [OSJS PR](https://github.com/oldschoolgg/oldschooljs/pull/402) . They are very afk (one click per 90s) compared to other thieving, so I have doubled trip length like for redwoods. The time to thieve is an average amount, based on a 100s cycle of 20s no-fail pickpocketing into ~80s downtime to match in game rates (55k per hour).

### Changes:

- Added Wealthy Citizens to `/steal` with double trip time as it is afk.
- Added named quest requirement structure to `stealables`.
- Corrected an issue where thieving would tell you you got a clue scroll, even if you didn't get a new one as you had one in bank. 

### Other checks:


gc:
Waiting for changes to the osjs pr

gc:
Not linted

PR #6158, opened 2024-10-29
crabs

DarkWorldsArtist:
### Description:

crabs

### Changes:

- Added crabs to killables
- Added craftable crab claw and helm
- Added fresh crab items to cl

### Other checks:


src/lib/data/buyables/buyables.ts
```diff
@@ -1106,6 +1106,16 @@ const Buyables: Buyable[] = [
 			thieving: 50
 		}
 	},
+	{
+		name: 'Diving apparatus',
+		gpCost: 0,
```
gc, line 1111: This may let people buy an infinite amount. I would set a cheap GP price, say 1k each.

PR #6160, opened 2024-10-29
Camdozaal Preparing/Offering

DarkWorldsArtist:
Camdozaal fish prayer stuff

### Description:

- Adds camdozaal fish as preparable/cookable (burn rate scaling roughly, data not available on wiki)
- Adds ability to offer prepared camdozaal fish for prayer xp

### Other checks:


gc:
This could go inside `/activities camdozaal action:offer` (and automatically use the fish you have available to offer)

DarkWorldsArtist:
> This could go inside `/activities camdozaal action:offer` (and automatically use the fish you have available to offer)

I'm not sure if you mean you would like preparing and offering to be combined into a single action or not? Don't think I am qualified enough to try tinkering with that. This was the best I could muster

PR #6188, opened 2024-11-14
Add KingGoldemar Sim

nwjgit:
### Description:
- Fix King Goldemar sim not showing dwwh in loot image.

### Changes:
- Add custom logic for `/kill name: King Goldemar`
- Update `calcDwwhChance` to be usable inside `simulatedKillables` so that variables only ever need updated once.
- Move `simulatedKillable` before `osjsMonster` inside kill.worker.ts to always prioritize custom logic first.
- Update the autofill logic for `/kill` to prevent monsters like King Goldemar from showing twice.
- Add a `message` field to `SimulatedKillable` interface. *See example image below

### Other checks:
- I didn't notice the simulate() at the bottom of Boss.ts until after making this PR. I'm not really sure where this is used at besides the unit test. Let me know if changing calcDwwhChance() is a no go, or if you want it done another way.
Discord_rAJSXqc6bE

### DIFF
--- a/src/lib/simulation/simulatedKillables.ts
+++ b/src/lib/simulation/simulatedKillables.ts
@@ -5,9 +5,11 @@ import { randArrItem, randInt, roll } from 'e';
 import { DOANonUniqueTable } from '../bso/doa/doaLootTable';
 import { nexUniqueDrops } from '../data/CollectionsExport';
 import { chanceOfDOAUnique, pickUniqueToGiveUser } from '../depthsOfAtlantis';
+import { KingGoldemarLootTable } from '../minions/data/killableMonsters/custom/bosses/KingGoldemar';
 import { MoktangLootTable } from '../minions/data/killableMonsters/custom/bosses/Moktang';
 import { NEX_UNIQUE_DROPRATE, nexLootTable } from '../nex';
 import { zygomiteFarmingSource } from '../skilling/skills/farming/zygomites';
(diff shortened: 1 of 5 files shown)


src/lib/simulation/simulatedKillables.ts
```diff
@@ -120,6 +123,21 @@ export const simulatedKillables: SimulatedKillable[] = [
...
+		name: 'King Goldemar',
+		isCustom: true,
+		message: '**Assumptions**:\n- Solo\n- Ring of Luck equipped',
+		loot: (quantity: number): Bank => {
+			const loot = new Bank();
+			for (let i = 0; i < quantity; i++) {
+				if (roll(calcDwwhChance(1, true))) {
+					loot.add('Broken dwarven warhammer');
+				}
+				loot.add(KingGoldemarLootTable.roll());
```
gc, line 136: This should be `roll(quantity)` outside the loop, its a lot more efficient/faster as it doesnt need to create an object for every single iteration. If you must do it in a loop (not in this case here), the other way is: `.roll(1, {targetBank: loot})`

gc (changes requested):

src/lib/simulation/simulatedKillables.ts
nwjgit: It doesn't necessarily need to be inside the loop, I was just going off how the others are structured. Let me know if this change looks good and or you want me change the others aswell.

PR #6191, opened 2024-11-17
update drygore prices

nwjgit:
### Description:
Make all drygore 200m value.
### Changes:
- Update all drygore to be 200m value (offhands were 100m but their dyed equivelant are 200m. main hand are also 200m base, and 200m dyed)
- Update drygore sets to be 400m value
### Other checks:


gc:
Can you explain the reasoning?

PR #6198, opened 2024-11-18
add book of the dead buyable

### LINKED ISSUE #6196: Add a way to get "book of the dead"
Now that the book of the dead boosts clues, it would be handy to have a way to obtain it.

DayV-git:
### Description:

add book of the dead buyable. Requires kingdom divided, set to 120 qp as roughly same difficulty as lunar diplomacy. Cost to buy from perdu is 9500 so that's ironman price, and regular price is 1m.

Closes #6196

### Changes:

- add book of the dead buyable

### Other checks:


TastyPumPum:
Closes #6196

PR #6204, opened 2024-11-21
Add nightmare wiki

DayV-git:
### Description:

Readded nightmare wiki pages. Largely copied over from the old versions, with some new formatting. I don't know what's going on with the gear images so I uploaded BIS phasani gear pic. In the old wiki, phosani was a sub-page of nightmare, but I don't know how to do that in the new style, and having a folder looked bad.

### Changes:

- Add Nightmare wiki page
- Add Phosani wiki page with manual BIS gear image

### Other checks:


gc:
I would rather have these on the same page.

PR #6213, opened 2024-11-24
Prevent repeating kill trips when slayer task is finished

### LINKED ISSUE #5972: Prevent repeat trip button working when slayer task is finished
## 🚀 Feature Proposal

When a monster is killed on slayer task, add a hidden `is_slayer_task` flag to the underlying trip, which can be tested when the trip is repeated. If the slayer task has ended, don't send the trip / send 'are you sure' message - and for bonus points, maybe roll a new task? though i guess that conflicts with patreon button. Alternatively, just don't have the repeat trip button at all in this scenario.

DayV-git:
### Description:

Add underlying `onTask` boolean to check if repeated monster killing trip is off task, but original trip was on task. If so, don't do the trip. Can then either get new task, or just use `/k` again to kill the same monster if you really want to. There were other options, like a confirmation message, but probably unnecessary.
https://github.com/oldschoolgg/oldschoolbot/pull/6185

Closes #5972.

### Changes:

- Add `onTask` boolean to `MonsterActivityTaskOptions` and `minionKCommand`, the latter only accessed via repeat trip.
- Return fail message when current slayer task has finished but trying to repeat kill the same monster.

### Other checks:


gc:
Did people agree/like this idea?

PR #6215, opened 2024-11-24
Add Varlamore Wealthy Citizen Thieving

DayV-git:
### Description:

Added wealthy citizens to `/steal`. ~~Requires corresponding [OSJS PR](https://github.com/oldschoolgg/oldschooljs/pull/402)~~ . They are very afk (one click per 90s) compared to other thieving, so I have doubled trip length like for redwoods. The time to thieve is an average amount, based on a 100s cycle of 20s no-fail pickpocketing into ~80s downtime to match in game rates (55k per hour).
https://github.com/oldschoolgg/oldschoolbot/pull/6157

### Changes:

- Added Wealthy Citizens to `/steal` with double trip time as it is afk.
- Added named quest requirement structure to `stealables`.
- Corrected an issue where thieving would tell you you got a clue scroll, even if you didn't get a new one as you had one in bank. 

### Other checks:


src/mahoji/commands/steal.ts
```diff
@@ -75,6 +75,16 @@ export const stealCommand: OSBMahojiCommand = {
...
 
+		if (stealable.questID && !user.user.finished_quest_ids.includes(stealable.questID)) {
+			return `You need to complete the "${stealable.questRequired}" quest to ${
+				stealable.type === 'pickpockable' ? 'pickpocket' : 'steal from'
+			} a ${stealable.name}. Send your minion to do the quest using: ${mentionCommand(
+				globalClient,
+				'activities',
+				'quest'
+			)}.`;
+		}
```
gc, line 86: As above, just get the name of the quest at runtime using something like this:

```ts
	if (monster.requiredQuests) {
		const incompleteQuest = monster.requiredQuests.find(quest => !user.user.finished_quest_ids.includes(quest));
		if (incompleteQuest) {
			return [
				false,
				`You need to have completed the ${bold(
					quests.find(i => i.id === incompleteQuest)!.name
				)} quest to kill ${monster.name}.`
			];
		}
	}
```

src/mahoji/commands/steal.ts
```diff
@@ -100,7 +110,7 @@ export const stealCommand: OSBMahojiCommand = {
 			return 'This NPC/Stall is missing variable respawnTime.';
 		}
 
-		const maxTripLength = calcMaxTripLength(user, 'Pickpocket');
+		const maxTripLength = (stealable.name === 'Wealthy Citizen' ? 2 : 1) * calcMaxTripLength(user, 'Pickpocket');
```
gc, line 113: a 2x higher max trip length, are they really _that_ afk?

gc (changes requested):

src/mahoji/commands/steal.ts
DayV-git: Given that regular pickpocketing is very click intensive compared to regular woodcutting, I figured the comparison was valid. In game, this is considered a 'second monitor' activity, like redwoods or amethysts.

PR #6257, opened 2024-12-08
Scar essence mine

Arodab:
### Description:

Adds everything relevant to the scar essence mine.

### Changes:

-Added `Tainted Essence Chunk` to `/mine` (requires DT2)
-Chunks can be used with coins to `/create` Warped/Twisted/Mangled/Scarred extracts (note: in OSRS, 1 chunk is converted into 5 tainted essence which is stored within the mine. Since this essence is not an item and more like a minigame currency, I just made the minion mine 5x chunks instead for simplicity).
-Added an option to use extracts to `/runecraft` for 250x or 60x runes per essence depending on the rune.
-Standardized the message for bonus runes from different sources.

### Other checks:


src/mahoji/commands/runecraft.ts
```diff
@@ -127,6 +139,27 @@ export const runecraftCommand: OSBMahojiCommand = {
...
+		} else if (twistedRunes.has(rune)) {
+			extractsOwned = bank.amount('Twisted extract');
+			extractType = 'Twisted extract';
+		} else if (mangledRunes.has(rune)) {
+			extractsOwned = bank.amount('Mangled extract');
+			extractType = 'Mangled extract';
+		} else if (scarredRunes.has(rune)) {
+			extractsOwned = bank.amount('Scarred extract');
+			extractType = 'Scarred extract';
+		}
```
gc, line 161: This should probably be something like (defined outside this function):

```ts
const types = [
{
item: getOSItem("Warped extract"),
runes: new Set(),
}
];
```

so you dont need the if elses's

src/mahoji/commands/runecraft.ts
```diff
@@ -296,6 +336,11 @@ export const runecraftCommand: OSBMahojiCommand = {
 		}
 		if (!user.owns(totalCost)) return `You don't own: ${totalCost}.`;
 
+		if (extracts) {
+			totalCost.add(`${extractType}`, quantity);
```
gc, line 340: and here youd be able to do `totalCost.add(item)`

gc (changes requested):
Nice work.

gc:
There was a git issue, merged here instead: https://github.com/oldschoolgg/oldschoolbot/pull/6282

PR #6277, opened 2024-12-31
Make DT2 Rings Tradeable

### LINKED ISSUE #5950: DT2 item incorrect data

DayV-git:
### Description:

Here we are again.

Closes #5950. 

### Changes:

- Take the actual IDs for the rings, generate the item data, then replace ID with the original ID that's used on the bot
- Put in safeguard so script cant reuse the beta IDs 
- The prices will be updated as they are already tracked in item_data, and they will have correct stats
- Repeat logic for a few more items - adds correct price for pharaoh's sceptre
- On bot side, add in a manual override so the original IDs are used for sprites, so they are updated

### Other checks:


DayV-git:
Unit test is failing on github's end but not mine, suspect github isn't actually rebuilding the monorepo before testing

PR #6290, opened 2025-01-17
Add result_quantity to open_until

TastyPumPum:

### DIFF
--- a/src/mahoji/commands/open.ts
+++ b/src/mahoji/commands/open.ts
@@ -57,13 +57,21 @@ export const openCommand: OSBMahojiCommand = {
 					value: i.name
 				}));
 			}
+		},
+		{
+			type: ApplicationCommandOptionType.Integer,
+			name: 'result_quantity',
+			description: 'The number of the target item you want to obtain before stopping.',
+			required: false,
+			min_value: 1,
+			max_value: 1000
 		}
 	],
 	run: async ({
 		userID,
 		options,
 		interaction
-	}: CommandRunOptions<{ name?: string; quantity?: number; open_until?: string }>) => {
+	}: CommandRunOptions<{ name?: string; quantity?: number; open_until?: string; result_quantity?: number }>) => {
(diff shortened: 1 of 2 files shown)


TastyPumPum:
@ellipsis-dev summarise this pr

TastyPumPum:
@ellipsis-dev review this pr

src/lib/rawSql.ts
```diff
@@ -5,15 +5,16 @@ const u = Prisma.UserScalarFieldEnum;
...
 	updateCLArray: (userID: string) => `UPDATE users
-SET ${u.cl_array} = (
-    SELECT (ARRAY(SELECT jsonb_object_keys("${u.collectionLogBank}")::int))
-)
-WHERE ${u.id} = '${userID}';`
+    SET ${u.cl_array} = (
+        SELECT ARRAY(SELECT jsonb_object_keys("${u.collectionLogBank}")::int)
+    )
+    WHERE ${u.id} = '${userID}';`
 };
```
gc, line 19: Is this an intentional change?

TastyPumPum: Yeah this was intentional. Some of the openings were not showing to the user, and I was getting this error message logged:
ERROR: operator does not exist: integer[] & integer[]

Removing the extra pair of parentheses and running linting helped resolve the issue and fixed the errors. If you think it's not needed though you can remove these changes.

gc: That error sounds like it may be because you dont have the intarray extension installed on your postgres. Try running `CREATE EXTENSION intarray;` in your db. I dont see any reason removing parens would fix that error though too, so i would try undo the changes and see if it works without it

DayV-git: The issue is when collection log is getting a lot of updates, here because of opening 5000 caskets at once. I guess this might be splitting up the sql lookup, so concatting is required? Bit annoying to require everyone to install sql extensions to setup bot

PR #6182, opened 2024-11-08
BSO - Colosseum Combat Achievement Fix

TastyPumPum:
Add a fix for the combat achievements One-Off and Denied which is currently only working if you die at wave 11 and 7 respectively or later, rather than colo completions too.

- [x] I have tested all my changes thoroughly.

----

> [!IMPORTANT]
> Fixes `One-Off` combat achievement logic in `elite.ts` and `master.ts` to award completion regardless of wave death.
> 
>   - **Behavior**:
>     - Fixes `One-Off` combat achievement in `elite.ts` and `master.ts` to award completion if player dies before or after wave 11.
>   - **Logic**:
>     - Updates `hasChance` function to check if `diedAt[index]` is undefined or greater than 11 in `master.ts` and greater than 7 in `elite.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=oldschoolgg%2Foldschoolbot&utm_source=github&utm_medium=referral)<sup> for 1e2ab175f86e39423a033cb1a04ee41e4f0e2a69. It will automatically update as commits are pushed.</sup>

PR #6317, opened 2025-02-07
BSO - Colosseum Furball CA Fix

### LINKED ISSUE #6316: Furball CA Appears to be Broken
## 🐛 Bug Report

The combat achievement Furball in the Fortis Colosseum appears to be not functioning properly. The achievement is as follows from the OSRS wiki "Complete Wave 4 without taking avoidable damage from a Manticore." I have done 200 Sol KC and have yet to receive the achievement looking at the code it appears to be a 1/12 chance.

Image

This would equate to a 1/36117042.31 chance of not completing the CA not counting any attempts where I did not get a completion but made it past wave 4 which would further make it rarer. Looking through the official server I could only find one instance of where someone got the achievement on a successful run back in July 2024, everyone else got it while dying at some point. I think the code is not functioning properly and if a run is succesfull its not able to complete. This makes it so that as your minion gets better at Colosseum and death chance drops the odds of "completing" the CA drop drastically. 

Image

TastyPumPum:
Similar fix to #6182 

Fixes the CA Furball on BSO, fixes #6316 

### DIFF
--- a/src/lib/combat_achievements/elite.ts
+++ b/src/lib/combat_achievements/elite.ts
@@ -1534,7 +1534,8 @@ export const eliteCombatAchievements: CombatAchievement[] = [
 		rng: {
 			chancePerKill: 12,
 			hasChance: (data, _user, index) =>
-				data.type === 'Colosseum' && (!data.diedAt || (Array.isArray(data.diedAt) && data.diedAt[index]! > 4))
+				data.type === 'Colosseum' && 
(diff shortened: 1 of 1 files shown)


gc:
Appears not linted, if this fix is still needed can be merged when its green

TastyPumPum:
> Appears not linted, if this fix is still needed can be merged when its green

Linted and ready for merge now

PR #6322, opened 2025-02-13
Fix prepare scripts

### LINKED ISSUE #6252: OSJS wiki monster lookup missing info
## 🐛 Bug Report

Expected vs current monster info for warped terrorbirds. None of the base combats stats / combat level have moved over. Also the wiki link is broken due to the space

image

DayV-git:
### Description:

Fix prep scripts so they can be run as normal. For prep item, just use the old item DB until someone updates it. For monsters, go back to last functioning monster DB, and parse wiki for what's left, with some bug fixes. I couldn't get the special export to work, so searching each monster one by one. Should ensure monsters are actually created and killable when their loot table is made (eg chilled jellies). Closes #6252. 

### Changes:

- Add osjs generate to root `pnpm monorepo:build`. Added `pnpm monorepo:build` to `pnpm dev`.
- Fix Item DB commit for prepareItems
- Add back prepareMonster scripts, using the last monster DB commit before monster DB messed up
- Fix some wiki lookups for remaining monsters (always search lower case in wiki, huey doesn't have `examine` as infobox is split in 3, but it has `examine1`)
- Ignore monsters without a regular table (and remove table from master farmer)
- Ensure output monsters_data.json is correctly linted

### Other checks:


packages/oldschooljs/package.json
```diff
@@ -25,7 +25,7 @@
 	"scripts": {
 		"test": "concurrently \"pnpm test:unit\" \"tsc --noEmit -p test\" \"tsc --noEmit -p src\"",
 		"test:unit": "vitest run --coverage",
-		"generate": "tsx scripts/prepare",
+		"prepare": "tsx scripts/prepare",
```
gc, line 28: This was renamed because "prepare" is a special script name that runs at certain times, like anytime you run `pnpm install`, it will automatically run the "prepare" script if you have one. I'm not 100% sure if it's the same problem with pnpm vs yarn, or when it runs, but its better we don't rename it back.

DayV-git: Not sure what the problem is. Pnpm looks for prepare script and runs it on install, allowing an easy way to run these scripts from root. Nevertheless, i've changed it back to generate, and updated monorepo:build and pnpm dev so it can run

gc: Well there's not a connection between the installation and the need to run the prepare script. We only need to run it when adding new monsters/items. It caused me trouble/problems having it here before which is why I renamed it.

Issue #5583, opened 2024-01-17
1t karambwan cooking on bso


DarkBugz:
Fishy coded it for osb but it never merged to bso. Cyr said probably a merge error

nwjgit:
I don't believe this has been added to BSO. Did magna ever comment on this?

PR #6344, opened 2025-03-18
bso 1t karambwan cooking

nwjgit:
### Description:
merge the osb 1 tick cooking karambwan to bso.

### Changes:
- re-work some logic in cook.ts to use `cookingBoostItems` and cookingBoost for all cookables including wines and karambwan.
- add a check to 1t cook karambwan if the user has >92 cooking.
  - I also added the remy boost similar to wines and this gives ~9.8m xp/h *see ss below*
  - xp/hr is similar to rocktails in this form so something may need adjusted if this is an issue. 
- update wiki for remy's boost

### Other checks:
- closes: https://github.com/oldschoolgg/oldschoolbot/issues/5583

### XP rates:
wines:
Discord_8I3cvPuGnm
karambwan:
Discord_z4tOB8Zj6B
rocktails:
Discord_EvPliPpPoh


gc:
I fixed it for you, but for future reference, you can only change the wiki from osb(master).

PR #6358, opened 2025-04-19
Fix 1/128 chance of integration tests failing due to Skotizo dropping dark totem

DayV-git:
### Description:

Currently integration test for Skotizo checks for 1 totem less in bank after trip. But wont be true if Skotizo drops it when killed

### Changes:

- Add `tripStartBank` to integration test kill result that contains bank after sending command, but before trip finishes
- Check this bank for Skotizo totem check

### Other checks:


gc:
👍

PR #6377, opened 2025-05-10
Change 3rd age jibwings => Third age jibwings so they don't /sell

themrrobert:
### Description:

Change the name of the 3rd age jibwings so they don't bot-sell as easily when people sell their 3rd age, like with Third age dye.


gc:
Just needs snapshot update: pnpm test:unit -u

PR #6395, opened 2025-05-21
Update BSO CoX wiki page: boosts, supplies, and item charges

chrisjeng:
This PR updates the Chambers of Xeric wiki page to reflect the current BSO implementation:

- Updated item boost list (Void staff to 9%, added Tumeken's shadow)
- Added Bandos godsword (or) and offhand spidergore rapier
- Clarified (equipped) and (charged) item rules
- Added full table of degradable item charges
- Rewrote supply section with KC-based table and solo note


gc:
Thank you

PR #6400, opened 2025-05-27
Add pages + ironman filter to lb minigames

TastyPumPum:
Added pagination and ironman filter to minigames leaderboard in src/mahoji/commands/leaderboard.ts.


gc:
Im not understanding why it was changed to a raw query?

PR #6415, opened 2025-06-02
Ensure infinity robes are equipable using /gear item

DayV-git:
### Description:

Currently parser catches 'infinity' as a number when trying to equip infinity robes by name and bugs out. 
PR fixes this by adding in an exception for this case, so it always assumes 'infinity' is part of the item name.

### Changes:

- Exclude 'infinity' from the quantity check in `parseQuantityAndItem`

### Other checks:


gc:
Perfect fix

PR #6436, opened 2025-06-13
Add buying trips (ores/runes/arrow shafts)

### LINKED ISSUE #3369: Suggestion: Blood Rune shopscape
Now with ToB being released we have Scythe of Vitur and Sanguinesti staff in the bot, which cost large amount of Blood Runes to use. These costs are very prohibitive for Ironmen who can't just buy runes from other players, and the general consensus seems to be no one is going to use the weapons.

In OSRS, Blood Rune consumption is also a real problem for most endgame Ironmen, and the normal way to acquire more after the initial lump you get from 99 runecrafting is to just hop worlds and buy them from rune shops. This method can be 10-20x faster than runecrafting yourself, assuming you have the gp to do so.

Hopping worlds and buying, i.e. "shopscape" trips, should be added to OSB too. They shouldn't be instant of course, but more like the +collect we already have, just consuming copious amounts of gp. For example half an hour trip that consumes 15 million gp and results into 30k blood runes.

TastyPumPum:
- Support buying trips via new quantityPerHour field
- implement generic buyingTripCommand
- Add bf ores, arrow shafts and some runes as the first trip buyables

Closes #3369 


src/lib/data/buyables/tripBuyables.ts
```diff
@@ -0,0 +1,162 @@
...
+		name: 'Pineapple',
+		outputItems: new Bank().add('Pineapple'),
+		gpCost: 5,
+		quantityPerHour: 3500,
+		shopQuantity: 15,
+		changePer: 2
+	},
+	{
+		name: 'Banana',
+		outputItems: new Bank().add('Pineapple'),
```
gc: banana

src/lib/util/minionStatus.ts
```diff
@@ -81,6 +82,7 @@ import type {
 	WoodcuttingActivityTaskOptions,
 	ZalcanoActivityTaskOptions
 } from '../types/minions';
+import { itemNameFromID } from '../util';
```
gc: Use `Items.itemNameFromId`

src/tasks/minions/buyActivity.ts
```diff
@@ -0,0 +1,53 @@
...
+	async run(data: BuyActivityTaskOptions) {
+		const { userID, channelID, itemID, quantity } = data;
+		const user = await mUserFetch(userID);
+
+		const buyable = Buyables.find(b => {
+			if (!b.quantityPerHour || !b.outputItems) return false;
+
+			const bank = typeof b.outputItems === 'function' ? b.outputItems(user) : b.outputItems;
+
+			return bank.has(itemID);
```
gc: This is a little sketchy but seems okay, maybe also a bank.length === 1 just to add some extra safety

src/tasks/minions/buyActivity.ts
```diff
@@ -0,0 +1,53 @@
...
+
+		if (!buyable) {
+			throw new Error(`No buyable found for item ${itemID}`);
+		}
+
+		const { totalCost } = data;
+		const average = Math.floor(totalCost / quantity);
+
+		const itemsPerHour = (quantity / (data.duration / Time.Minute)) * 60;
+		const itemName = itemNameFromID(itemID);
```
gc: Use `Items.itemNameFromId`

src/tasks/minions/buyActivity.ts
```diff
@@ -0,0 +1,53 @@
...
+		});
+
+		if (!buyable) {
+			throw new Error(`No buyable found for item ${itemID}`);
+		}
+
+		const { totalCost } = data;
+		const average = Math.floor(totalCost / quantity);
+
+		const itemsPerHour = (quantity / (data.duration / Time.Minute)) * 60;
```
gc: Optional/no change needed, but `calcPerHour` function exists to make this a little easier

```ts
import { calcPerHour } from '@oldschoolgg/toolkit/util';
const itemsPerHour  = calcPerHour(quantity, data.duration);
```

Its up to you in this case

src/lib/util/calculateShopBuyCost.ts
```diff
@@ -0,0 +1,25 @@
+export function calculateShopBuyCost(
+	gpCost: number,
+	quantity: number,
+	shopQuantity?: number,
+	changePer?: number
+): { total: number; average: number } {
```
gc: This should be changed to an object for the parameters, so that its less prone to accidentally passing them in the wrong order.

```ts
export function calculateShopBuyCost({
	gpCost,
	quantity,
	shopQuantity,
	changePer
}: { gpCost: number; quantity: number; shopQuantity?: number; changePer?: number }): {
	total: number;
	average: number;
} {```

File: tests/unit/calculateShopBuyCost.test.ts [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+import calculateShopBuyCost from '../../src/lib/util/calculateShopBuyCost';
+
+describe('calculateShopBuyCost', () => {
+	test('fixed price cost', () => {
+		const result = calculateShopBuyCost(50, 3);
+		expect(result).toEqual({ total: 150, average: 50 });
+	});
+
+	test('variable cost within one world', () => {
+		const result = calculateShopBuyCost(100, 3, 1000, 10);
```
gc: Here is a good example of the previously recommended object params change for `calculateShopBuyCost`, looking at this, I have no idea what these numbers are.

src/mahoji/lib/abstracted_commands/buyingTripCommand.ts
```diff
@@ -0,0 +1,78 @@
...
+	interaction: ChatInputCommandInteraction
+) {
+	let quantityPerHour = buyable.quantityPerHour!;
+	if (buyable.name === 'Coal') {
+		if (user.owns('Coal bag')) {
+			quantityPerHour *= 1.6;
+		}
+	}
+	const timePerItem = Time.Hour / quantityPerHour;
+	const osItem = getOSItem(buyable.name);
```
gc: Optional, but recommended for future: use `Items.getOrThrow` instead of `getOSItem`. If we didn't expect it to ALWAYS return a item, it would be `Items.getItem`, in this case it would be a bug if nothing was returned, so we use getOrThrow.

src/lib/data/buyables/tripBuyables.ts
```diff
@@ -0,0 +1,162 @@
...
+		name: 'Chaos rune (pack)',
+		outputItems: new Bank().add('Chaos rune', 100),
+		gpCost: 9950,
+		quantityPerHour: 15_000,
+		shopQuantity: 35,
+		changePer: 0.1
+	}
+];
+
+export default tripBuyables;
```
gc: dont export default here

src/lib/data/buyables/tripBuyables.ts
```diff
@@ -0,0 +1,162 @@
+import { Bank } from 'oldschooljs';
+
+import type { Buyable } from './buyables';
+
+export const tripBuyables: Buyable[] = [
```
gc: If these tripBuyables are entirely separate from normal buyables, which it seems like they are, they could instead not use the `Buyable` interface, or be

```ts
export const tripBuyables: ({item: EItem} & Omit<Buyable, 'name' | 'outputItems'>)[] = [
{
		item: EItem.PINEAPPLE,
		gpCost: 5,
		quantityPerHour: 3500,
		shopQuantity: 15,
		changePer: 2
	},
```

Just an idea, not required

src/lib/util/calculateShopBuyCost.ts
```diff
@@ -0,0 +1,36 @@
...
+	const remainder = quantity % shopQuantity;
+
+	const sumFullWorld = prices.reduce((acc, p) => acc + p, 0);
+	const sumRemainder = prices.slice(0, remainder).reduce((acc, p) => acc + p, 0);
+
+	const total = sumFullWorld * worldCount + sumRemainder;
+	return { total, average: total / quantity };
+}
+
+export default calculateShopBuyCost;
```
gc: This is already exported above, and there's basically never a time I ever want a `export default`  in the code, as theyre more unclear and not as explicit, less good IDE support, etc.

src/lib/util/calculateShopBuyCost.ts
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	const worldCount = Math.floor(quantity / shopQuantity);
+	const remainder = quantity % shopQuantity;
+
+	const sumFullWorld = prices.reduce((acc, p) => acc + p, 0);
+	const sumRemainder = prices.slice(0, remainder).reduce((acc, p) => acc + p, 0);
+
+	const total = sumFullWorld * worldCount + sumRemainder;
+	return { total, average: total / quantity };
```
gc: It maybe isn't possible (not sure?), but id be careful of this returning a non-integer price. Assuming total is the total price? (naming can help here)

tests/unit/priceAbuses.test.ts
```diff
@@ -69,7 +69,9 @@ describe('Price Abusing', () => {
 			i.gpCost !== undefined &&
 			i.itemCost === undefined &&
 			!isFunction(i.outputItems) &&
-			(!i.outputItems || i.outputItems.length === 1)
+			(!i.outputItems || i.outputItems.length === 1) &&
+			!i.shopQuantity &&
+			!i.quantityPerHour
```
gc, line 74: It looks like we're ignoring these items from the price-abuses test, is that because its just tricky/annoying to update the test or these items just definitely are impossible to price-abuse? Either scenario is fine, you don't have to update the test if you don't want to, although I'd just want to know.

tests/unit/calculateShopBuyCost.test.ts
```diff
@@ -0,0 +1,37 @@
...
+		const result = calculateShopBuyCost({ gpCost: 100, quantity: 3, shopQuantity: 1000, changePer: 10 });
+		expect(result).toEqual({ total: 330, average: 110 });
+	});
+
+	test('variable cost with world hops', () => {
+		const result = calculateShopBuyCost({ gpCost: 100, quantity: 3, shopQuantity: 2, changePer: 10 });
+		expect(result.total).toBe(310);
+		expect(result.average).toBeCloseTo(103.33, 2);
+	});
+
```
gc: ZAMMMNNN tests 😋😋😋👨‍❤️‍💋‍👨

gc:
It all looks good, just a couple little comments. Provided it's tested and its accurate to osrs (in terms of cost and items per hour, etc), you can merge this yourself.

PR #6442, opened 2025-06-17
Add scarred extract support for Dark Altar runecrafting

TastyPumPum:
## Summary
- allow dark altar runs to use Scarred extracts
- track extracts usage in command args and minion status
- award extract bonus when crafting at the Dark Altar
- update repeat trip logic for extracts

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -15,7 +15,7 @@
 		"@astrojs/preact": "^3.5.4",
 		"@astrojs/starlight": "^0.28.2",
 		"@tailwindcss/vite": "^4.1.5",
-		"astro": "^4.16.18",
+		"astro": "^4.16.19",
 		"collapse-white-space": "^2.1.0",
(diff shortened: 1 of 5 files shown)


src/lib/util/repeatStoredTrip.ts
```diff
@@ -276,7 +276,10 @@ const tripHandlers = {
 	},
 	[activity_type_enum.DarkAltar]: {
 		commandName: 'runecraft',
-		args: (data: DarkAltarOptions) => ({ rune: `${data.rune} rune (zeah)` })
+		args: (data: DarkAltarOptions) => ({
+			rune: `${darkAltarRunes[data.rune].item.name} (zeah)`,
```
DayV-git: Just put it back to `rune: `${data.rune} rune (zeah)`

TastyPumPum: Fixed, thanks!

PR #6456, opened 2025-06-20
Add plugin to fix link spacing

### LINKED ISSUE #6454: Hyperlinks incorrectly formatted on wiki
Spaces before or after hyperlinks disappear if there is formatted content on that side, see screenshot. 

Image

DayV-git:
### Description:

Adds rehype plugin to ensure whitespace before or after hyperlinks is preserved in wiki. Closes #6454. 

### Changes:

- Adds rehype plugin to ensure whitespace before or after hyperlinks is preserved
- Adds unist-util-visit to requirements

### Other checks:

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,6 @@
 		"simple-statistics": "^7.8.3",
 		"skia-canvas": "^1.0.2",
 		"sonic-boom": "^4.2.0",
-		"unist-util-visit": "^5.0.0",
 		"zod": "^3.24.3"
 	},
 	"devDependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -68,6 +68,7 @@
 		"simple-statistics": "^7.8.3",
 		"skia-canvas": "^1.0.2",
 		"sonic-boom": "^4.2.0",
+		"unist-util-visit": "^5.0.0",
```
gc: This dependency was installed to the root of the repo, (package.json instead of docs/package.json) - its already installed there, so this should be okay to remove, and no changes needed to the pnpm-lock.yaml either, unless unist-util-visit is being updated.

DayV-git: Removed.

PR #6474, opened 2025-07-01
Ensure autoslay boss has valid target and add difficulty rating to amox

DayV-git:
### Description:

Currently autoslay boss fails if no task monster has `difficultyRating`, currently warped creatures and nagua. Fixes this by starting `maxDiff` at -1 so at least the first monster will be valid. Adds `difficultyRating` to amox so it will be target of boss autoslay. 

### Changes:

Initialise `maxDiff` to -1 when choosing autoslay boss target so there is at least 1 valid monster.
Set Amoxliatl `difficultyRating`.
Add quest req for Amox task

### Other checks:


DayV-git:
closed as stale

PR #6484, opened 2025-07-10
Improve trade and pay clarity

TastyPumPum:
A new Bank.toStringFull method outputs item quantities with comma formatting, and both /trade and /pay confirmations now show exact amounts.

Key changes include:
- Introduced toStringFull in the bank class to show comma-separated quantities without shortening
- /pay confirmation uses amount.toLocaleString() so large GP amounts display fully, replacing the prior toKMB format
- /trade confirmation strings use toStringFull() for both sent and received items, ensuring full quantities are displayed to each user, if this exceeds the character limit though, reverts back to the shortened way. 

### DIFF
--- a/packages/oldschooljs/src/structures/Bank.ts
+++ b/packages/oldschooljs/src/structures/Bank.ts
@@ -285,6 +285,17 @@ export default class Bank {
 			.join(', ');
 	}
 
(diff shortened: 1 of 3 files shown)


KoDj2:
Brilliant

PR #6504, opened 2025-08-14
Royal  Titans

Arodab:
### Description:

Adds the Royal Titans to OSB.

### Changes:

- Added three versions of the Royal Titans to `/k`
    - only Branda drops the Fire element staff crown and Mystic vigour prayer scroll
    - only Eldric drops the Ice element staff crown and Deadeye prayer scroll
    - the sacrice option only drops clues and the pet at a rate of  1/1500 instead of 1/3000

- Added Giantsoul amulet, Twinflame staff, Burnt/Soaked/Soiled page to `/create`
- Added the new prayer scrolls to `/use` (boosts for gauntlet)
- Added Giantsoul amulet boost to Obor and Bryophita

### Other checks:

### DIFF
--- a/src/lib/minions/data/killableMonsters/bosses/misc.ts
+++ b/src/lib/minions/data/killableMonsters/bosses/misc.ts
@@ -1437,8 +1437,9 @@ const killableBosses: KillableMonster[] = [
 					feet: 'Dragon boots',
 					ring: 'Berserker ring (i)'
 				}).stats,
-				defence_magic: 0,
-				prayer: 0,
+				attack_ranged: -100,
+				defence_magic: -10,
+				prayer: -100,
 				defence_crush: 0,
 				defence_ranged: 0,
 				defence_slash: 0,
--- a/packages/oldschooljs/metadata.txt
+++ b/packages/oldschooljs/metadata.txt
@@ -4,15 +4,14 @@ Files in ./dist/cjs:
 	./dist/cjs/EItem.cjs: 334.61 KB 342,641 bytes
 	./dist/cjs/EMonster.cjs: 25.67 KB 26,285 bytes
 	./dist/cjs/gear/index.cjs: 1.77 KB 1,810 bytes
-	./dist/cjs/index.cjs: 864.83 KB 885,583 bytes
-	./dist/cjs/item_data-OPRFPN4I.json: 5.25 MB 5,509,046 bytes
-	./dist/cjs/monsters_data-5F625L5X.json: 363.54 KB 372,262 bytes
-	./dist/cjs/monsters_data-ZWSBZZS5.json: 363.54 KB 372,262 bytes
+	./dist/cjs/index.cjs: 868.86 KB 889,709 bytes
+	./dist/cjs/item_data-EJRRI4VO.json: 5.41 MB 5,676,428 bytes
+	./dist/cjs/monsters_data-GXY364KR.json: 368.1 KB 376,937 bytes
 	./dist/cjs/structures/Hiscores.cjs: 40.12 KB 41,084 bytes
 	./dist/cjs/structures/Wiki.cjs: 3.8 KB 3,889 bytes
 	./dist/cjs/util/util.cjs: 38.22 KB 39,137 bytes
 
-Total size: 7.36 MB
+Total size: 7.18 MB
 
 
 Files in ./dist/esm:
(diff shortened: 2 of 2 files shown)


fixgems:
🔥

packages/oldschooljs/src/simulation/monsters/bosses/RoyalTitans.ts
```diff
@@ -0,0 +1,117 @@
...
+	.add('Fire battlestaff', 3, (2 * 12) / 387)
+	.add('Rune plateskirt', 1, (2 * 12) / 387)
+	.add('Rune scimitar', 1, (2 * 12) / 387)
+	.add('Rune pickaxe', 1, (2 * 12) / 387)
+
+	.add('Chaos rune', [100, 200], (2 * 15) / 387)
+	.add('Death rune', [75, 150], (2 * 15) / 387)
+	.add('Nature rune', [50, 75], (2 * 15) / 387)
+	.add('Law rune', [50, 75], (2 * 15) / 387)
+	.add('Soul rune', [50, 75], (2 * 15) / 387)
```
nwjgit: Did you test these drop tables? The normal format is using the numerator in the expanded fraction (a/b) from the wiki, not the entire fraction. The total weight will be dynamically made based off all the items and their weight. You can look at `TheHueycoatl.ts` as an example. As for the 2x i'm not quite sure what this refers to on the wiki, maybe 2 rolls per kill, if thats the case then something else could be done to give the user 2 rolls to make the rates accurate.

Arodab: Yeah it works fine like that, but I will standardize it. And 2x actually is 2 rolls on the drop table, thanks for catching that.

src/lib/minions/data/killableMonsters/bosses/misc.ts
```diff
@@ -1044,6 +1047,404 @@ const killableBosses: KillableMonster[] = [
...
+		attackStyleToUse: GearStat.AttackSlash,
+		attackStylesUsed: [GearStat.AttackMagic],
+		defaultAttackStyles: [SkillsEnum.Attack],
+		healAmountNeeded: 10 * 10,
+		minimumGearRequirements: {
+			melee: {
+				...new Gear({
+					neck: 'Amulet of fury',
+					cape: 'Fire cape',
+					body: 'Bandos chestplate',
```
DayV-git: Bandos seems a bit too high for a base requirement. Wiki has fighter torso and black dhide as minimum gear

packages/oldschooljs/src/data/items/item_data.json
DayV-git, line 1: Should just have the minimum changes to item data for your PR. The items deleted in particular will cause issues. Can see from armoured zombies 2 PR #6382, it had full item data changes and magna reverted most of it.

src/mahoji/lib/abstracted_commands/minionKill/newMinionKill.ts
```diff
@@ -157,8 +159,20 @@ export function newMinionKillCommand(args: MinionKillOptions) {
...
 	const ephemeralPostTripEffects: PostBoostEffect[] = [];
+
+	let kcForBonus = args.monsterKC; // Default to the single monster's KC
+	const royalTitansGroupIDs = [Monsters.Branda.id, Monsters.Eldric.id, Monsters.RoyalTitans.id];
+
+	if (royalTitansGroupIDs.includes(monster.id)) {
+		const brandaKC = await args.user.getKC(Monsters.Branda.id);
+		const eldricKC = await args.user.getKC(Monsters.Eldric.id);
+		const sacrificeKC = await args.user.getKC(Monsters.RoyalTitans.id);
+		kcForBonus = brandaKC + eldricKC + sacrificeKC;
```
DayV-git: Could just add these in the definition of monsterKC in minionKillCommand, so this function doesn't need async added

src/lib/minions/data/killableMonsters/bosses/misc.ts
Arodab: Wiki has worse minimum gear for pretty much every single boss in osb lol, I just copied these from a boss that's a lot easier than royal titans. If anyone wants to go over the rest and let me know what reqs are too high, I can just update all bosses at once.

DayV-git (commented):
Go back through the comments from Ellipsis, looks like the typos are still there

src/lib/minions/data/killableMonsters/bosses/misc.ts
DayV-git: Tbh the entire minimum gear section probably isn't needed. People will wear barrows for food use and that will be fine for base requirements.

PR #6506, opened 2025-08-15
Bso

5Rage:
### Description:
Added Item cost validation and removal for tames kill command

### Changes:

function to check requireditemcost
Check monsters item cost, check bank items.
if required items are found, proceed and remove

### Other checks:

untested as yet


5Rage:
didnt mean to pull yet

PR #6509, opened 2025-08-26
Unique and total IC donation leaderboards

nwjgit:
### Description:
Adds two new leaderboads for bso
- `/lb unique_ic_donation_given`
- `/lb total_ic_donation_given`

### Changes:
- add itemContractDonationGivenLb() to handle both leaderboards

### Other checks:


gc:
Could be laggy/slow but we'll see in the real bot

PR #6517, opened 2025-09-03
Royal titans stat requirements and zcb boost

Arodab:
### Description:

Previous stat reqs fix PR only fixed it for the sacrifice option, this fixes it for Branda and Eldric too, and some other requested changes.

### Changes:

- Added a +70 magic attack req in mage gear.
- Zcb is now bis spec boost over claws.

### Other checks:


DayV-git (commented):
Should refactor this to remove duplicate code, so you don't need to do identical changes 3 times

nwjgit:
Unsure of the impact but monsters.json also needs updated when you make these changes. `pnpm dev` will handle making that change. Its coming up on a different PR i'm doing but will delete since its unrelated to what i'm working on.

PR #6524, opened 2025-09-06
Update RoboChimp more frequently

TastyPumPum:


gc:
It's not a replacement for this PR, but I've added a change that does a robochimp sync when your minion finishes a trip (the only limitation being it only does this once per 30 minutes), so if all your trips over >30 minutes, it will sync every trip: https://github.com/oldschoolgg/oldschoolbot/commit/7865114761adcf63ac204b6de754360284800e6d

PR #6558, opened 2025-10-02
Vale Totems

Jettucis:
### Description:

In short, I've added Vale Totems minigame together with almost everything that comes with it.

### Changes:
- Vale Totems minigame
- Fletching knife and it's perks to fletching
- Greenman mask (may add functionality for it in future)
- Bow string spool
- Vale Research Exchange (buying/selling items from it)
- Collection Log (haven't tested)
- Bale of flax (as openable)
- Dirty arrowtips (automatically cleaned)
- Shadows of Custodia quest
- Custodian stalkers
- Slayer extension for Custodian stalkers
- Atlatl dart fletching

### Other checks:

- [Y] I have tested almost all my changes thoroughly.
Definitely more people should test, but so far on my end seemed fine.
And possibly timing needs to be changed, but I was referencing the time from Vale Totems in OSRS, so it should be pretty accurate (wiki xp rates may differ from what I could achieve).

P.S. Haven't added cause wanted for you to decide, but there could be another perk added for Bow string spool. Allow longer ventures since you can fletch 27 bows instead of 14 per one banking. At first I added it, but then if we are for reall, then it would save only couple seconds, so not a big deal.


src/lib/skilling/skills/fletching/fletchables/atlatl.ts
```diff
@@ -0,0 +1,45 @@
...
+		id: itemID('Atlatl dart shaft'),
+		level: 74,
+		xp: 30,
+		inputItems: new Bank({ 'Ent branch': 1 }),
+		tickRate: 3,
+		outputMultiple: 100,
+		fletchingKnifeBoost: true
+	},
+	{
+		name: 'Headless atlatl dart',
```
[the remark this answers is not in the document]
Jettucis: [That's intended](https://oldschool.runescape.wiki/w/Headless_atlatl_dart)

Jettucis:
Collection log for Vale Totems also tested. All good from my end. Still if someone wants, can test.

docs/src/content/docs/osb/Activities/vale-totems.md
```diff
@@ -0,0 +1,34 @@
+---
+title: "Vale Totems"
+---
+
+Vale Totems is a [[fletching]] minigame where you can receive [[Fletching knife]], [[Bow string spool]], [[Ent branch]] and **Vale research points** to purchase all of those items. With all available time reductions, it is possible to achieve upwards of [[fletching:300k XP]] per hour while fletching [[Magic logs]].
+
+## Requirements
+
+- [[fletching:20]]
+- Children of the Sun quest ([[/activities quest name:Children of the Sun]]).
```
DayV-git: `[[/activities quest name\:Children of the Sun]]`

docs/src/content/docs/osb/Activities/vale-totems.md
DayV-git, line 1: Should be in minigames folder (ignore the other minigames that are also here, they should all be in minigames)

src/lib/minions/data/killableMonsters/chaeldarMonsters.ts
```diff
@@ -117,6 +117,29 @@ export const chaeldarMonsters: KillableMonster[] = [
...
+	{
+		id: Monsters.ElderCustodianStalker.id,
+		name: Monsters.ElderCustodianStalker.name,
+		aliases: Monsters.ElderCustodianStalker.aliases,
+		timeToFinish: Time.Second * 65,
+		table: Monsters.ElderCustodianStalker,
+		wildy: false,
+		existsInCatacombs: false,
+		difficultyRating: 3,
+		notifyDrops: resolveItems(['Antler guard']),
```
DayV-git: Probably not high enough value for notification

src/lib/minions/data/killableMonsters/chaeldarMonsters.ts
```diff
@@ -117,6 +117,29 @@ export const chaeldarMonsters: KillableMonster[] = [
...
+		qpRequired: 77,
```
DayV-git: Quest point requirements are mainly for quest reqs without a specificQuest. Not needed here

src/lib/minions/data/quests.ts
```diff
@@ -353,6 +354,33 @@ export const quests: Quest[] = [
 			}
 			return duration;
 		}
+	},
+	{
+		id: QuestID.ShadowsOfCustodia,
+		qp: 2,
+		name: 'Shadows of Custodia',
+		prerequisitesQuests: [QuestID.ChildrenOfTheSun],
+		qpReq: 75,
```
DayV-git: why qpReq?

src/mahoji/commands/fletch.ts
```diff
@@ -63,11 +64,14 @@ export const fletchCommand: OSBMahojiCommand = {
 		}
 
 		const userBank = user.bank;
+		const hasFletchingKnife = user.hasEquippedOrInBank(itemID('Fletching knife'));
 
 		// Get the base time to fletch the item then add on quarter of a second per item to account for banking/etc.
 		let timeToFletchSingleItem = fletchable.tickRate * Time.Second * 0.6 + Time.Second / 4;
 		if (fletchable.tickRate < 1) {
 			timeToFletchSingleItem = fletchable.tickRate * Time.Second * 0.6;
+		} else if (fletchable.tickRate > 1 && fletchable.fletchingKnifeBoost && hasFletchingKnife) {
```
DayV-git: Can replace this with `!Items.getItem(fletchable.id).stackable` (assuming this is what you are going for), so you don't need to label every fletchable with `fletchingKnifeBoost`

src/mahoji/commands/fletch.ts
```diff
@@ -108,6 +112,8 @@ export const fletchCommand: OSBMahojiCommand = {
 
 		return `${user.minionName} is now Fletching ${quantity}${sets} ${
 			fletchable.name
-		}, it'll take around ${formatDuration(duration)} to finish. Removed ${itemsNeeded} from your bank.`;
+		}, it'll take around ${formatDuration(duration)} to finish. Removed ${itemsNeeded} from your bank.${
+			hasFletchingKnife ? '\n\n**Fletching knife bonus:** -1 tick per item' : ''
```
DayV-git: Only add this if user is getting boost

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+		return `${user.minionName} doesn't own: ${logCostLap}x ${logType.name}.`;
+	}
+
+	if (fletchingLvl < decoration.level) {
+		return `${user.minionName} needs ${decoration.level} Fletching to fletch ${decoration.item.name}.`;
+	}
+
+	let staminaPotItem: Item | undefined;
+	if (staminaPot) {
+		staminaPotItem = Items.getOrThrow(12625);
```
DayV-git: EItem.STAMINA_POTION4

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+		}
+	} else {
+		messages.push(`-${formatDuration(NO_BANK_SHORTCUT_PENALTY)} for Log basket`);
+	}
+
+	if (agilityLevel < LOG_SHORTCUTS_AGILITY_LVL) {
+		// Shortcuts
+		timePerLap += NO_SHORTCUT_PENALTY;
+	} else {
+		messages.push(`-${formatDuration(NO_BANK_SHORTCUT_PENALTY)} for other shortcuts`);
```
DayV-git: 'shortcuts' wont make sense to user, should just say boost due to specific level (and can combine the agility boosts)

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+	if (decoration.stringing) {
+		const stringCost = STRINGS_PER_LAP * laps;
+		cost.add(stringItem!.id, stringCost);
+	}
+	if (staminaPot) {
+		cost.add(staminaPotItem);
+	}
+
+	await user.removeItemsFromBank(cost);
+	//await userStatsBankUpdate(user, 'vale_totems_cost_bank', cost); removed cause no idea if needed
```
DayV-git: Yes, it's used as the cost in `loot view` command

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+		type: 'ValeTotems',
+		offerings: totalOfferings,
+		fletchXp: fletchingXp,
+		userID: user.id,
+		duration: duration,
+		channelID: channelID.toString(),
+		minigameID: 'vale_totems',
+		quantity: laps,
+		logID: logType.name,
+		itemID: fletchableItem.name,
```
DayV-git: log and item variable names are misleading

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+	)}, with each lap taking ${formatDuration(timePerLap)}.`;
+
+	if (messages.length > 0) {
+		str += `\n\n**Per lap:** ${messages.join(', ')}.`;
+	}
+
+	return str;
+}
+
+// xpPerLap (carving + decorating + fletching) * 8
```
DayV-git: Remove unnecessary comments

src/mahoji/commands/fletch.ts
Jettucis: The boost is provided only when fletching something from logs, so if I replaced the statement as you're suggesting I'd be providing boost to Unfinished and finished cbows and slayer fletchables, but I see your point, could just use a check if  inputItems contains "logs" or "bark"

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+	);
+
+	const { vale_research_points: newPoints } = await user.statsUpdate({
+		vale_research_points: {
+			increment: gain
+		}
+	});
+
+	await user.transactItems({
+		itemsToRemove: shopItem.output.multiply(quantity)
```
gc: `.multiply` mutates the bank, causing it to multiply every time someone uses the command, for this situation we usually: (1) Call `.freeze()` on all Bank's that should never ever change (2) do `.clone().multiply(x)` instead.

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+
+	if (itemOwned < quantity) {
+		return `${user.minionName} doesn't have enough ${shopItem.name} to sell.\n${
+			itemOwned < 1
+				? `In fact, ${user.minionName} doesn't have any ${shopItem.name} in its bank.`
+				: `${user.minionName} only has ${itemOwned}x ${shopItem.name} in its bank.`
+		}`;
+	}
+
+	if (!user.isIronman && sellable.name === 'Greenman mask') {
```
gc: It's not a thing you must change, if this string accidentally wasnt a perfect match of the name, then it would fail. so now we generally prefer using `EItem`: `.id === EItem.GREENMAN_MASK`,

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+// Per lap basis
+export const ValeTotemsDecorations: TotemDecoration[] = groups.flatMap(({ log, items, offerings }) =>
+	items.map(([itemName, level, xp, stringing, amount = 40]) => ({
+		log: Items.getOrThrow(log as string),
+		logAmount: amount as number,
+		item: Items.getOrThrow(itemName as string),
+		level: level as number,
+		xpPerLap: xp as number,
+		offerings: offerings as number,
+		stringing: stringing as boolean
```
gc: These casts are needed because you havent given a type for the arrays, just adding `as const`  to `groups` should work. if not, giving it a type would work (`const groups: {...}[] = [...`).

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+
+	if (!user.hasEquippedOrInBank(itemID('Fletching knife'))) {
+		timePerLap += FLETCHING_KNIFE_PENALTY;
+	} else {
+		messages.push(`-${formatDuration(FLETCHING_KNIFE_PENALTY)} for Fletching knife`);
+	}
+
+	const agilityLevel = user.skillLevel('agility');
+	const hasBasket =
+		user.hasEquippedOrInBank(itemID('Log basket')) || user.hasEquippedOrInBank(itemID('Forestry basket'));
```
gc: Preferably use `EItem` here. If you're writing the item name in the code and just want the ID, then use `Eitem` is probably always a better idea.

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	if (!itemToFletch) {
+		return `${user.minionName} have to select an item to fletch during Vale Totems.`;
+	}
+
+	const fletchableItem = Items.getOrThrow(itemToFletch);
+
+	const decoration = ValeTotemsDecorations.find(i => i.item.id === fletchableItem.id);
+	if (!decoration) return `Internal error: ${itemToFletch} not found`;
```
gc: **For the `getOrThrow`:** the user can pass any item name they want to, meaning we'll throw an error if they pass an invalid item (e.g. a fish instead of a log), in that case, we should just say its in an invalid item and not throw.
 and the `return `internal error`.  
**Option 1:** Replace the command to use `choices: choicesOf(groups.map(g => g.item.name))`, meaning the user MUST choose a valid item, and cant write any item they want.
**Option 2:** use `.getItem` instead, and if the item is null, `return "That's not a valid item."` 

**For the `return Internal error`**: If this would happen, we wouldnt know about it unless a user told us, because the user can pass any item name they want to. If you change the above thing to uses `choices`, then you can use a non-null assertion, or throw an actual error. Otherwise, this isn't really an error, the user just gave an invalid item, so remove the 'internal error' part.

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+		if (userBank.amount(stringItem.id) < STRINGS_PER_LAP)
+			return `${user.minionName} need at least ${STRINGS_PER_LAP} Bow strings to fletch ${decoration.item.name}!`;
+		if (!user.hasEquippedOrInBank(itemID('Bow string spool'))) {
+			timePerLap += STRING_NO_SPOOL_PENALTY;
+		} else {
+			messages.push(`-${formatDuration(STRING_NO_SPOOL_PENALTY)} for Bow string spool`);
+		}
+	}
+
+	const maxTripLength = calcMaxTripLength(user);
```
gc: Use `user.calcMaxTripLength('ValeTotems');`

src/mahoji/commands/minigames.ts
```diff
@@ -1063,6 +1070,92 @@ export const minigamesCommand = defineCommand({
...
+					options: [
+						{
+							type: 'String',
+							name: 'item_to_fletch',
+							description: 'Item to fletch during minigame.',
+							required: true,
+							autocomplete: async (value: string) => {
+								return ValeTotemsDecorations.filter(i =>
+									!value ? true : i.item.name.toLowerCase().includes(value.toLowerCase())
+								).map(i => ({ name: i.item.name, value: i.item.name }));
```
gc: For your 3 autocomplete's, you can change these to be strict-choices (instead of letting users type any string they want to). For examples, search the code for `choicesOf`. Ensure `ValeTotemsDecorations` is typed or is `as const` so typescript knows what choices are allowed.

**However,** the benefit of using an autocomplete is that: you can (1) filter the results so you only show users things that they have the level for (and/or the items for). secondly, you can show how much of the items they have in the results. e.g. instead of `Oak shortbow`, you can show `Oak shortbow (350x in bank)`. if you want to do this, just use `autocomplete: async (value: string, user: MUser) => {` and `user.bank`.

I am okay with leaving it as is, or these other 2 options. you can do which one you think is best.

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+		fletchXp: fletchingXp,
+		userID: user.id,
+		duration: duration,
+		channelID: channelID.toString(),
+		minigameID: 'vale_totems',
+		quantity: laps,
+		logName: logType.name,
+		itemName: fletchableItem.name,
+		staminaPot: staminaPot
+	});
```
gc, line 199: Ideally, you would never use item names in the trip data here, because they take more space in the database compared to just the item id, and in general using the item id is a better idea anyway. however, you dont have to.

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+		messages.push(`+${boost}% to offerings for minion learning`);
+	}
+
+	const totalOfferings = Math.floor(laps * decoration.offerings * (1 + boost));
+
+	const cost = new Bank();
+	cost.add(logType.id, logCost);
+	if (decoration.stringing) {
+		const stringCost = STRINGS_PER_LAP * laps;
+		cost.add(stringItem!.id, stringCost);
```
gc: Here you're using a non-null assertion, but just above you're checking if its truthy. but it looks like `stringItem` is only ever a bow string, so getting it in the first place doesnt seem necessary? just use `EItem.BOW_STRING`

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
```diff
@@ -0,0 +1,414 @@
...
+		return `${user.minionName} doesn't own: ${logCostLap}x ${logType.name}.`;
+	}
+
+	if (fletchingLvl < decoration.level) {
+		return `${user.minionName} needs ${decoration.level} Fletching to fletch ${decoration.item.name}.`;
+	}
+
+	let staminaPotItem: Item | undefined;
+	if (staminaPot) {
+		staminaPotItem = Items.getOrThrow(EItem.STAMINA_POTION4);
```
gc: Same here for stringItem/staminaPotItem (this variable doesnt really need to exist)

src/lib/skilling/skills/fletching/fletchables/atlatl.ts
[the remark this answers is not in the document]
gc: FYI: you can just ignore ellipsis, you dont have to respond. it gets a lot of things wrong

src/tasks/minions/minigames/valeTotemActivity.ts
```diff
@@ -0,0 +1,113 @@
...
+			if (roll(100)) {
+				loot.add(preRollTable.roll());
+				continue;
+			}
+			loot.add(rummageOfferings(fletchingLvl, hasForestryKit));
+		}
+
+		// Dirty arrowtips and it's logic
+		const DIRTY_ARROWTIPS_ID = 31047;
+		const hasDirtyArrowtips = loot.has(DIRTY_ARROWTIPS_ID);
```
gc: You can just use `loot.has('Dirty arrowtips')`, it should not really ever be necessary to write the item id out like this.

src/tasks/minions/minigames/valeTotemActivity.ts
```diff
@@ -0,0 +1,113 @@
...
+		const fletchingLvl = user.skillLevel('fletching');
+		const hasForestryKit = user.hasEquippedOrInBank('Forestry kit');
+
+		const loot = new Bank();
+		for (let i = 0; i < rewards; i++) {
+			if (roll(100)) {
+				loot.add(preRollTable.roll());
+				continue;
+			}
+			loot.add(rummageOfferings(fletchingLvl, hasForestryKit));
```
gc: the way this is coded, you're remaking the exact same loottable for every iteration, and rolling it once each time. alternatively, you could do something like: `const rummageTable = createRummageOffferingsTable();` outside the loop. this isn't a must, its just inefficient though.

gc (changes requested):
This is impressive for your first PR, i was surprised this was your first. I merged master into your PR to get rid of the conflicts. Some comments:

(1) Is the loot accurate? for a 8 hour trip with your code, with redwood hiking staffs, I get 3278 bird nests and 327 yew seeds. in this youtube video (also 8 hours of redwoods), they get like 10x less: https://www.youtube.com/watch?v=aeWrOjohQGE. The loot and xp/hr being accurate is arguably the most important thing. FYI: i did this by just setting the maxtriplength to `Time.Hour * 8`

For a potentially useful pattern, I now would code the meat of the logic outside of the activity/command, in a function that you can easily call anywhere, to simulate the loot/xp without needing to run trips. For example, like at the `fish.ts` command and `fishingActivity.ts`, 

(2) If the discord bot you're using to test this isnt in our test server yet, let me know in discord and i'll add it. this way others can help you with testing.

(3) I've left some comments/requests in the PR, most aren't a 'must' but just good practice. If you have any questions or anything, you can ping me #developers.

src/tasks/minions/minigames/valeTotemActivity.ts
Jettucis: My mistake, overlooked.

Jettucis:
```
lootTable.add(seedTable, 1, 11);
lootTable.add(valeOfferingNests, nestQty, 10);
```

Clearly the seeds should be more than nests

src/mahoji/commands/minigames.ts
Jettucis: Initially I had choices (at least for `item_to_fletch`), but I did get discord api error (because you can't have over x amount of choices present), so I did it as autocomplete, but buy and sell command could be remade as choices, on it.

src/mahoji/lib/abstracted_commands/valeTotemsCommand.ts
Jettucis: I'm quite not understanding this comment. Do I just write `shopItem.output.clone().multiply(quantity)` ? Or I have to add that `.freeze()`?

Jettucis: Can I just use the same logic I did for Buy command?
i.e.:
`itemsToRemove: new Bank(shopItem.output).multiply(quantity)`

gc: freeze wouldnt be used anywhere in the command, you dont _need_ it, it just prevents modifying a bank that shouldnt be modified. that logic is fine.

Jettucis:
I refactored some code (again), to accommodate latest changes.

Jettucis:
@gc sir, review pls

PR #6560, opened 2025-10-03
Add admin desync command

TastyPumPum:


gc:
I read the description before you changed it, and it seemed like a good idea (to automatically remove all commands on shutdown), although it would probably be very annoying for regular development because the commands would get synced on every reboot, and then that _might_ make the commands briefly unavailable when you switch to discord to try one. Having to manually /desync your commands once you're done would probably be better.

We should also just kick bots that are very inactive (it takes a few seconds to re-add them), that would probably make the biggest difference and be the easiest thing.

TastyPumPum:
> I read the description before you changed it, and it seemed like a good idea (to automatically remove all commands on shutdown), although it would probably be very annoying for regular development because the commands would get synced on every reboot, and then that _might_ make the commands briefly unavailable when you switch to discord to try one. Having to manually /desync your commands once you're done would probably be better.
> 
> We should also just kick bots that are very inactive (it takes a few seconds to re-add them), that would probably make the biggest difference and be the easiest thing.

Yeah I was switching to an /admin desync_commands command after some thinking which is why I deleted the description and stuff while I was changing it. 

I did kick 4 test bots that havent been used in over a year from the server to tidy it up in the meantime. But alot of people get put off testing because of how hard it is using commands with 5+ different bots showing so i think something like this is really needed in the testing server.

If you want to handle it yourself im more than happy with that, otherwise ill let you know when I have this ready for review

PR #6565, opened 2025-10-06
Pglite

gc:
### Other checks:


gc:
(1) PGLite; test bots dont need a real postgres database setup now, you can just run the bot and it'll just magically work. It stores the database files in the `.db` folder, you can delete this folder anytime. Add `USE_REAL_PG=1` to .env to use a normal postgres database.
(2) Prisma updated, it now uses the javascript client instead of the native client, which has several benefits.
(3) Switched robochimp from fastify to hono.

PR #6576, opened 2025-10-09
Add botOwnerID support to adminUserIDs

TastyPumPum:
Added optional BOT_OWNER_ID support to the global config schema, parsing the value from the environment and automatically including it in the admin user list. 

Documented the new optional BOT_OWNER_ID entry in .env.example so developers can configure their own account without editing source files.


gc:
I've done this here: https://github.com/oldschoolgg/oldschoolbot/commit/a330714a907f119258e9969fd79c662d2a34f5a0 - in a way so that you dont need to edit your .env, since you can get the bot owner in this way.

PR #6636, opened 2025-11-05
Nex Nihil Shard Droprate

LamofGod01:
### Description:

Updated droprate and quantity of nihil shards on NexNonUniqueTable to match OSRS rates.

### Changes:

Replaced current tertiary 1/25 drop of 1-20 nihil shards with 2 separate tertiary drops with rates and quantities matching the confirmed droprates of nihil shards from OSRS Nex (as confirmed by Mod Ash)

### Other checks:

- Method of having 2 tertiary nihil shard drops gives possibility of both drops in one kill (different behavior from OSRS) but this does not affect expected average

### DIFF
--- a/src/lib/simulation/misc.ts
+++ b/src/lib/simulation/misc.ts
@@ -217,5 +217,6 @@ const NexNonUniqueBaseTable = new LootTable()
 
 export const NexNonUniqueTable = new LootTable()
 	.every(NexNonUniqueBaseTable, 2)
-	.oneIn(25, 'Nihil shard', [1, 20])
+	.oneIn(16.4, 'Nihil shard', [80, 85])
+	.oneIn(26.13, 'Nihil shard', [85, 95])
 	.oneIn(100, 'Rune sword');


src/lib/simulation/misc.ts
```diff
@@ -217,5 +217,6 @@ const NexNonUniqueBaseTable = new LootTable()
 
 export const NexNonUniqueTable = new LootTable()
 	.every(NexNonUniqueBaseTable, 2)
-	.oneIn(25, 'Nihil shard', [1, 20])
+	.oneIn(16.4, 'Nihil shard', [80, 85])
```
DayV-git, line 220: I think these numbers are per kill, so would need to be scaled by team size (and maybe contribution?)  Easiest thing to do would be just divide by 4, as that's the team size we use for solos

LamofGod01: The drop quantities of all the drops on this table are per kill, so this isn't the right place to adjust for team size or contribution.

DayV-git: This table is rolled once for each user in the team. It's likely that all the drop quantities are wrong and don't reflect this

LamofGod01: Yea I think you might be right. The place to handle this would probably be when nex is rolling the drop though, rather than hard coding a specific assumption about team size or contribution into the drop rates. I'm less confident making a change to the nex drop mechanics, but it could be pretty simple to just multiply the drop quantity by contribution % to bring the quantities down to more realistic levels.

code1100: As a stopgap between how it is now and dynamically scaling the drops based on team size and contribution, keeping the single shard line in NexNonUniqueTable and changing to `	.oneIn(20, 'Nihil shard', [18, 22])` is a rough approximation of the 1 shard/kill average that 4 man Nex drops and a really easy change to make

Issue #6644, opened 2025-11-12
Lumina in potions bank filter


thueolesen-commits:
## Lumina in potions bank filter

Is it possible to have Lumina be placed in the /bank filter Potions along restores and brews?

PR #6645, opened 2025-11-12
Add Lumina to potions filter

nwjgit:
### Description:
Add Lumina to potions filter.

### Changes:
Add Lumina to potions filter.

### Other checks:
- Closes: #6644


gc:
Doesn't matter here, but FYI, you can do this: `[BSOItem.LUMINA, ...otherItems]`, so that you wouldnt need to call resolveItems. Thank you for the PR.

PR #6650, opened 2025-11-18
Allow per-user tracking for patron XP gains and drystreak tools

TastyPumPum:
This PR extends the /tools patron xp_gains and /tools patron drystreak commands so they can target a specific account instead of only showing the top 10.

For XP gains, an optional user_id argument is now accepted. When provided, the command sanitises and validates the Discord ID, filters the xp_gains query to that user, and replies with a single-line summary of their XP over the chosen interval (or a friendly “no gains found” message). Leaving user_id blank keeps the existing leaderboard behaviour unchanged.

For drystreaks, an optional user_id argument has been added alongside new helper functions for per-user checks. For monsters and openables, the command now returns a tailored message that:
- tells the user if they already have the CL item,
- explains if they have 0 KC / 0 opens (so they’re not actually dry), or
- shows their current dry streak in KC or openable count.

If no user_id is supplied, the original “top 10 driest players” leaderboard behaviour is preserved.


src/mahoji/commands/tools.ts
```diff
@@ -26,6 +26,16 @@ import {
...
+const skillMap = new Map(skillsVals.map(skill => [skill.id, skill] as const));
+const discordIDRegex = /^\d{17,20}$/;
+
+function sanitizeDiscordID(raw: string): string | null {
+	const cleaned = raw.replace(/[^0-9]/g, '');
+	if (!discordIDRegex.test(cleaned)) {
+		return null;
+	}
+	return cleaned;
+}
```
gc: We have a `isValidDiscordSnowflake` that would be better to use here, removing the discordIdRegex here.

```ts
import { isValidDiscordSnowflake } from '@oldschoolgg/util';

if (!isValidDiscordSnowflake(inputSnowflake)) return 'Invalid';
```

However, I think it would be better to instead just use a `type: 'User',` instead of a `type: 'String',`, then you just get a user object directly given to you, and for users they dont need to figure out how to get a user id.

Side note: this lets you check anyones xp gains/drystreaks? seems okay to me although someone might complain

gc (changes requested):

src/mahoji/commands/tools.ts
TastyPumPum: Fixed, i did also consider your side note when I was coding it. But thought people can do this on osrs anyway with the use of wise old man or osrs hiscores etc. So thought it was fair to get added here.

Side note - I also added it to clue_gains and kc_gains, I intentionally left most_drops as i thought that would be more of an invasion and could lead to people dming someone for stuff etc

PR #6664, opened 2025-11-26
Preparations for Sailing Part 1

Jettucis:
### Description:

Preparations for Sailing Part 1 [Items]

### Changes:

New items
New eatables (commented out to not affect existing logic)
New Collection Logs (Other subtype for now)
New creatable (Medallion of the Deep - commented out)

### Other checks:

- [x] I haven't tested all my changes thoroughly.


packages/oldschooljs/src/assets/item_data.json
```diff
@@ -225202,66 +225672,55 @@
 		"wiki_name": "Runescroll of bloodbark",
 		"price": 8000
 	},
-	"25484": {
```
DayV-git, line 225205: There are some items that don't exist in the osrsbox db but are still used here, make sure they aren't removed. Basically just don't remove any items

Jettucis: added them back

PR #6683, opened 2025-12-12
Add optional trip end timestamp for Perk Tier 4+ users

### LINKED ISSUE #1879: Return time as time of day
## 🚀 Feature Proposal
Ability for t3+ patrons(or any1 idk) to set a time zone and instead of having “your minion will return in x minutes y seconds have.for example:(+timezone -8:00) to set ur timezone as pst(?) then when u send a minion out, lets say it’s 11:00:15am and ur trip will take 30mins, instead of reading as “your minion will return in 30 minutes” it would instead display  “your minion will return at 11:30:15am”, this feature could also be implemented without the ability to specify timezone(so that it would show gmt(?) time always) I believe this would be not too hard to implement(I’m just not familiar with how the bot is coded), I think this would be a great addition to the bot :).

TastyPumPum:
Perk Tier 4+ users can now toggle showing the exact local time their trip will finish.

Added formatTripDuration
Uses Discord timestamp formatting
Controlled by BitField.ShowMinionReturnTime

Redo of #6431

Closes https://github.com/oldschoolgg/oldschoolbot/issues/1879


TastyPumPum:
Fixed latest conflicts and added a test file

PR #6693, opened 2025-12-14
Fix return trip Wintertodt

BartSimp3:
### Description:

Changed the way how wintertoad get rolled what gave errors on return trip

### Changes:

Changed the logic how wintertoad gets rolled.

### Other checks:


gc:
I've fixed the bug in a different way (in the clAdjustedDroprate itself, incase this same bug is elsewhere), and committed it myself to save us time since your PR isn't formatted. Thank you for the PR

BartSimp3:
Thanks! Kinda new to this.

PR #6724, opened 2026-01-13
Add mention command to repeat last minion trip

TastyPumPum:
- Provide a quick mention-based command so users can type @bot r (or @bot repeat) to rerun their most recent repeatable minion trip.
- Reuse the existing repeatTrip flow so mention-based repeats behave the same as button/command repeats.
- Create a lightweight MInteraction wrapper for message mentions so command handlers expecting an interaction can be invoked from plain messages.


TastyPumPum:
Fixed for latest conflicts, the change in giveaway.ts was just a readability cleanup while resolving conflicts as there was already a variable called perkTier earlier in the file

PR #6729, opened 2026-01-20
Add the Stealing artefacts minigame

TastyPumPum:
Implement the Stealing artefacts minigame.

Provide an option to glassblow items during the activity that consumes Molten glass and awards Crafting XP (with its own hourly cap), and wire the minigame to the Kourend & Kebos Hard diary requirement.

### DIFF
--- a/data/osb/commands.json
+++ b/data/osb/commands.json
@@ -1,6 +1,6 @@
 {
-	"hash": "4e6b5f19ee746f8ca06542ac4d142529",
-	"date": "2026-02-20T07:19:34.772Z",
+	"hash": "06036cfe98398bd7359c3290cc8776b1",
+	"date": "2026-03-31T15:05:49.510Z",
 	"data": [
 		{
 			"name": "activities",
@@ -353,6 +353,7 @@
 				"sepulchre",
 				"shades_of_morton",
 				"soul_wars",
+				"stealing_artefacts",
 				"tears_of_guthix",
 				"temple_trek",
 				"trouble_brewing",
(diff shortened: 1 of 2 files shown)


src/lib/types/minions.ts
```diff
@@ -408,6 +408,18 @@ export interface PlunderActivityTaskOptions extends MinigameActivityTaskOptions
 	rooms: number[];
 }
 
+export interface StealingArtefactsActivityTaskOptions extends MinigameActivityTaskOptions {
+	type: 'StealingArtefacts';
+	stamina: boolean;
+	teleportOptionEnabled: boolean;
+	teleportEligible: boolean;
+	glassblow?: {
+		product: import('@/lib/minions/data/stealingArtefacts.js').StealingArtefactsGlassblowingProductKey;
```
gc: Shouldnt have an import here

src/tasks/minions/minigames/stealingArtefactsActivity.ts
```diff
@@ -0,0 +1,89 @@
+import { randInt } from '@oldschoolgg/rng';
```
gc: `import { randInt } from '@oldschoolgg/rng';`

Needs to change to:

```ts
async run(data: StealingArtefactsActivityTaskOptions, { user, handleTripFinish, rng }) {

rng.randInt();
```

gc (changes requested):
Is xp/hr, loot/rewards/etc, accurate? Generating a table of these rates to validate would be nice, but not required.

PR #6733, opened 2026-01-23
Improve Wiki command

TastyPumPum:
Improves the wiki command


gc (changes requested):
Can you describe what your changes are doing? what is the in flight mechanism for?

- Any cache, if necessary, needs to go in `src/lib/cache.ts`, so that we can keep an eye on them.  
-  `pruneWikiAutoCache` might not be necessary at all if you switch the Map to a LRUCache.

TastyPumPum:
> > The in-flight mechanism prevents multiple identical autocomplete searches from hitting the wiki simultaneously by reusing the same request. Let me know if you want it removed for ease as the commands not used often I'd imagine.
> 
> I think its better this is removed. The chances of this are 1 in a million, and even if it does happen, theres no harm in it, the extra code/complexity isnt worth it.

I removed the in-flight mechanism, I have moved the wiki autocomplete into its own helper file and added a small test for this.

TastyPumPum:
To add onto this now, wiki command seems to work. This feature should still improve user experience and is probably still a good idea.

PR #6735, opened 2026-01-23
Update fishingcontest reset time to 00:00 UTC

TastyPumPum:
Updated fishing contest daily queries to use an explicit UTC day range so resets align with 00:00 UTC.


TastyPumPum:
> I'm not understanding how this actually changes anything, it seems like both are identical (just written in different ways).
> 
> Before:
> 
> ```sql
> WHERE date::date = '2026-02-09'
> ```
> 
> After:
> 
> ```sql
> AND date >= 2026-02-09T00:00:00.000Z AND date < 2026-02-10T00:00:00.000Z
> ```

Same intent, but the range query should be safer. 

With date::date it changes the value before comparing it, whereas date >= start AND date < end compares the original timestamp directly. 

It should also remove timezone ambiguity by using explicit UTC bounds. e.g. a record at 00:03 could otherwise fall into the wrong day depending on the DB/session timezone.

PR #6737, opened 2026-01-27
Add Black swan to DC pets bank filter

TastyPumPum:
Added a DC Pets bank filter that includes Black Swan without altering collection log entries

image


src/lib/data/filterables.ts
```diff
@@ -962,6 +965,11 @@ export const baseFilters: Filterable[] = [
 		aliases: ['tmb'],
 		items: () => tmbTable
 	},
+	{
+		name: 'Custom Pets (Discontinued)',
+		aliases: ['dc pets', 'dcpets', 'dcpet', 'dcp', 'discontinued custom pet', 'disc custom pet'],
+		items: () => [...discontinuedCustomPetsCL, ...dcPetsExtras]
```
gc, line 971: No changes needed, but for future reference, I'm preferring using `BSOItem.BLACK_SWAN` rather than  `resolveItems(['Black swan'])` (eliminates typos/wrong names, and requires no cpu work to figure out the item id).

PR #6742, opened 2026-01-30
Generate commands reference documentation

TastyPumPum:
Added a static command-doc generator script that parses the slash command definitions and writes the commands reference page.

Generated the OSB Commands reference page that documents usage, parameters, and examples from command definitions.


TastyPumPum:
> This is an interesting idea;
> 
> 1. Have you ran the docs locally to see what these pages actually look like? I'm okay with you using A.I, but you need to check everything that its doing. Every single generated command page errors (cant even be opened).
> 2. I'm not sure how helpful these pages are, atleast in some cases, for example:
> 
> chrome_6KFcgWClGZ
> It could definitely use some work to make it more useful to users or the design improved.

image

Made some significant improvements to the command reference since the original draft.

The docs are now generated from structured command data and rendered dynamically, which removes stale-page risk and makes the system much easier to maintain going forward.

I’ve also focused heavily on usability:
- collapsed command cards for scanability
- live search + jump navigation
- nested subcommands
- cleaner options tables with unused fields hidden
- Build, schema validation, and link checks are all passing locally.

This has been a commonly requested feature from newer players, so it should give us a much clearer place to direct them for command discovery.

gc:
I dont like that you submitted a broken/untested PR that was written with A.I, then after fixing it, you respond with an A.I generated message. its not a nice feeling, I read what people say and put thought into what i say. you havent said if you tested it this time. That issue aside, im still not particularly happy with a few things:

(1)  The javascript looks pretty overcomplex for just showing commands, if you're unfamilar with react or web dev in general then I can look past it, ideally, it would be either (a) react component(s) (b) generated like the Monsters page.
(2) The collapse/expand buttons dont match the theme, and it seems maybe better to make the entire thing clickable instead of just the button?
(3) The search bar and dropdown are unstyled.
(4) I'm not sure how useful it is to have `Examples: 0` for commands with no examples, just in general these ui elements showing the count seem just like visual clutter.
(5) The choices text seems unpolished: 
image
(6) The examples could use the special discord command syntax we have for the wiki (unless it doesnt work for some reason)

TastyPumPum:
Used AI to do the following:

1. Make commands/subcommands clickable to copy and paste into discord
2. Should now be a react-based components but have no idea how to test if it is (tried reading up on https://react.dev/reference/react/Component to see if I could get my head around it but had no luck)
3. Expand/collapse should now be theme consistent
4. Search/jump are theme consistent
5. Removed the clutter like "examples: 0"
6. Choices text should now be more user friendly without the dupes etc
7. Add tests to cover a few variations

Note: Only I have tested this, I am unsure how I could host the docs for others to test

Images that could be helpful for your review are placed below:

image
image
image
image
image

PR #6743, opened 2026-02-04
Remove unobtainable CLs

TastyPumPum:
Remove the unobtainable cls from showing. Should only show when they are actually obtainable so its more user-friendly and for people competing near the top of the mastery lbs


gc:
Maybe just comment them out, so we can much more easily add them back (and see that theyre missing). you can merge yourself when ready.

PR #6753, opened 2026-02-20
Enforce users to complete Learning the Ropes before using minion commands

TastyPumPum:
- Change the tutorial island quest to the correct name
- Block users from using minion commands until they have bought a minion and then completed the Learning the Ropes quest
- Add two new buttons to the minion buy response, one for becoming an ironman, the other for starting the Learning the Ropes quest. (there was no quest icon emoji from what I could see so I have used the leagues one for now, it can always be amended down the line or added now if it is required)
- Add the learning the ropes quest button to the ironman confirmation string.
- On completion of learning the ropes quest, adds a link to the wiki beginner guide for users
- Adds some small relevant tests

This PR has only been tested by myself and I have utilised AI tools to achieve my desired goals with this PR.

image

image

image


src/mahoji/lib/abstracted_commands/minionBuyCommand.ts
```diff
@@ -25,6 +31,6 @@ export async function minionBuyCommand(user: MUser, ironman: boolean): CommandRe
 <:BSO:863823820435619890> **BSO:** I run a 2nd bot called BSO (Bot School Old), which you can also play, it has lots of fun and unique changes, like 5x XP and infinitely stacking clues. Type \`/help\` for more information.
 
 Please click the buttons below for important links.`,
-		components: mahojiInformationalButtons
+		components: [learningTheRopesButton, becomeIronmanButton, ...mahojiInformationalButtons]
```
themrrobert, line 34: Can we just make it auto-send the trip when you buy a minion? 

It needs to be very clear, 'You have bought a minion! Your minion starts off with the tutorial, and will be ready soon!.'  (Runs the learning the ropes command starts activity)

I think this is the best way to do it.

themrrobert (changes requested):
Again, I am happy to make this change.... in fact i'll try to do it now

src/mahoji/lib/abstracted_commands/minionBuyCommand.ts
TastyPumPum: The only trouble with what you said is if they want to become an ironman, if they click a become ironman button it should cancel the trip and make them iron or something

TastyPumPum:
Closing as stale/unwanted.

PR #6754, opened 2026-02-20
BSO - Slayer Masks KC Fix

### LINKED ISSUE #6666: Slayer Masks kc not adding
## 🐛 Bug Report

Killing black demons on tasks does not save the kills to the mask. It works for abyssal demons 

image

TastyPumPum:
- KC was being added and tracked through a Bank and swap it to a plain object instead Fixed the leaderboard for slayer mask
- Updated slayer mask rates to match what was on the wiki
- Added some relevant tests that hopefully help ensure it doesn't break in future 

Used AI on this as it's been a long standing issue where none of the devs have had the time to look into it

Closes #6666

image


TastyPumPum:
> @TastyPumPum (Also, remove the `type` keyword prefixed to Bank. While it may not be necessary now, it could cause issues in the future, and there is zero benefit, since esbuild combines everything into one file, without duplicating the imported libraries.)

I tried to do this but whenever i run pnpm dev it automatically adds it back, not sure what else to do

PR #6762, opened 2026-02-23
Add a new Commit Command

TastyPumPum:
Adds a new commit command that shows the commit date and the code difference with buttons to view commit and recent commits

Added an optional boolean to show further details such as hash, status and bot started date.

Added relevant tests.

Used AI tools to help achieve my desired goals here, let me know if anything needs to be changed/adjusted.


TastyPumPum:
> It should probably show the commit message, or the last few commit messages (that are live in the bot). I think it would probably be better in a general `/info` command, which could show the commit, uptime, and other general things, rather than a command just to show the git commit.

Done, let me know if anything further should be added to the base command or the more detailed boolean version

image

src/mahoji/commands/info.ts
```diff
@@ -0,0 +1,67 @@
...
+			new ButtonBuilder().setLabel('View Commit').setStyle(ButtonStyle.Link).setURL(META_CONSTANTS.GITHUB_URL),
+			new ButtonBuilder()
+				.setLabel('Recent Commits')
+				.setStyle(ButtonStyle.Link)
+				.setURL(META_CONSTANTS.GITHUB_COMMITS_URL)
+		]
+	};
+}
+
+export const infoCommand = defineCommand({
```
themrrobert, line 82: So i'm not a mind reader, so Idk what magna meant exactly, but I would like the 'commit' to be a subcommand. 

So Like:
/info
-> commit  (Current full hash + name of commit)
-> recent_updates  (Last few commits with names + short hash, maybe 10)
-> uptime  (Does what it says on the tin)
-> overview (Shows commit, uptime, and maybe last 3 commits -> subject to change to whatever is relevent in the future)

Then we can add new subcommands for anything else we want

Also, you can add the details option, ie. `/info commit details:true` just like it exists here

themrrobert (changes requested):
I am happy to make these changes myself when I have time :) Anyone is welcome to do it meanwhile

PR #6830, opened 2026-03-27
Vale Totems Improvements

TastyPumPum:
This update improves vale totems visibility for players. 
It adds a new stats view, shows lifetime offerings rummaged, and includes that progress in collection log display. 
It also updates Vale Totems drop behaviour and docs to better match OSRS expectations and adds Vale Totems to drystreak tracking.

Closes #6825 


TastyPumPum:
Changed the drystreak to be based off rummages instead as that is more accurate

PR #6894, opened 2026-05-15
Persist maxSizeInput on CoX trips so Trio/5-Scale Speed-Runner CAs are reachable solo

typat:
The four CoX Speed-Runner Combat Achievements (Trio, 5-Scale, CM Trio, CM 5-Scale) reference `(data as RaidsOptions).maxSizeInput` as the second branch of an OR predicate, allowing solo players to qualify via `max_team_size`. `RaidsOptions` declares the field in `src/lib/types/minions.ts:514`, but `coxCommand` never persisted it to the trip — making that predicate branch dead code on BSO.

This means solo CoX/CM trips on BSO can never roll those CAs, regardless of the `max_team_size` parameter the user sets.

Fix: pass the already-parsed `maxSize` value to `ActivityManager.startTrip`, mirroring master's non-fake-mass behavior.

## Affected CAs (all unreachable solo before this fix)

  - CoX (Trio) Speed-Runner — `grandmaster.ts:102`
  - CoX (5-Scale) Speed-Runner — `grandmaster.ts:57`
  - CoX CM (Trio) Speed-Runner — `grandmaster.ts:152`
  - CoX CM (5-Scale) Speed-Runner — `grandmaster.ts:173`

## Test plan

  - [ ] Run a solo CM with `/raid cox start type:solo challenge_mode:True max_team_size:5`
  - [ ] Confirm `maxSizeInput` is now persisted on the trip (the field appears in the trip data passed to the CA roll)
  - [ ] Confirm the predicate at `grandmaster.ts:152-153` (Trio CM) and `:173-174` (5-Scale CM) evaluates true
  - [ ] Confirm both Trio and 5-Scale Speed-Runner CAs roll on a sub-27/25-minute trip
  - [ ] Sanity: a normal mass CoX trip still works (the field is also populated there, since `maxSize` is parsed from input regardless of party type)

Follow-up comment to post after the PR is open

Master has diverged further than just this field — it also has an `isFakeMass` feature where solo players can simulate a team of N (`maxSizeInput: isFakeMass ? fakeUsers : maxSize`, plus an `isFakeMass` flag persisted on the trip and a different `users` array population path). This PR intentionally stays minimal — just restores the persisted field so the existing CA predicate works. A follow-up could backport the full `isFakeMass` feature from master to BSO for upstream parity, but that's a larger change that would need to merge cleanly with BSO's chincannon code, so I left it out of this fix.


typat:
bump. is there anything i can do to help get this across the line?

PR #6906, opened 2026-05-21
Make Chincannon an eligible item in toa

Padl-osrs:
since the code for speed improvements is already there, i just added chincannon to the gear check / arrow reduction code

### Description:

add chincannon to eligible range items

### Changes:

add chincannon to eligible range items

### Other checks:

### DIFF
--- a/src/lib/simulation/toa.ts
+++ b/src/lib/simulation/toa.ts
@@ -199,7 +199,7 @@ const primarySpecWeaponBoosts = [
 	['Dragon claws', 6]
 ] as const;
 
-const REQUIRED_RANGE_WEAPONS = resolveItems(['Twisted bow', 'Bow of faerdhinen (c)']);
+const REQUIRED_RANGE_WEAPONS = resolveItems(['Twisted bow', 'Bow of faerdhinen (c)', 'Chincannon']);
 const MELEE_REQUIRED_WEAPONS = resolveItems(['Zamorakian hasta', 'Ghrazi rapier', "Osmumten's fang"]);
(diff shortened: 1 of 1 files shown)


themrrobert:
Most importantly: Has this been tested and verified that the chincannon works as intended with this change?

Also it needs to properly NOT work if the chincannon is disabled via the invention toggle. (This should happen automatically when it checks for charges, I don't remember off-hand how the charges are consumed). -- This one we could help you with if you aren't so sure of.

Padl-osrs:
> Most importantly: Has this been tested and verified that the chincannon works as intended with this change?
> 
> Also it needs to properly NOT work if the chincannon is disabled via the invention toggle. (This should happen automatically when it checks for charges, I don't remember off-hand how the charges are consumed). -- This one we could help you with if you aren't so sure of.

currently waiting for tasty to get pc access, since i dont have a testchannel yet. if you could try it out, would help me alot

BartSimp3:
This will not work tbf. This only allow you to wear a chincannon while doing toa.

Padl-osrs:
> This will not work tbf. This only allow you to wear a chincannon while doing toa.

yes but the code for it to work was already there. But you couldnt reach it because you were not allowed to wear it

BartSimp3:
Where is the code for it to work? Check the CoX files.

Padl-osrs:
just checked, either have been removed or wasnt there. only checks for !chincannon.
wiki also says 60% boost. Maybe i will find something in the git logs

Padl-osrs:
https://github.com/oldschoolgg/oldschoolbot/pull/6915

i created a new one from my pc instead of the online editor - tested it already

PR #6913, opened 2026-05-30
Set real postgres db as default in .env.example

DayV-git:
### Description:

PGLite is janky

### Changes:

Set real postgres db as default in .env.example

### Other checks:


Padl-osrs:
good idea!

maybe we can also pull the server id in all spots from .env instead of hardcoded real server

PR #6935, opened 2026-06-12
Improve Minion Kill Message

TastyPumPum:
Adds headings for boosts, charges, pk, food and item costs

Adds back all the details we had about food reduction 2 years ago

Adds charges remaining to charge boost message


DayV-git:
Messages on trip send/return were intentionally removed to reduce bloat

TastyPumPum:
> Messages on trip send/return were intentionally removed to reduce bloat

Now put it under the show detailed info bitfield toggle so players can opt in for the more detailed message

PR #6966, opened 2026-07-09
Update Zulrah Boss Task Quantity

TastyPumPum:
Not sure why zulrah was set with a 15 cap, checked all over the osrs wiki and couldnt see any reason

So this pr updates it to a cap of 35 like the rest instead


DayV-git (approved):
Solid implementation of an osrs update from 5 years ago

PR #7048, opened 2026-08-19
Remove supplies

DarkWorldsArtist:
### Description:

remove supplies from dt

### Other checks:


DarkWorldsArtist:
will be fixed in patch 2

PR #7050, opened 2026-08-19
fix tradeable/tmb

DarkWorldsArtist:


DarkWorldsArtist:
will also be fixed in patch 2

